PR #3, opened 2019-01-25
Add docker

RoriCremer:


.eslintrc.js
```diff
@@ -0,0 +1,86 @@
+module.exports = {
+  'plugins': ['lodash-fp', 'react'],
+  'extends': 'react-app',
+  'globals': {
+    'SATURN_VERSION': false,
```
ddietterich, line 5: Will these eventually be JADE things?

ddietterich (commented):
LGTM FWIW :)

.eslintrc.js
thathert: Yes, eventually once we remove the instance of SATURN_VERSION, etc from the code.

docker_push
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+docker push USER/REPO
```
thathert: I don't think we need this file. We can just update the .travis.yml and say `... && docker push "$DOCKER_USERNAME"/"$DOCKER_REPO"`

thathert (changes requested):
Another minor change, looks good though!

PR #6, opened 2019-02-27
preview: Rc dr 149 material

RoriCremer:
testing coverage only at 79%...

redux not picking up states on the DatasetCreate


src/components/DatasetCreateView.jsx
```diff
@@ -113,73 +120,152 @@ export class DatasetCreateView extends React.PureComponent {
...
           <FormRow>
-            <label htmlFor="dataset.name">Dataset name:</label>
             <Control.text
               model="dataset.name"
               id="dataset.name"
               required
               validators={{ name: this.validateName }}
+              component={(props) =>
+                <TextField
+                  {...props}
```
thathert, line 174: What's going on here? Does this solve the problem that was happening where uploading IDs was resetting the form?

src/components/DatasetCreateView.jsx
```diff
@@ -113,73 +120,152 @@ export class DatasetCreateView extends React.PureComponent {
...
+                  label="Dataset Name"
+                  defaultValue={this.props.model}
+                  name={this.props.model}
```
thathert: I'm confused, `name` doesn't feel like it should be the same thing as `defaultValue`.

src/components/DatasetCreateView.jsx
```diff
@@ -113,73 +120,152 @@ export class DatasetCreateView extends React.PureComponent {
...
-            <Control.text id="dataset.readers" model="dataset.readers" validators={{ isEmail }} />
+            <Control.textarea
+              model="dataset.description"
+              id="dataset.description"
+              required
+              component={(props) =>
+                <TextField
+                  {...props}
+                  defaultValue={this.props.model}
+                  style={{width: '800px'}} //fullWidth
```
thathert, line 193: I'm concerned about hardcoding pixel values here. Would like to see if we can find an approach following the flex model (out of the scope of this PR).

src/components/DatasetCreateView.jsx
```diff
@@ -113,73 +120,152 @@ export class DatasetCreateView extends React.PureComponent {
...
+                </Select>
+              }
+            />
           </FormRow>
 
           <FormRow>
-            <label htmlFor="dataset.upload">Import IDs: </label>
-            <input type="file" id="dataset.upload" onChange={this.parseFile} />
+            <input type="file" id="dataset.upload" onChange={this.parseFile}  style={{display: 'none'}}/>
+            <label htmlFor="dataset.upload">
```
thathert, line 258: Nice

src/routes/Private.jsx
```diff
@@ -12,23 +16,26 @@ class Private extends React.Component {
...
-            <Route exact path="/studies" component={StudiesView} />
-            <Route exact path="/datasets" component={DatasetView} />
-            <Route exact path="/datasets/create" component={DatasetCreateView} />
-          </Switch>
+          <Route
+            path="/"
+            render={({ location }) => (
+              <Fragment>
+                <AppBar position="static">
+                  <Tabs value={location.pathname && '/' + location.pathname.split('/')[1]}>
```
thathert: :(

thathert (changes requested):
Looks good! A couple things noted above. Also, I'm concerned about the material ui tabs vs react router (see the frowning emoticon comment)

thathert (approved):
LGTM, thanks!

PR #7, opened 2019-03-02
UI cleanup

thathert:
- removes snapshots
- stops using styled components
- gets themes set up for material ui


.stylelintrc
```diff
@@ -16,6 +15,7 @@
         "ignoreAtRules": ["else"]
       }
     ],
+    "declaration-block-trailing-semicolon": null,
```
thathert, line 18: This started acting up -- it was incorrectly thinking I should add a comma after things like `style={{ display: 'block' }}`. I'm turning it off

PR #14, opened 2019-03-19
Create dataset poller

RoriCremer:
Screen Shot 2019-03-19 at 2 23 45 PM

Needs a bit of cleanup, also renaming of the actions--though I suppose this could be delayed until we have SAM hooked in

### DIFF
--- a/src/actions/app.js
+++ b/src/actions/app.js
@@ -8,8 +8,6 @@ import { createActions } from 'redux-actions';
 
 import { ActionTypes } from 'constants/index';
 
-export { goBack, go, push, replace } from 'modules/history';
-
 export const { createDataset } = createActions({
   [ActionTypes.CREATE_DATASET]: dataset => dataset,
 });
@@ -32,4 +30,9 @@ export const { getDatasetById } = createActions({
 export const { getStudyById } = createActions({
   [ActionTypes.GET_STUDY_BY_ID]: study => study,
   [ActionTypes.GET_STUDY_BY_ID_SUCCESS]: study => study,
+});
+
+export const { getJobById } = createActions({
+  [ActionTypes.GET_JOB_BY_ID_SUCCESS]: datasets => datasets,
+  [ActionTypes.GET_JOB_BY_ID]: datasets => datasets,
 });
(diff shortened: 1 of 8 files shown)


src/sagas/repository.js
```diff
@@ -13,13 +14,55 @@ import { ActionTypes } from 'constants/index';
...
  */
+
+/**
+ * Saga poller
+ */
+function* pollJobWorker(jobId) {
+  let jobStatus = 'running';
+  while (jobStatus === 'running') {
+    try {
+      const response = yield call(axios.get, '/api/repository/v1/jobs/' + jobId);
```
thathert: Instead of a while loop I think there is a nicer way to approach just using the response from the job endpoint.

src/actions/app.js
```diff
@@ -8,7 +8,7 @@ import { createActions } from 'redux-actions';
 
 import { ActionTypes } from 'constants/index';
 
-export { goBack, go, push, replace } from 'modules/history';
+export { goBack, go, push, replace } from 'modules/history'; // TODO why is this here?
```
thathert: I think we can remove this -- this was not here before, did you add it?

src/sagas/repository.js
```diff
@@ -13,13 +14,49 @@ import { ActionTypes } from 'constants/index';
...
+      yield put({
+        type: ActionTypes.GET_JOB_RESULT_SUCCESS,
+        payload: { jobResult: resultResponse.data },
+      });
+    } else {
+      yield put({
+        type: ActionTypes.GET_JOB_BY_ID_SUCCESS,
+        payload: { status: response.job_status },
+      });
+      yield call(delay, 4000);
```
thathert: Let's make this 1000 ms

src/sagas/repository.js
```diff
@@ -13,13 +14,49 @@ import { ActionTypes } from 'constants/index';
...
+    const jobStatus = response.data.job_status;
+    if (jobStatus !== 'running') {
+      const resultResponse = yield call(axios.get, '/api/repository/v1/jobs/' + jobId + '/result');
+      yield put({
+        type: ActionTypes.GET_JOB_RESULT_SUCCESS,
+        payload: { jobResult: resultResponse.data },
+      });
+    } else {
+      yield put({
+        type: ActionTypes.GET_JOB_BY_ID_SUCCESS,
```
thathert, line 40: We should distinguish between success and failure here -- this branch will be reached by both cases (I think it should be two different actions)

thathert (changes requested):
Looks good so far, a couple points

thathert (approved):

PR #16, opened 2019-03-24
Short term fix for a user to stay signed in

RoriCremer:
This returns the google user obj on load, so as long as signed somewhere in the browser, the user shouldn't be logged out

This is a shim --
Once we have SAM and Terra registration integrated, we may need to roll our own component rather than using the react-google-login
Should be a straightforward process given the code already used by the Terra team

signinui

### DIFF
--- a/src/components/WelcomeView.jsx
+++ b/src/components/WelcomeView.jsx
@@ -82,11 +82,16 @@ class WelcomeView extends React.PureComponent {
           Welcome to the Jade Data Repository
           </div>
           <div>
-            <GoogleLogin // stealing clientId from Terra
+            <GoogleLogin // TODO this component may be unuseable once we require a terra registration
+              // stealing clientId from Terra
               clientId="500025638838-s2v23ar3spugtd5t2v1vgfa2sp7ppg0d.apps.googleusercontent.com"
               buttonText="Sign in with Google"
               onSuccess={onSignInSuccess}
               onFailure={onSignInFailure}
+              isSignedIn
+              cookiePolicy={'single_host_origin'}
+              prompt="select_account"
+              scope="openid profile email"
               theme="dark"
             />
           </div>


src/components/WelcomeView.jsx
```diff
@@ -67,6 +67,7 @@ class WelcomeView extends React.PureComponent {
     };
 
     const onSignInSuccess = (user) => {
+      console.log(user);
```
thathert: missed `console.log`

src/components/WelcomeView.jsx
```diff
@@ -82,11 +83,16 @@ class WelcomeView extends React.PureComponent {
           Welcome to the Jade Data Repository
           </div>
           <div>
-            <GoogleLogin // stealing clientId from Terra
+            <GoogleLogin // TODO this component may be unuseable once we require a terra registration
```
thathert, line 85: This will still be required -- the user needs to have a google account AND be a terra user. We'll need to do the terra check separately after we confirm they are logged in.

src/components/WelcomeView.jsx
```diff
@@ -82,11 +83,16 @@ class WelcomeView extends React.PureComponent {
...
+              // stealing clientId from Terra
               clientId="500025638838-s2v23ar3spugtd5t2v1vgfa2sp7ppg0d.apps.googleusercontent.com"
               buttonText="Sign in with Google"
               onSuccess={onSignInSuccess}
               onFailure={onSignInFailure}
+              isSignedIn
+              cookiePolicy={'single_host_origin'}
```
thathert, line 92: What are the other options here? Thinking we will want to pick something that makes it possible to switch here from terra without re-logging in.

thathert (changes requested):
Looks good, one change and a TODO update/question

src/components/WelcomeView.jsx
RoriCremer: I'm guessing I will need to re-write this anyway once SAM has been hooked up.

thathert (approved):

PR #17, opened 2019-03-25
Manage users

RoriCremer:
On the manage users modal:

Screen Shot 2019-03-25 at 1 04 56 PM

And on the create dataset page:

Screen Shot 2019-03-25 at 1 08 07 PM

Need to set up SAM and add full functionality to the chips
Chips currently hardcoded!

### DIFF
--- a/src/components/DatasetCreateView.jsx
+++ b/src/components/DatasetCreateView.jsx
@@ -255,7 +255,7 @@ export class DatasetCreateView extends React.PureComponent {
               <Link to="/datasets">Cancel</Link>
             </Button>
             <Button variant="contained" color="primary">
-              <Link to="/datasets/preview">Preview Data</Link>
+              <Link to="/datasets/preview">Create Dataset</Link>
             </Button>
           </FormRow>
         </Form>
--- a/src/components/DatasetCreateView.jsx
+++ b/src/components/DatasetCreateView.jsx
@@ -83,18 +83,18 @@ export class DatasetCreateView extends React.PureComponent {
     return studiesList;
   }
 
-  addReader(newEmail) {
+  addUser(newEmail) {
     const { dispatch, readers } = this.props;
     if (!_.includes(readers, newEmail)) {
       dispatch(actions.change('dataset.readers', _.concat(readers, newEmail)));
     }
   }
 
-  removeReader(removeableEmail) {
+  removeUser(removeableEmail) {
     const { dispatch, readers } = this.props;
-    const newReaders = _.clone(readers);
-    _.remove(newReaders, r => r === removeableEmail);
-    dispatch(actions.change('dataset.readers', newReaders));
+    const newUsers = _.clone(readers);
+    _.remove(newUsers, r => r === removeableEmail);
+    dispatch(actions.change('dataset.readers', newUsers));
   }
 
   parseFile = event => {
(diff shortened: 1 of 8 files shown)


src/components/ManageUsersModal.jsx
```diff
@@ -0,0 +1,138 @@
...
+    this.setState({
+      open: true,
+    });
+  };
+
+  handleClose = () => {
+    this.setState({ open: false });
+  };
+
+  addReader(newEmail) {
```
thathert: Let's make this more generic. We want this to be used for custodians, readers, and discoverers. How about calling this `addUser`?

src/components/ManageUsersModal.jsx
```diff
@@ -0,0 +1,138 @@
...
+    const { datasetUUID } = this.props;
+    //TODO where do I want to keep readers? maybe directly gotten out of SAM?
+    // so need to call something to put them directly in SAM
+    console.log(newEmail);
+    console.log(datasetUUID);
+    // dispatch();
+  }
+
+  removeReader(removeableEmail) {
```
thathert: Same here.

src/components/DatasetDetailView.jsx
```diff
@@ -97,7 +97,11 @@ export class DatasetDetailView extends React.PureComponent {
             <div className={classes.header}> Last Modified: </div>
             <div className={classes.values}> {dataset.readers} </div>
             {/*TODO hook this up to SAM?!?!?*/}
-            <div>Manage Viewers</div>
+            <div>
+              {dataset && dataset.id && (
+                <ManageUsersModal datasetUUID={dataset.id} dispatch={dispatch} />
```
thathert: We don't want this to be too tied to datasets or readers. I think we'd like to use this modal for managing studies/custodians + datasets/{custodians, readers, discoverers}. Instead of having it be connected/passing dispatch, let's have it accept callbacks for add/remove and the list of user emails as props.

src/components/DatasetCreateView.jsx
```diff
@@ -186,7 +186,7 @@ export class DatasetCreateView extends React.PureComponent {
                 <ManageUsers
                   {...props}
                   addReader={newEmail => this.addReader(newEmail)}
```
thathert: These should be more generic since we will use for more than just readers.

src/components/DatasetCreateView.jsx
```diff
@@ -186,7 +186,7 @@ export class DatasetCreateView extends React.PureComponent {
...
-                  defaultValue="Custodian Email Address"
+                  defaultValue="Add viewer email address"
```
thathert, line 189: Let's have this be override-able by a prop that's passed in and defaults to "Add email address" if nothing is provided.

src/components/ManageUsersModal.jsx
```diff
@@ -0,0 +1,138 @@
...
+    // TODO send this reader to SAM
+    // _.remove(newReaders, r => r === removeableEmail);
+    // dispatch();
+  }
+
+  render() {
+    const { classes } = this.props;
+    const { open } = this.state;
+    const modalText = 'Manage Viewers';
+    const readers = [
```
thathert: Let's have this and modalText be a prop that's passed in. We can have modalText default to 'Manage Users' if nothing is passed in for the prop.

src/components/ManageUsersModal.jsx
```diff
@@ -0,0 +1,138 @@
...
+            </Button>
+          </DialogActions>
+        </Dialog>
+      </div>
+    );
+  }
+}
+
+/* istanbul ignore next */
+function mapStateToProps(state) {
```
thathert: Like I mentioned above, I think this can be dis`connect`ed

thathert (changes requested):
This is a great start, I think there is some extra work to make this more generic that will save us time when we get to managing custodians.

thathert (approved):
Looks good, thanks!

PR #21, opened 2019-03-27
Pass auth token on axios calls

RoriCremer:
Saga fully tested

### DIFF
--- a/src/actions/user.js
+++ b/src/actions/user.js
@@ -8,6 +8,11 @@ import { createActions } from 'redux-actions';
 import { ActionTypes } from 'constants/index';
 
 export const { userLogin: logIn, userLogout: logOut } = createActions({
-  [ActionTypes.USER_LOGIN]: (name, image) => ({ name, image }),
+  [ActionTypes.USER_LOGIN]: (name, image, token, tokenExpiration) => ({
+    name,
+    image,
+    token,
+    tokenExpiration,
+  }),
   [ActionTypes.USER_LOGOUT]: () => ({}),
 });
--- a/src/sagas/repository.js
+++ b/src/sagas/repository.js
@@ -2,9 +2,10 @@
  * @module Sagas/App
  * @desc App
  */
-import { all, put, call, takeLatest } from 'redux-saga/effects';
+import { all, put, call, select, takeLatest } from 'redux-saga/effects';
 import { delay } from 'redux-saga';
 import axios from 'axios';
+import moment from 'moment';
 
 import { ActionTypes, STATUS } from 'constants/index';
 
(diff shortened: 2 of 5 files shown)


src/actions/user.js
```diff
@@ -8,6 +8,11 @@ import { createActions } from 'redux-actions';
 import { ActionTypes } from 'constants/index';
 
 export const { userLogin: logIn, userLogout: logOut } = createActions({
-  [ActionTypes.USER_LOGIN]: (name, image) => ({ name, image }),
+  [ActionTypes.USER_LOGIN]: (name, image, token, tokenExpiration) => ({
+    name,
```
thathert, line 12: email too?

src/sagas/repository.js
```diff
@@ -15,15 +16,46 @@ import { ActionTypes } from 'constants/index';
...
+const getToken = state => state.user.token;
+const getTokenExpiration = state => state.user.tokenExpiration;
+
+function* checkToken() {
+  const tokenExpiration = yield select(getTokenExpiration);
+  // if this fails, should isAuthenticated be flipped?
+  return moment(tokenExpiration.valueOf()).isSameOrAfter(moment().valueOf(), 'second');
+}
+
+function* authGet(url) {
```
thathert: We'll want a way for the user to pass in params as an object instead of putting them all into the URL

src/sagas/repository.js
```diff
@@ -15,15 +16,46 @@ import { ActionTypes } from 'constants/index';
  *
  */
 
+const getToken = state => state.user.token;
+const getTokenExpiration = state => state.user.tokenExpiration;
+
+function* checkToken() {
+  const tokenExpiration = yield select(getTokenExpiration);
+  // if this fails, should isAuthenticated be flipped?
+  return moment(tokenExpiration.valueOf()).isSameOrAfter(moment().valueOf(), 'second');
```
thathert: Out of the scope of this PR, but when a token has expired we need to either refresh it (if we can) or prompt the user to go through a flow to get a new one. I'm making a ticket to track.

thathert (changes requested):
Let's chat about a testing strategy for this. Also it's about time to get minikube set up for you

test/sagas/repository.spec.js
```diff
@@ -1,37 +1,157 @@
...
-  it('should have the expected watchers', done =>
-    expectSaga(repository)
-      .run({ silenceTimeout: true })
-      .then(() => {
-        done();
-      }));
-
-  it('should have the get datasets saga', () =>
+  it('test token expiration', () => {
+    const tokenExpiration = 1553537329627;
```
thathert, line 27: comment about the magic number? (what time is it? or explicitly make something that expires tomorrow)

test/sagas/repository.spec.js
```diff
@@ -1,37 +1,157 @@
...
+        dispatch: action => dispatched.push(action),
+        getState: () => ({ user: { tokenExpiration } }),
+      },
+      checkToken,
+    );
+    expect(test.result()).toBe(true);
+  });
+
+  it('test token invalid', () => {
+    const tokenExpired = moment() - 360000;
```
thathert, line 40: magic number

test/sagas/repository.spec.js
```diff
@@ -1,37 +1,157 @@
...
+        dispatch: action => dispatched.push(action),
+        getState: () => ({ user: { token, tokenExpiration } }),
+      },
+      authGet,
+    );
+    expect(test.result()).toEqual({ data: { job_status: 'succeeded' } });
+  });
+
+  it('test post with auth header', () => {
+    const token = 'Cherlene';
```
thathert, line 67: For both of these it would be nice to test that axios was getting the Authorization header set correctly

thathert (approved):
Looks good, couple minor comments but I think ready to go.

PR #23, opened 2019-04-02
preview: Ui testing

RoriCremer:
The dumbest ui test.

To be filled out so they are less dumb!


src/components/DatasetDetailView.jsx
```diff
@@ -78,7 +78,7 @@ export class DatasetDetailView extends React.PureComponent {
       },
     ];
     return (
-      <div className={classes.wrapper}>
+      <div id="datasetDetailView" className={classes.wrapper}>
```
thathert: `kebab-case` is preferred over camelCase

test/components/DatasetView.spec.js
```diff
@@ -0,0 +1,25 @@
...
+      <DatasetView {...ownProps} />
+    </Provider>,
+  );
+}
+
+describe('DatasetView', () => {
+  const wrapper = setup();
+
+  it('hey look a datasets view', () => {
+    wrapper.find('#datasets').simulate('click');
```
thathert: I think better to assert that the wrapper can find the element here. If we click something in a test we should verify that the component did something right when it got clicked.

thathert (approved):
Approving based on convo earlier. I'll leave it up to you if you'd like to make any changes.

PR #26, opened 2019-04-04
preview: additional testing

RoriCremer:


src/sagas/repository.js
```diff
@@ -15,15 +16,46 @@ import { ActionTypes, STATUS } from 'constants/index';
  *
  */
 
+export const getToken = state => state.user.token;
+export const getTokenExpiration = state => state.user.tokenExpiration;
+
+export function* checkToken() {
+  const tokenExpiration = yield select(getTokenExpiration);
+  // if this fails, should isAuthenticated be flipped?
```
thathert: if it fails we either need to refresh the token or prompt them to re-authenticate, not sure of the right method based on a little bit of research just now. If flipping isAuthenticated causes them to have to go back to the login screen I think that is a good first approach.

thathert (approved):
LGTM. Ready to merge once conflicts are resolved.

PR #28, opened 2019-04-05
Even more terra-fying

RoriCremer:
Screen Shot 2019-04-04 at 11 02 34 PM

Screen Shot 2019-04-05 at 2 13 28 PM

### DIFF
--- a/src/components/DatasetView.jsx
+++ b/src/components/DatasetView.jsx
@@ -1,12 +1,11 @@
 import React from 'react';
 import { connect } from 'react-redux';
-import moment from 'moment';
-import { Link, NavLink } from 'react-router-dom';
+import { NavLink } from 'react-router-dom';
 import PropTypes from 'prop-types';
 import { withStyles } from '@material-ui/core/styles';
 
 import { getDatasets } from 'actions/index';
-import JadeTable from './table/JadeTable';
+import DatasetTable from './table/DatasetTable';
 import AddSVG from '../../assets/media/icons/plus-circle-solid.svg';
 
 const styles = theme => ({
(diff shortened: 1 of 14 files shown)


src/components/DatasetView.jsx
```diff
@@ -1,39 +1,27 @@
...
-import { Link, NavLink } from 'react-router-dom';
 import PropTypes from 'prop-types';
 import { withStyles } from '@material-ui/core/styles';
 
 import { getDatasets } from 'actions/index';
-import JadeTable from './table/JadeTable';
-import AddSVG from '../../assets/media/icons/plus-circle-solid.svg';
+import DatasetTable from './table/DatasetTable';
 
 const styles = theme => ({
```
ddietterich, line 9: Is something like this shareable? It looks like these style thingies are the same or very similar to ones above.

ddietterich (commented):
I kinda get what this is doing, but only kinda. One question for my education.

src/components/table/StudyTable.jsx
```diff
@@ -0,0 +1,62 @@
...
+        render: row => <Link to={`/studies/details/${row.id}`}>{row.name}</Link>,
+      },
+      {
+        label: 'Description',
+        property: 'description',
+      },
+      {
+        label: 'Last changed',
+        property: 'modifiedDate',
+        render: row => moment(row.createdDate).fromNow(),
```
ahaessly, line 45: should this be row.modifiedDate?

src/sagas/repository.js
```diff
@@ -164,7 +161,7 @@ export function* addReaderToDataset({ payload }) {
   const datasetId = payload.datasetId;
   try {
     const response = yield call(
-      authGet,
+      authPost,
```
ahaessly, line 164: where do you pass the payload?

ahaessly (approved):
2 comments/questions

src/components/table/StudyTable.jsx
thathert: It will be eventually -- we need to add modified dates to the API.

thathert (approved):
LGTM -- will need to add payloads to authPost

PR #29, opened 2019-04-05
Router bug

RoriCremer:
Tabs not lighting up as expected with out this additional logic

tQ2yX4Gy6T


thathert (changes requested):
It looks like the dataset details view is showing up at the bottom of the dataset create form.

src/routes/Private.jsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { BrowserRouter, Link, Switch, Route } from 'react-router-dom';
```
thathert, line 2: It looks like my changes got dropped after we merged my branch into yours yesterday.

RoriCremer: I'll put up a pr with that as the diff

PR #30, opened 2019-04-07
Directional modal after dataset creation

RoriCremer:
Screen Shot 2019-04-06 at 2 35 21 PM

### DIFF
--- a/src/components/DatasetCreateView.jsx
+++ b/src/components/DatasetCreateView.jsx
@@ -111,6 +111,7 @@ export class DatasetCreateView extends React.PureComponent {
   selectStudy(study) {
     const { dispatch } = this.props;
     dispatch(actions.change('dataset.study', study));
+    // TODO then get the asset info
   }
 
   selectAsset(asset) {
@@ -257,9 +258,12 @@ export class DatasetCreateView extends React.PureComponent {
                 component={props => (
                   <MultiSelect
                     {...props}
+                    isDisabled={!study}
                     onChange={e => this.selectAsset(e.value)}
                     options={assetOptions}
-                    placeholder="Select Asset Type..."
+                    placeholder={
+                      study ? 'Select Asset Type...' : 'Select Study to Select Asset Type...'
+                    }
                     value={assetOptions.filter(option => option.value === asset)}
                   />
                 )}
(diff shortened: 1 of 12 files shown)


ahaessly:
when i have a PR for a branch that is not off of develop, i usually create the PR with the base branch being the one it depends on so diffs aren't included in both PRs.
also, i like the 3rd example above.

src/components/DatasetCreateView.jsx
```diff
@@ -119,9 +115,17 @@ export class DatasetCreateView extends React.PureComponent {
...
-    return studiesList;
+    const studyOptions = [];
+    studies.studies.map(study => studyOptions.push({ value: study.id, label: study.name }));
+    return studyOptions;
+  }
+
+  getAssetOptions(study) {
+    const assetOptions = [];
+    if (study && study.schema && study.schema.assets) {
+      study.schema.assets.map(asset => assetOptions.push({ value: asset.name, label: asset.name }));
```
thathert: `map` should only be used to create a new array. It feels strange to use it to mutate something like `assetOptions` here instead of taking a more functional approach. Either `forEach` should be used here, or we should switch assetOptions to use `let` above and then have this line be
```javascript
assetOptions = study.schema.assets.map(asset => ({ value: asset.name, label: asset.name }));
```
(I prefer using `map`)

src/components/DatasetCreateView.jsx
```diff
@@ -119,9 +115,17 @@ export class DatasetCreateView extends React.PureComponent {
   }
 
   getStudyOptions(studies) {
-    const studiesList = [];
-    studies.studies.map(study => studiesList.push({ value: study.id, label: study.name }));
-    return studiesList;
+    const studyOptions = [];
+    studies.studies.map(study => studyOptions.push({ value: study.id, label: study.name }));
```
thathert: Similar comment as below (let's switch to just assigning the result of the map to studyOptions)

src/components/DatasetDirectionalModal.jsx
```diff
@@ -0,0 +1,180 @@
...
+      <div id="dataset-modal">
+        {success ? (
+          <Dialog
+            onClose={this.handleClose}
+            aria-labelledby="customized-dialog-title"
+            open={open}
+            maxWidth={false}
+          >
+            <DialogTitle className={classes.dialogTitle} id="customized-dialog-title">
+              <div className={classes.closetitle}>Where to next?</div>
```
thathert: nit: I know I suggested this text so feel free to throw something at me, but the use of next twice in this modal seems excessive now that I see it -- let's just have this be "Where to?"

thathert (approved):
Approving with the expectation that comments are addressed or discussed before merging

PR #33, opened 2019-04-10
Add Terra branding

RoriCremer:
Screen Shot 2019-04-10 at 12 56 41 PM

I'm not sure how we want to handle the config file:
I'm unclear on how whitelisting will work and it would be good to know how much "Terra" info should be hardcoded or put into the config file for changes later on


src/sagas/repository.js
```diff
@@ -157,12 +158,15 @@ export function* getDatasetPolicy({ payload }) {
   }
 }
 
-export function* addReaderToDataset({ payload }) {
-  const datasetId = payload.datasetId;
+export function* addReadersToDataset({ payload }) {
+  // TODO is this going to need a loop for the list?
```
thathert: Yes, one call per reader in the list.

thathert (changes requested):
Replied to your TODO, otherwise after making that change this should be good.

thathert (approved):

PR #41, opened 2019-04-20
error toast

RoriCremer:
Will we have non-error alerts too?

It's not storing the errors--that will need to added as separate storage

dB8u9njEc3


src/components/DatasetPreviewView.jsx
```diff
@@ -64,18 +64,7 @@ const styles = theme => ({
...
-      '"Segoe UI"',
-      'Roboto',
-      '"Helvetica Neue"',
-      'Arial',
-      'sans-serif',
-      '"Apple Color Emoji"',
-      '"Segoe UI Emoji"',
-      '"Segoe UI Symbol"',
-    ].join(','),
+    fontFamily: theme.typography.fontFamily,
```
thathert, line 67: Thank you.

src/components/Toast.jsx
```diff
@@ -0,0 +1,81 @@
...
+    dispatch(hideAlert());
+  }
+
+  render() {
+    const { classes, errorMsg } = this.props;
+    return (
+      <Card className={classes.card}>
+        <CardContent className={classes.content}>
+          <ErrorIcon className={classes.icon} alt="logo" />
+          <div className={classes.text}>{errorMsg || 'WTF did you do?'}</div>
```
thathert: "An error occurred, please try again or submit a bug report." will probably look better in a demo :)

thathert (approved):
Looks good so far. Approving with expectation that alerts can be dismissed by id and that that message gets updated

src/components/Toast.jsx
RoriCremer: sorry--it was a Saturday so I couldn't help myself.

ahaessly (approved):
👍

PR #43, opened 2019-04-23
DR-336: add terra link

RoriCremer:
waEHFeURr9


thathert (approved):
LGTM. Have you verified that going through the terra registration with a new identity allows you to use the app? (Does it make SAM not throw 401s with the new identity?)

PR #50, opened 2019-05-20
Paging, sort, filter etc

RoriCremer:
Currently not using react virtualize--but a placeholder until then
lNOoIlA32s

### DIFF
--- a/src/components/table/DatasetTable.jsx
+++ b/src/components/table/DatasetTable.jsx
@@ -40,7 +40,7 @@ class DatasetTable extends React.PureComponent {
   static propTypes = {
     classes: PropTypes.object.isRequired,
     datasetCount: PropTypes.number,
-    datasets: PropTypes.object.isRequired,
+    datasets: PropTypes.array.isRequired,
     dispatch: PropTypes.func.isRequired,
     summary: PropTypes.bool,
   };
@@ -95,7 +95,7 @@ class DatasetTable extends React.PureComponent {
           handleFilter={this.handleFilterDatasets}
           rows={datasets}
           summary={summary}
-          totalCount={datasetCount || datasets.length}
+          totalCount={datasetCount}
         />
       </div>
     );
--- a/src/components/table/JadeTable.jsx
+++ b/src/components/table/JadeTable.jsx
@@ -120,7 +120,7 @@ export class JadeTable extends React.PureComponent {
   render() {
     const { classes, columns, rows, summary, totalCount } = this.props;
     const { orderBy, orderDirection, page, rowsPerPage } = this.state;
-    const emptyRows = rowsPerPage - Math.min(rowsPerPage, 8 - page * rowsPerPage);
+    const emptyRows = rowsPerPage - Math.min(rowsPerPage, totalCount - page * rowsPerPage);
     return (
       <div>
         {!summary && (
@@ -143,7 +143,7 @@ export class JadeTable extends React.PureComponent {
             <JadeTableHead
               columns={columns}
               onRequestSort={this.handleRequestSort}
-              order={orderDirection}
+              orderDirection={orderDirection}
               orderBy={orderBy}
             />
             <TableBody>
--- a/src/components/table/StudyTable.jsx
+++ b/src/components/table/StudyTable.jsx
@@ -39,7 +39,7 @@ class StudyTable extends React.PureComponent {
   static propTypes = {
     classes: PropTypes.object.isRequired,
     dispatch: PropTypes.func.isRequired,
-    studies: PropTypes.object.isRequired,
+    studies: PropTypes.array.isRequired,
     summary: PropTypes.bool,
     studyCount: PropTypes.number,
     studyListName: PropTypes.string,
@@ -90,7 +90,7 @@ class StudyTable extends React.PureComponent {
           rows={studies}
           handleFilter={this.handleFilterStudies}
           summary={summary}
-          totalCount={studyCount || studies.length}
+          totalCount={studyCount}
         />
       </div>
     );


src/components/table/DatasetTable.jsx
```diff
@@ -37,11 +39,24 @@ const styles = theme => ({
 class DatasetTable extends React.PureComponent {
   static propTypes = {
     classes: PropTypes.object.isRequired,
-    rows: PropTypes.array,
+    datasetCount: PropTypes.number,
+    datasets: PropTypes.object.isRequired,
```
thathert: Should this be an array?

src/components/table/DatasetTable.jsx
```diff
@@ -75,10 +90,23 @@ class DatasetTable extends React.PureComponent {
             <AddSVG className={classes.plusButton} />
           </NavLink>
         </div>
-        <JadeTable columns={columns} rows={rows} />
+        <JadeTable
+          columns={columns}
+          handleFilter={this.handleFilterDatasets}
+          rows={datasets}
+          summary={summary}
+          totalCount={datasetCount || datasets.length}
```
thathert: I'd prefer if we just used the count, it should always be included from the API.

src/components/table/JadeTable.jsx
```diff
@@ -4,98 +4,184 @@ import { withStyles } from '@material-ui/core/styles';
...
+  };
 
-    this.setState({ order: newOrder, orderBy: property });
+  handleSearchString = event => {
+    const { handleFilter } = this.props;
+    const { page, orderDirection, orderBy, rowsPerPage } = this.state; // limit
+    const offset = page * rowsPerPage;
+    const searchString = event.target.value;
+    this.setState({ searchString }); // filter
+    handleFilter(rowsPerPage, offset, orderBy, orderDirection, searchString);
```
thathert: This is not specific to just filtering -- let's call it something like `handleEnumeration`?

src/components/table/JadeTable.jsx
```diff
@@ -4,98 +4,184 @@ import { withStyles } from '@material-ui/core/styles';
...
   };
 
   render() {
-    const { classes, columns, rows } = this.props;
-    const { order, orderBy } = this.state;
+    const { classes, columns, rows, summary, totalCount } = this.props;
+    const { orderBy, orderDirection, page, rowsPerPage } = this.state;
+    const emptyRows = rowsPerPage - Math.min(rowsPerPage, 8 - page * rowsPerPage);
```
thathert: magic number

src/components/table/JadeTable.jsx
```diff
@@ -4,98 +4,184 @@ import { withStyles } from '@material-ui/core/styles';
...
+              onChange={this.handleSearchString}
+            />
+          </div>
+        )}
+        <Paper className={classes.root}>
+          <Table className={classes.table}>
+            <JadeTableHead
+              columns={columns}
+              onRequestSort={this.handleRequestSort}
+              order={orderDirection}
```
thathert: Let's make this `orderDirection={orderDirection}`

I'm getting: 
```
The prop `orderDirection` is marked as required in `JadeTableHead`, but its value is `undefined`.
```

src/components/table/JadeTable.jsx
```diff
@@ -4,98 +4,184 @@ import { withStyles } from '@material-ui/core/styles';
...
+                <TableRow key={row.id} className={classes.row}>
+                  {columns.map(col => (
+                    <TableCell key={col.property}>
+                      {col.render ? col.render(row) : row[col.property]}
+                    </TableCell>
+                  ))}
+                </TableRow>
+              ))}
+              {rows.length < rowsPerPage && (
+                <TableRow style={{ height: 50 * emptyRows }}>
```
thathert: magic number

src/components/table/StudyTable.jsx
```diff
@@ -58,22 +73,35 @@ class StudyTable extends React.PureComponent {
...
     return (
       <div className={classes.wrapper}>
         <div className={classes.header}>{studyListName || 'STUDIES'}</div>
-        <JadeTable columns={columns} rows={rows} />
+        <JadeTable
+          columns={columns}
+          rows={studies}
+          handleFilter={this.handleFilterStudies}
+          summary={summary}
+          totalCount={studyCount || studies.length}
```
thathert: We should always have studyCount.

src/sagas/repository.js
```diff
@@ -136,10 +136,17 @@ export function* createDataset() {
...
+  const offset = payload.offset || 0;
+  const limit = payload.limit || 5;
+  const filter = payload.searchString || '';
+  const sort = payload.sort || 'created_date';
+  const direction = payload.direction || 'desc';
   try {
-    const response = yield call(authGet, '/api/repository/v1/datasets');
+    const response = yield call(
+      authGet,
+      `/api/repository/v1/datasets?offset=${offset}&limit=${limit}&sort=${sort}&direction=${direction}&filter=${filter}`,
```
thathert: Can't we pass an object with params into authGet/axios instead of doing the string interpolation ourselves here?

thathert (changes requested):
Looks good! I have a few requests here. Also can we remove the <div> that is inside of the `<Switch>` component so we stop getting the computedMatch error? Thanks :)

src/components/table/JadeTable.jsx
```diff
@@ -4,98 +4,184 @@ import { withStyles } from '@material-ui/core/styles';
...
+                  <TableCell colSpan={columns.length} />
+                </TableRow>
+              )}
+            </TableBody>
+          </Table>
+          {!summary && (
+            <TablePagination
+              rowsPerPageOptions={[5, 10, 25]}
```
ahaessly: magic numbers?

ahaessly (approved):
👍

src/components/table/JadeTable.jsx
RoriCremer: do I want to pass the styling theme into the react component?

src/components/table/JadeTable.jsx
RoriCremer: These feel like standard paging choices to me--but you're right, we could enum them somewhere.

RoriCremer:
The rest of the comments will be addressed in a follow on pr

thathert (approved):

src/components/table/JadeTable.jsx
thathert: I don't think so unless we need it inside. I was just thinking of defining this as `ROW_HEIGHT` or something up above.

PR #52, opened 2019-05-24
Styling punch up

RoriCremer:
http://jade-rc.datarepo-dev.broadinstitute.org/

Re the bug fix:
I got too greedy on the cleanup...this passes the dataset request information to the created dataset object/store so that the preview page can show the appropriate info

### DIFF
--- a/src/components/table/DatasetTable.jsx
+++ b/src/components/table/DatasetTable.jsx
@@ -10,7 +10,7 @@ import JadeTable from './JadeTable';
 
 const styles = theme => ({
   jadeLink: {
-    color: theme.palette.primary.main,
+    color: theme.palette.common.link,
     textDecoration: 'none',
     '&:hover': {
       textDecoration: 'underline',
--- a/src/components/table/DatasetTable.jsx
+++ b/src/components/table/DatasetTable.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Link, NavLink } from 'react-router-dom';
+import { Link } from 'react-router-dom';
 import moment from 'moment';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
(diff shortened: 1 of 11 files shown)


src/sagas/repository.js
```diff
@@ -148,6 +148,7 @@ export function* getDatasets({ payload }) {
       authGet,
       `/api/repository/v1/datasets?offset=${offset}&limit=${limit}&sort=${sort}&direction=${direction}&filter=${filter}`,
     );
+    console.log(response);
```
thathert: extra console.log

src/components/table/DatasetTable.jsx
```diff
@@ -83,13 +65,7 @@ class DatasetTable extends React.PureComponent {
...
     ];
     return (
-      <div className={classes.wrapper}>
-        <div className={classes.header}>
-          DATASETS
-          <NavLink to="/datasets/create">
-            <AddSVG className={classes.plusButton} />
-          </NavLink>
-        </div>
+      <div>
```
thathert, line 64: What's this div for? I think we can remove it

src/components/table/JadeTable.jsx
```diff
@@ -76,7 +76,7 @@ export class JadeTable extends React.PureComponent {
     handleFilter: PropTypes.func,
     rows: PropTypes.arrayOf(PropTypes.object),
     summary: PropTypes.bool,
-    totalCount: PropTypes.number.isRequired,
+    totalCount: PropTypes.number, // TODO why getting undefined sometimes?
```
thathert: Are you pulling this from the response? It could have been from branches that didn't have the new counting code. It should never be undefined in the response. When you initialize this is there a default value you're passing in from the store?

src/components/table/StudyTable.jsx
```diff
@@ -83,8 +65,7 @@ class StudyTable extends React.PureComponent {
       },
     ];
     return (
-      <div className={classes.wrapper}>
-        <div className={classes.header}>{studyListName || 'STUDIES'}</div>
+      <div>
```
thathert: I think we can remove the div

thathert (approved):
Approving with a couple minor points.

PR #53, opened 2019-05-30
Empty rows bug

RoriCremer:
Two bug fixes:

Add safety rails in case there are no studies or datasets

I got too greedy on the cleanup...this passes the dataset request information to the created dataset object/store so that the preview page can show the appropriate info

### DIFF
--- a/src/components/DatasetPreviewView.jsx
+++ b/src/components/DatasetPreviewView.jsx
@@ -94,7 +94,7 @@ const styles = theme => ({
 export class DatasetPreviewView extends React.PureComponent {
   static propTypes = {
     classes: PropTypes.object.isRequired,
-    createdDataset: PropTypes.object,
+    createdDataset: PropTypes.object, // TODO is this even?
     createdDatasets: PropTypes.arrayOf(PropTypes.object),
     dispatch: PropTypes.func.isRequired,
     exception: PropTypes.bool.isRequired,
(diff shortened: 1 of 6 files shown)


src/components/DatasetPreviewView.jsx
```diff
@@ -155,12 +155,12 @@ export class DatasetPreviewView extends React.PureComponent {
                 </div>
               ) : (
                 <div className={classes.values}>
-                  {createdDataset && createdDataset.dataset.name}
+                  {createdDataset && createdDataset.datasetRequest.name}
```
ahaessly: do we need to check createdDataset.datasetRequest here and on line 163?

src/sagas/repository.js
```diff
@@ -148,6 +148,7 @@ export function* getDatasets({ payload }) {
       authGet,
       `/api/repository/v1/datasets?offset=${offset}&limit=${limit}&sort=${sort}&direction=${direction}&filter=${filter}`,
     );
+    console.log(response);
```
ahaessly: remove debug message

ahaessly (approved):
2 comments

src/components/DatasetPreviewView.jsx
RoriCremer: hmmmm--yes. It feels more likely that createdDataset would be the thing that gets lost, but no reason not to have the extra safety rails

thathert (approved):
LGTM. yes I think createdDataset is needed.

PR #54, opened 2019-05-30
Paging follow-ons

RoriCremer:
Small changes based around pr review for paging.

Still to do:
"There is a way to pass an object of params
instead of interpolating all of the params into the string manually -- check the docs"


thathert (approved):
LGTM, can you add a follow on ticket in JIRA?

ddietterich (approved):
LGTM FWIW

PR #59, opened 2019-06-12
add enter keydown function

RoriCremer:
be able to hit return on the "add viewers" part of the dataset creation rather than have to explicitly hit the button


thathert (approved):
LGTM

PR #60, opened 2019-06-13
Jobid datasetid bugfix

RoriCremer:
Fix on a bug where after a dataset is created, the job id (from the dataset creation) is used instead of the dataset id of the newly created dataset


thathert (approved):
LGTM

PR #61, opened 2019-06-14
Share a study with custodians

RoriCremer:
kFxa15Hecy


src/sagas/repository.js
```diff
@@ -274,18 +274,77 @@ export function* getStudyById({ payload }) {
...
+    yield put({
+      type: ActionTypes.EXCEPTION,
+      payload: err,
+    });
+  }
+}
+
+export function* addCustodianToStudy({ payload }) {
+  const studyId = payload.studyId;
+  const custodian = payload.users[0];
```
ahaessly, line 295: will the UI ever want to add multiple users in one call?

src/actions/app.js
```diff
@@ -66,6 +50,41 @@ export const { removeReaderFromDataset } = createActions({
...
+  [ActionTypes.GET_STUDY_BY_ID_SUCCESS]: study => study,
+});
+
+export const { getStudyPolicy } = createActions({
+  [ActionTypes.GET_STUDY_POLICY]: policy => policy,
+  [ActionTypes.GET_STUDY_POLICY_SUCCESS]: policy => policy,
+});
+
+export const { addCustodianToStudy } = createActions({
+  [ActionTypes.ADD_CUSTODIAN_TO_STUDY]: (studyId, users) => ({
```
ahaessly, line 75: similar to comment below - but if we are only going to process one user at a time, should we be passing users here?

ahaessly (approved):
looks good

src/actions/app.js
RoriCremer: I'm going to bring up during standup tomorrow -- wonder if this makes sense as a follow on.
The current UX is just to do one at a time, and I think my desire to keep this parallel to datasets is getting in the way

PR #91, opened 2019-08-02
DR-487 remove client id hardcode and switch UI to use the /config endpoint

myessail:


test/components/WelcomeView.spec.js
```diff
@@ -30,6 +40,7 @@ describe('WelcomeView', () => {
 
     expect(mockGoogleLogin).not.toBeNull();
     expect(mockGoogleLogin.length).toEqual(1);
-    expect(mockDispatch).toHaveBeenCalledWith({ payload: {}, type: 'USER_LOGIN' });
+    // expect(mockDispatch).toHaveBeenCalledWith({ payload: {}, type: 'USER_LOGIN' });
```
thathert, line 43: What happened here? It'd be nice to keep this since it is more specific than just `toHaveBeenCalled`.

thathert (approved):
LGTM -- one request at the bottom to fix a test or get rid of some commented out code.

PR #102, opened 2019-08-22
DR-541 Use automatic node version switching in the UI

myessail:
requires that you have nvm installed (not through homebrew)


thathert (approved):
LGTM assuming README gets updated with directions

RoriCremer (approved):

PR #105, opened 2019-08-29
DR-513 filter panel

thathert:


src/sagas/repository.js
```diff
@@ -379,7 +379,6 @@ export function* runQuery({ payload }) {
     const url = `https://bigquery.googleapis.com/bigquery/v2/projects/${payload.projectId}/queries`;
     const body = {
       query: payload.query,
-      maxResults: payload.maxResults,
```
myessail: I'm not sure if this should be removed, I think we need this to set the paging limit?

myessail (approved):

PR #106, opened 2019-08-30
DR-536 Connect facet panel to query runner

myessail:


src/components/dataset/query/QueryViewSidebarItem.jsx
```diff
@@ -0,0 +1,42 @@
...
+export class QueryViewSidebarItem extends React.PureComponent {
+  static propTypes = {
+    column: PropTypes.object,
+    handleChange: PropTypes.func,
+  };
+
+  handleChange = event => {
+    const { column, handleChange } = this.props;
+
+    const idAndValue = {
```
thathert: maybe change to `nameAndValue`?

src/components/dataset/query/QueryViewTable.jsx
```diff
@@ -91,39 +92,10 @@ export class QueryViewTable extends React.PureComponent {
...
-        pageSize = numRows;
-      }
 
-      options = {
-        pageSize,
-        pageSizeOptions: [pageSize],
-        showFirstLastPageButtons: false,
-      };
-    }
+    const bigquery = new BigQuery({});
```
thathert: Let's make the default in the constructor an empty object so that we don't need to pass it in here.

I *think* it's something like: `constructor(options = {}) { ... }`

src/modules/bigquery.js
```diff
@@ -0,0 +1,117 @@
...
+    this.pageTokenMap = pageTokenMap;
+  }
+
+  pageData = (query, queryResults, columns, token) =>
+    new Promise(resolve => {
+      let rawData = {};
+      const data = [];
+
+      const { jobId } = queryResults.jobReference;
+      const { projectId } = queryResults.jobReference;
```
thathert: You can say `const { jobId, projectId } = queryResults.jobReference;` to unpack both variables in one line

thathert (commented):
Looking good so far

thathert (approved):
LGTM

RoriCremer (approved):

PR #107, opened 2019-09-10
DR-521 Add Google Analytics

myessail:


src/index.jsx
```diff
@@ -13,6 +14,12 @@ import { store } from 'store/index';
 import config from 'config';
 import App from 'containers/App';
 
+ReactGA.initialize('UA-147471357-1', {
```
thathert: I think we should include both the `isProd` flag and this UA string in the `/configuration` endpoint. What do `debug` and `gaOptions/cookieDomain` do?

thathert (approved):
LGTM

myessail:
Closing until we come to a consensus on what we want to do with this

PR #109, opened 2019-09-16
DR-569 Hit BigQuery to get possible ranges for numeric columns

myessail:
rangefilter

### DIFF
--- a/src/modules/bigquery.js
+++ b/src/modules/bigquery.js
@@ -118,9 +118,9 @@ export default class BigQuery {
     return '';
   };
 
-  getColumnMin = (columnName, dataset, tableName, token) => {
+  getColumnMinMax = (columnName, dataset, tableName, token) => {
     const url = `https://bigquery.googleapis.com/bigquery/v2/projects/${dataset.dataProject}/queries`;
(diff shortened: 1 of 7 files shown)


src/modules/bigquery.js
```diff
@@ -117,4 +117,44 @@ export default class BigQuery {
     }
     return '';
   };
+
+  getColumnMin = (columnName, dataset, tableName, token) => {
+    const url = `https://bigquery.googleapis.com/bigquery/v2/projects/${dataset.dataProject}/queries`;
+    const query = `SELECT MIN(${columnName}) FROM [${dataset.dataProject}.datarepo_${dataset.name}.${tableName}]`;
```
ddietterich: I am concerned about getting each statistic with a separate query. I'd like to understand the performance and cost effects of that. You can easily get min and max of a column in a single query and that would be 2x better. Given that BigQuery scans by column, doing each column one at a time maybe doesn't matter for cost, but you do get the queuing delay from running multiple jobs and marshaling the BQ resources for each.

myessail: you're right, moving this to get min/max in one query

RoriCremer (approved):

thathert (approved):
LGTM

PR #122, opened 2019-11-01
DR-632 Filter panel can get out of sync with the snapshot card

myessail:


thathert:
Looks good except for one issue I just found with the free-text boxes. Steps to reproduce:

1. Go to query builder for ancestry_specific_meta_analysis.
2. Expand phenotype, type BMI, click apply filters
3. Click the (x) next to the phenotype in the snapshot card.

BMI stays in the input field, despite being cleared out of the card. Also, if I click Apply Filters at this point, nothing happens. I think the input should be cleared out if the (x) is clicked.

RoriCremer (approved):
Im wondering if we are going to want to do this as two states in the main store after all? This feels like as we get more complex it may get messier.

thathert (approved):
LGTM

PR #123, opened 2019-11-04
DR-627 Applying a filter should update the other filter panels

myessail:


src/components/dataset/query/sidebar/filter/CategoryWrapper.jsx
```diff
@@ -25,16 +25,31 @@ export class CategoryWrapper extends React.PureComponent {
...
+      );
+    }
+  }
+
   render() {
     const { values } = this.state;
     const { column, filterData, handleChange } = this.props;
 
-    if (values.length <= 10) {
+    if (values && values.length <= 10) {
```
RoriCremer: why does this need to be less than/equal to 10?

RoriCremer (approved):

thathert:
Let's fix the text-becomes-checkbox thing :)

PR #126, opened 2019-11-22
DR-660 Add commas to Large numbers on the UI

myessail:


RoriCremer (approved):

src/modules/bigquery.js
```diff
@@ -64,25 +68,12 @@ export default class BigQuery {
...
-          title: colData.name,
-          field: colData.name,
-        };
-
-        columns.push(col);
-      });
-    }
-    return columns;
+  commaFormatted = amount => {
+    return amount.replace(/\B(?=(\d{3})+(?!\d))/g, ',');
```
thathert: Let's use this built-in instead of a regex: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat

`return new Intl.NumberFormat('en-US', { maximumSignificantDigits: 3 }).format(amount);`

This function defaults `useGrouping` to true, which specifies "whether to use grouping separators, such as thousands separators or thousand/lakh/crore separators."

thathert (changes requested):
Requested one change (that will also help with the significant digit ticket).

thathert (approved):
👍

PR #132, opened 2019-11-29
DR-677 Limit number of significant digits to 4 in the query table

myessail:


src/modules/bigquery.js
```diff
@@ -72,6 +74,9 @@ export default class BigQuery {
     return new Intl.NumberFormat('en-US').format(amount);
   };
 
+  significantDigits = amount =>
+    new Intl.NumberFormat('en-US', { maximumSignificantDigits: 4 }).format(amount);
```
thathert: Let's make this 3

thathert (approved):
Approved assuming we switch from 4 -> 3 significant digits.

src/modules/bigquery.js
myessail: ask and you shall receive

PR #134, opened 2019-12-02
DR-678 Michael reorganizes the QueryView

myessail:
The organization of the QueryView component was bothering me so I decided to refactor it to completely use the grid system -- end result looks the same as before, but no more ugly scrollbar overlap on the right!


thathert (approved):
LGTM, especially happy about removing the vendor-specific `-webkit-*` prefix :)

RoriCremer (approved):

PR #135, opened 2019-12-02
fix login and deep linking when refreshing

thathert:


RoriCremer (approved):

src/actions/app.js
```diff
@@ -117,11 +117,6 @@ export const { hideAlert } = createActions({
   [ActionTypes.HIDE_ALERT]: index => index,
 });
 
-export const { getConfiguration } = createActions({
-  [ActionTypes.GET_CONFIGURATION_SUCCESS]: configuration => configuration,
-  [ActionTypes.GET_CONFIGURATION]: configuration => configuration,
```
myessail, line 122: I think this is the only place we're using these action types, might as well remove them from the enum?

myessail (approved):

src/index.jsx
```diff
@@ -75,11 +75,11 @@ export const app = {
...
           <Helmet
             defer={false}
-            htmlAttributes={{ lang: 'pt-br' }}
+            htmlAttributes={{ lang: 'en-us' }}
             encodeSpecialCharacters={true}
             defaultTitle={'Terra Data Repository' || config.title}
             titleTemplate={`%s | ${config.name}`}
-            titleAttributes={{ itemprop: 'name', lang: 'pt-br' }}
+            titleAttributes={{ itemprop: 'name', lang: 'en-us' }}
           />
```
RoriCremer: coisa loca!

src/actions/app.js
thathert: We are using the SUCCESS, but not the regular one so I removed it.

PR #141, opened 2020-01-02
DR-634 Ability to hit enter to apply filters for free-text entry

myessail:


src/components/dataset/query/sidebar/filter/FreetextFilter.jsx
```diff
@@ -20,6 +21,13 @@ export class FreetextFilter extends React.PureComponent {
 
   getOptions = values => (values == null ? [] : values.map(a => a.f[0].v));
 
+  handleReturn = event => {
+    const { handleFilters } = this.props;
+    if (event.key === 'Enter') {
```
RoriCremer, line 26: are we going to be using this a lot? should we have an enum or const for it?

RoriCremer (approved):

PR #149, opened 2020-01-15
DR-727: Separate each filtered field in the query panel

mackenziekerwin:


src/components/dataset/query/sidebar/QuerySidebarPanel.jsx
```diff
@@ -28,33 +28,51 @@ export class QuerySidebarPanel extends React.PureComponent {
...
       if (Array.isArray(data)) {
-        dataString = _.join(data, ', ');
+        dataString = data.map(datum => {
+          return (
+            <ListItemText>
+              {datum}
+              <Button onClick={() => this.clearFilter(filter, datum)}>
+                <HighlightOff />
+              </Button>
+            </ListItemText>);
```
myessail: is there a way we can reuse this since it's returning the same case for the array and plain object cases? maybe we can set dataString based on whether or not it's an array and then we can just `map` over it once?

myessail (approved):

src/components/dataset/query/sidebar/QuerySidebarPanel.jsx
thathert: Yeah I think we can assign something like `const list = Array.isArray(data) ? data : _.keys(data);` and then map over that if data is not a string? Would DRY this up nicely.

thathert (changes requested):
Looks good! 1 suggestion

myessail (approved):

thathert (approved):
LGTM

thathert (approved):
LGTM

PR #196, opened 2020-02-05
DR-747 Indirect filtering

myessail:


src/modules/bigquery.js
```diff
@@ -128,4 +150,105 @@ export default class BigQuery {
...
+        neighbors[u] = [];
+      }
+      neighbors[u].push(v);
+      if (neighbors[v] === undefined) {
+        // Also add the edge v -> u in order
+        neighbors[v] = []; // to implement an undirected graph.
+      } // For a directed graph, delete
+      neighbors[v].push(u); // these four lines.
+    });
+
```
RoriCremer, line 172: These comments feel a little disjointed

RoriCremer (approved):

src/modules/bigquery.js
myessail: heh whoops

src/components/dataset/query/QueryView.jsx
```diff
@@ -47,9 +47,11 @@ export class QueryView extends React.PureComponent {
   static propTypes = {
     classes: PropTypes.object,
     dataset: PropTypes.object,
-    datasetPolicies: PropTypes.object,
+    datasetPolicies: PropTypes.array,
```
fboulnois, line 50: object to array?

src/modules/bigquery.js
```diff
@@ -95,12 +112,15 @@ export default class BigQuery {
...
-    const query = `SELECT MIN(${columnName}) AS min, MAX(${columnName}) AS max FROM [${dataset.dataProject}.datarepo_${dataset.name}.${tableName}]`;
+    const query = `SELECT MIN(${columnName}) AS min, MAX(${columnName}) AS max FROM \`${dataset.dataProject}.datarepo_${dataset.name}.${tableName}\``;
 
     return axios
       .post(
         url,
-        { query },
+        {
+          query,
+          useLegacySql: false,
```
fboulnois, line 122: is this new param required for the framework, for BQ, or for something else?

src/polyfills.js
```diff
@@ -4,6 +4,5 @@
  * @desc Add support for older browsers.
  */
 import '@babel/polyfill';
-import 'react-app-polyfill/ie11';
```
fboulnois, line 7: was this an automatic change or manual? is there a directive regarding which browsers we need to support?

src/modules/bigquery.js
```diff
@@ -128,4 +150,105 @@ export default class BigQuery {
...
+          }
+          path.push(u);
+          path.reverse();
+          return path;
+        }
+        predecessor[v] = u;
+        queue.push(v);
+      }
+    }
+  };
```
fboulnois, line 209: 👍 bfs!

fboulnois (commented):
lgtm, does this require my explicit approval?

src/components/dataset/query/QueryView.jsx
myessail: this was wrong the first time, I was getting console errors that it's actually an array so I changed the type

src/modules/bigquery.js
myessail: this is required for BQ, the backtics around the table name are specific to their "standard SQL" implementation which is enabled by setting `useLegacySql` to false

src/polyfills.js
myessail: this was a manual change, it was causing an error during deploy time because of a dependencies issue. I chatted with @thathert about it and we decided not to support IE since most people are using something else these days (chrome/firefox) and IE11 is reaching EoS from microsoft as they move towards the chromium based version of Edge

src/components/dataset/query/sidebar/QuerySidebarPanel.jsx
```diff
@@ -49,47 +52,53 @@ export class QuerySidebarPanel extends React.PureComponent {
...
+    const listTables = _.keys(filterData).map(table => {
+      const listFilters = _.keys(filterData[table]).map(filter => {
+        const data = _.get(filterData[table], filter);
+        let dataString = data.value;
+        if (data.type === 'range') {
+          dataString = (
+            <Chip
+              onDelete={() => this.clearFilter(table, filter)}
+              className={classes.inline}
+              label={_.join(data.value, ' \u2013 ')}
```
thathert: Can you make this a constant and describe what that escape sequence means?

PR #198, opened 2020-02-10
DR-757 include names of tables in filter panel

mackenziekerwin:
collapse

### DIFF
--- a/src/components/dataset/query/sidebar/AppliedFilterList.jsx
+++ b/src/components/dataset/query/sidebar/AppliedFilterList.jsx
@@ -1,0 +1,164 @@
+import React from 'react';
+import _ from 'lodash';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { withStyles } from '@material-ui/core/styles';
+
+import Card from '@material-ui/core/Card';
+import List from '@material-ui/core/List';
+import ListSubheader from '@material-ui/core/ListSubheader';
+import ListItem from '@material-ui/core/ListItem';
+import ListItemText from '@material-ui/core/ListItemText';
+import { applyFilters } from '../../../../actions';
+import { Chip, Typography, Collapse, Badge, Avatar } from '@material-ui/core';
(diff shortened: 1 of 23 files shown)


src/components/dataset/query/sidebar/AppliedFilterList.jsx
```diff
@@ -0,0 +1,166 @@
...
+    const listFilters = _.keys(filterData[table]).map(filter => {
+      const data = _.get(filterData[table], filter);
+      let dataString = data.value;
+
+      if (data.type === 'range') {
+        dataString = (
+          <Chip
+            onDelete={() => this.clearFilter(table, filter)}
+            className={classes.inline}
+            label={_.join(data.value, ' \u2013 ')}
```
myessail: I had \u2013 pulled out as a constant named emDash which might have gotten lost in the merge conflicts, it might be a good idea to do that again here so it's clearer what the unicode is for

myessail (approved):

src/components/dataset/query/sidebar/AppliedFilterList.jsx
```diff
@@ -0,0 +1,166 @@
...
+    if (clonedFilter.type !== 'range') {
+      if (Array.isArray(filterValue)) {
+        filterValue.splice(filterValue.indexOf(datum), 1);
+      }
+      if (_.isPlainObject(filterValue)) {
+        delete filterValue[datum];
+      }
+    }
+
+    if (_.isEmpty(filterValue) || clonedFilter.type === 'range') {
```
fboulnois, line 84: If `clonedFilter.type === 'range'`, what is `filterValue`?

fboulnois (commented):
lgtm

src/components/dataset/query/sidebar/AppliedFilterList.jsx
mackenziekerwin: It's an array, with `filterValue[0]` being the lower bound of the range and `filterValue[1]` equal to the upper bound. I think it comes from `handleChange` on our range slider component. In this case we don't want to be able to delete just one item from the array, so we remove the entire filter. Hope that makes sense!

PR #199, opened 2020-02-11
DR-743 show number of results on the filter panel

mackenziekerwin:
number-of-results

### DIFF
--- a/src/components/dataset/query/sidebar/QuerySidebarPanel.jsx
+++ b/src/components/dataset/query/sidebar/QuerySidebarPanel.jsx
@@ -10,11 +10,12 @@ import ListSubheader from '@material-ui/core/ListSubheader';
 import ListItem from '@material-ui/core/ListItem';
 import ListItemText from '@material-ui/core/ListItemText';
 import { applyFilters } from '../../../../actions';
-import { Chip } from '@material-ui/core';
+import { Chip, Typography, Collapse } from '@material-ui/core';
+import AppliedFilterList from './AppliedFilterList';
 
 const styles = () => ({
-  listHeader: {
-    paddingTop: '16px',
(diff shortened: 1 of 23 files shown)


myessail (approved):

fboulnois (approved):

src/components/dataset/query/sidebar/QuerySidebarPanel.jsx
```diff
@@ -29,18 +20,28 @@ export class QuerySidebarPanel extends React.PureComponent {
...
+    const { classes, filterData, selected, results, polling } = this.props;
     const listTables = _.keys(filterData).map(table => {
       return <AppliedFilterList table={table} selected={selected} />;
     });
 
+    const resultsLabel = results == 1 ? 'Result' : 'Results';
+
     return (
       <Card>
-        <List subheader={<ListSubheader component="div">Properties</ListSubheader>}>
```
fboulnois, line 43: Will users still need this text somewhere?

fboulnois (commented):
lgtm

src/components/dataset/query/sidebar/QuerySidebarPanel.jsx
mackenziekerwin: I think we decided yesterday in the design meeting that it wasn't necessary, especially since it looks weird when there are no filters applied. We can add it back in somewhere else if Jerome wants it.

mackenziekerwin: I think we decided yesterday in the design meeting that it wasn't necessary, especially since it looks weird when there are no filters applied. We can add it back in somewhere else if Jerome wants it.

PR #206, opened 2020-02-24
DR-790 Clear filter data

myessail:
clearbutton
clearcontext

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "classlist-polyfill": "^1.2.0",
     "classnames": "^2.2.6",
     "clsx": "^1.0.4",
+    "connected-react-router": "^6.7.0",
     "date-fns": "2.0.0-beta.2",
     "ease-component": "^1.0.0",
     "events-polyfill": "^2.1.2",
@@ -46,8 +47,8 @@
     "http-proxy-middleware": "^0.20.0",
     "immutability-helper": "^3.0.1",
     "jest-cli": "^24.9.0",
-    "jsonfile": "^4.0.0",
     "js-combinatorics": "^0.5.5",
+    "jsonfile": "^4.0.0",
     "jss-plugin-extend": "^10.0.3",
     "lodash": "^4.17.15",
     "mobile-detect": "^1.4.4",
@@ -240,7 +241,7 @@
   },
   "husky": {
     "hooks": {
-      "post-merge": "node tools update"
+      "post-merge": "npm install"
     }
   },
   "browserslist": [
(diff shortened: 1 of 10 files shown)


mackenziekerwin (approved):

package.json
```diff
@@ -240,7 +241,7 @@
   },
   "husky": {
     "hooks": {
-      "post-merge": "node tools update"
+      "post-merge": "npm install"
```
fboulnois, line 244: Why did you need to switch?

src/components/dataset/query/sidebar/filter/CategoryFilter.jsx
```diff
@@ -24,14 +24,14 @@ export class CategoryFilter extends React.PureComponent {
...
-    count: PropTypes.string,
+    count: PropTypes.number,
     filterData: PropTypes.object,
     handleChange: PropTypes.func,
     name: PropTypes.string,
     table: PropTypes.string,
   };
 
-  componentWillReceiveProps(nextProps) {
+  UNSAFE_componentWillReceiveProps(nextProps) {
```
fboulnois: Is the UNSAFE tag something that was added automatically? What caused this change?

fboulnois (changes requested):

src/components/dataset/query/sidebar/filter/CategoryFilter.jsx
thathert: It's recommended we switch to [getDerivedStateFromProps](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) instead of this. Let's try it out with this PR.

src/components/dataset/query/sidebar/filter/CategoryFilterGroup.jsx
```diff
@@ -51,7 +51,7 @@ export class CategoryFilterGroup extends React.PureComponent {
   render() {
     const { column, filterData, values, table, originalValues } = this.props;
     const checkboxes = _.keys(originalValues).map(name => {
-      const count = values.hasOwnProperty(name) ? values[name] : 0;
+      const count = values.hasOwnProperty(name) ? parseInt(values[name], 10) : 0;
```
thathert, line 54: 👍

thathert (changes requested):
I like the approach, but the `UNSAFE_` makes me sad.

package.json
myessail: this is updating packages every time i switch branches, which is becoming more of an annoyance than something helpful. we're going to start auto-merging dependabot PRs once we have end to end tests (next sprint) so I'm not too concerned about it

fboulnois (approved):
lgtm

thathert (commented):
Thanks for fixing!

PR #242, opened 2020-03-12
Add e2e tests (cypress edition)

myessail:


cypress/fixtures/example.json
```diff
@@ -1 +1,5 @@
-{}
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
```
fboulnois: We probably want to change this or remove the file, right?

Dockerfile
```diff
@@ -1,13 +1,3 @@
-FROM node:12
-WORKDIR /app
-COPY package* ./
-RUN npm install
-COPY . .
-RUN npm run build --production
-
 FROM nginx:1.15.10-alpine
```
fboulnois, line 1: Is the nginx container proxying requests?

fboulnois (approved):
Right now this is just testing rendering, correct? A few other questions:

Dockerfile
thathert: It does -- check `ops/nginx.conf`

cypress/integration/queryBuilder.spec.js
```diff
@@ -0,0 +1,32 @@
+describe('test query builder', () => {
+  beforeEach(() => {
+    cy.server();
+
+    cy.route('GET', 'api/repository/v1/datasets/**').as('getDataset');
```
thathert, line 5: What does `**` do here

cypress/integration/queryBuilder.spec.js
```diff
@@ -0,0 +1,32 @@
...
+    cy.contains('V2F_GWAS_Summary_Stats').should('be.visible');
+    cy.contains('V2F_GWAS_Summary_Stats').click();
+    cy.wait(['@getDataset', '@getDatasetPolicies']);
+    cy.get('[data-cy=queryDatasetButton]').click();
+  });
+
+  it('does render', () => {});
+
+  it('applies filters', () => {
+    cy.get('div.MuiButtonBase-root:nth-child(2) > svg:nth-child(1)').click();
```
thathert, line 25: For selectors like this that aren't semantic like the ones below, it'd be useful to add a comment above about what is actually getting selected.

src/routes/HeadlessLogin.jsx
```diff
@@ -0,0 +1,54 @@
...
+    this.setState({
+      token: event.target.value,
+    });
+  };
+
+  handleLoginButton = () => {
+    const { dispatch } = this.props;
+    const { token } = this.state;
+
+    dispatch(logIn('user.name', 'user.imageUrl', 'user.email', token, 1615064728000));
```
thathert: So that our tests don't start failing out of the blue next year, let's calculate this dynamically? Something like a day into the future should be good.

thathert (commented):
Looks good! A few nits

smark88 (approved):

smark88:
Let's squash like 70 commits and merge this.

cypress/integration/queryBuilder.spec.js
myessail: it's a wildcard in the url, since the uuid in this request is different in my local env vs. what's running in integration

PR #267, opened 2020-04-03
mk DR-880 share snapshot during creation 

mackenziekerwin:
share-panel
One major change:
This share panel slides out in the query builder to add readers _before_ creating the snapshot (basically edits `snapshot.readers` in the store).
And one little change:
The `ADD_SNAPSHOT_POLICY_MEMBER` action that uses the api endpoint now supports adding discoverers as well. I realized after making the change that this is different from sharing before snapshot creation, but I kept those changes in because this is probably getting implemented soon.


cypress/integration/queryBuilder.spec.js
```diff
@@ -30,4 +30,18 @@ describe('test query builder', () => {
     cy.get('[data-cy="applyFiltersButton"]').click();
     cy.get('[data-cy=appliedFilterList-ancestry_specific_meta_analysis]', {}).should('be.visible');
   });
+
+  it.only('adds/removes readers', () => {
```
myessail: make sure you remove the `.only` before you merge

package-lock.json
```diff
@@ -3107,7 +3107,7 @@
         "source-map": {
           "version": "0.6.1",
           "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.6.1.tgz",
-          "integrity": "sha512-UjgapumWlbMhkBgzT7Ykc5YXUT46F0iKu8SGXq0bcwP5dz/h0Plj6enJqjz1Zbq2l5WaqYnrVbwWOWMyF3F47g==",
+          "integrity": "sha1-dHIq8y6WFOnCh6jQu95IteLxomM=",
```
myessail: i would remove all changes to this file (`git checkout develop package-lock.json`), it looks like we didn't make any package changes here so no need to change the lockfile

src/components/dataset/query/sidebar/panels/ShareSnapshot.jsx
```diff
@@ -0,0 +1,247 @@
...
+
+  /**
+   * adds the email as a tag to the text box
+   */
+  inputEmail = (event, value) => {
+    this.setState({ usersToAdd: value, currentInput: '' });
+  };
+
+  /**
+   * updates the snapshot readers to include users that have been entered
```
myessail, line 107: yay comments!!

myessail (approved):
Looks great! Few comments before you merge:

thathert (commented):
LGTM!

fboulnois (approved):
lgtm 😃

cypress/integration/queryBuilder.spec.js
```diff
@@ -30,4 +30,18 @@ describe('test query builder', () => {
...
   });
+
+  it('adds/removes readers', () => {
+    // selects the share button in the sidebar
+    cy.get('div.MuiButtonBase-root:nth-child(3) > svg:nth-child(1)').click();
+    cy.get('[data-cy=enterEmailBox]').type('mkerwin.dev@gmail.com');
+    cy.get('[data-cy=inviteButton]').click();
+    cy.get('[data-cy=readers]')
+      .contains('mkerwin.dev@gmail.com')
+      .should('be.visible');
```
thathert, line 41: How about a test to type in a comma between emails and see if you get two entries?

thathert (approved):
One request

cypress/integration/queryBuilder.spec.js
mackenziekerwin: done!

PR #290, opened 2020-04-08
MK DR-891 Add description box to the 'save snapshot' panel

mackenziekerwin:
Screen Shot 2020-04-08 at 12 42 07 PM
I didn't make too many changes since we're gonna want to move this component to its own panel in the near future. I just adjusted the size for now and added the text box.
Also fixed a quick bug that was preventing the save button from pushing to the snapshots page!

### DIFF
--- a/src/components/dataset/query/sidebar/panels/CreateSnapshotPanel.jsx
+++ b/src/components/dataset/query/sidebar/panels/CreateSnapshotPanel.jsx
@@ -1,5 +1,4 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import React, { useState } from 'react';
 import clsx from 'clsx';
 import { withStyles } from '@material-ui/core/styles';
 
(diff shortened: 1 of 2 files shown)


src/components/dataset/query/sidebar/panels/CreateSnapshotPanel.jsx
```diff
@@ -23,40 +22,60 @@ const styles = theme => ({
...
-export class CreateSnapshotPanel extends React.PureComponent {
-  static propTypes = {
-    classes: PropTypes.object,
-    handleCreateSnapshot: PropTypes.func,
-    handleSaveSnapshot: PropTypes.func,
-  };
-
-  render() {
-    const { classes, handleCreateSnapshot, handleSaveSnapshot } = this.props;
+function CreateSnapshotPanel(props) {
```
thathert: Why the switch to a functional component instead of class-based? It's good for us to remain consistent if we can.

mackenziekerwin: I thought since the component doesn't do much in terms of changing the state, it would be cleaner and more concise to use the `useState` hook (I liked how it could be called directly from `onChange`). It would be fine to revert back to a class component; I would just have to write some additional functions.

src/components/dataset/query/sidebar/panels/CreateSnapshotPanel.jsx
```diff
@@ -23,40 +22,60 @@ const styles = theme => ({
...
+  return (
+    <div className={clsx(classes.rowTwo, classes.saveButtonContainer)}>
+      <TextField
+        id="snapshotName"
+        label="Name"
+        variant="outlined"
+        size="small"
+        fullWidth
+        className={classes.textField}
+        onChange={event => setName(event.target.value)}
```
thathert: Could the class-based version of this be `event => this.setState('name', event.target.value)` ?

mackenziekerwin: I tried that, and it didn't seem to work. It would only work if I passed in a function that basically wrapped `setState`. I can look into it again because I don't remember the exact error but I could have just misinterpreted it

thathert: Ah that might have to do with how `this` works inside of functions. One workaround is to declare `const self = this;` at the top of the render function and then inside of the `onChange` callback say `self.setState(...)`.

Useful read: https://stackoverflow.com/questions/3127429/how-does-the-this-keyword-work

src/components/dataset/query/sidebar/panels/CreateSnapshotPanel.jsx
```diff
@@ -23,40 +22,60 @@ const styles = theme => ({
   cancelButton: {
     alignSelf: 'flex-end',
   },
+  textField: {
+    backgroundColor: theme.palette.common.white,
+    borderRadius: '5px',
```
RoriCremer, line 28: can we use a theme.spacing for these?

mackenziekerwin: Yeah, we could use `theme.spacing(0.5)`, which would be 4px. This only applies to the white background, though. The `variant="outlined"` prop is what gives the text fields the grey border, which gets all its styling from default MUI classes, so that border radius would still be 5px.

mackenziekerwin: It might be useful to make a reusable Jade text field component with its own custom attributes? It would let us be consistent with Terra and we wouldn't have to constantly override MUI styling

myessail (approved):

fboulnois (approved):
lgtm, I like the idea of a reusable text field suggested earlier.

PR #336, opened 2020-04-17
DR-912 Exclude array values from query

myessail:


RoriCremer (approved):

fboulnois (approved):
lgtm, a short description like "Enables queries to exclude specific columns" would be useful.

PR #369, opened 2020-04-23
DR-930 Count query results

mackenziekerwin:
count-results
Since the query builder is more of a preview of the snapshot about to be created, we limit the results to 1000 rows. This PR updates the number displayed in the snapshot card to reflect all the results matching the query, not just the first 1000 returned.


src/components/dataset/query/QueryView.jsx
```diff
@@ -99,6 +107,15 @@ export class QueryView extends React.PureComponent {
           PAGE_SIZE,
         ),
       );
+      const bq = new BigQuery();
+      bq.getTotalRows(
+        dataset.dataProject,
+        `#standardSQL
+          SELECT COUNT(*) FROM \`${dataset.dataProject}.datarepo_${dataset.name}.${selected}\` AS ${selected}
```
danxmoran: I would suggest `COUNT(1)` instead of `COUNT(*)`. `COUNT(*)` will scan every column in the table. `COUNT(1)` won't scan anything for counting, so you're only paying for the columns mentioned in the joins and filters.

src/components/dataset/query/QueryView.jsx
```diff
@@ -99,6 +107,15 @@ export class QueryView extends React.PureComponent {
...
+          ${joinStatement}
+          ${filterStatement}`,
+        token,
+      ).then((response) => dispatch(countResults(parseFloat(response))));
```
myessail: i would use `parseInt` here instead since this isn't going to be a decimal number

src/modules/bigquery.js
```diff
@@ -110,6 +111,26 @@ export default class BigQuery {
...
+        },
+        {
+          headers: {
+            'Content-Type': 'application/json',
+            Authorization: `Bearer ${token}`,
+          },
+        },
+      )
+      .then((response) => response.data.rows[0].f[0].v);
+  };
```
myessail: i think if we're putting this in the store, we should probably move this call to `repository.js` to be consistent

src/actions/app.js
```diff
@@ -9,14 +9,14 @@ import { createActions } from 'redux-actions';
 import { ActionTypes } from 'constants/index';
 
 export const { createSnapshot } = createActions({
-  [ActionTypes.CREATE_SNAPSHOT]: snapshot => snapshot,
-  [ActionTypes.CREATE_SNAPSHOT_JOB]: snapshot => snapshot,
-  [ActionTypes.CREATE_SNAPSHOT_SUCCESS]: snapshot => snapshot,
-  [ActionTypes.CREATE_SNAPSHOT_FAILURE]: snapshot => snapshot,
+  [ActionTypes.CREATE_SNAPSHOT]: (snapshot) => snapshot,
```
fboulnois, line 12: Are these changes due to linting?

fboulnois (commented):
I agree with a few of the earlier comments, one other question:

src/actions/app.js
mackenziekerwin: Yes, there was a recent dependabot that updated the linter

myessail (approved):

PR #385, opened 2020-05-05
mk DR-721 add buttons to use filter panel to exclude values

mackenziekerwin:
See frames [Filter 1–Filter 1.9](https://projects.invisionapp.com/d/main#/projects/prototypes/19763083) in InVision.


mackenziekerwin:
new-stuff
Here's what the new changes look like. Each column in the filter panel has its own 'apply filters' button--which used to live at the bottom next to 'create snapshot'--as well as a 'reset filter' button.

src/components/dataset/query/sidebar/QueryViewSidebarItem.jsx
```diff
@@ -17,15 +27,28 @@ export class QueryViewSidebarItem extends React.PureComponent {
     token: PropTypes.string,
   };
 
-  handleChange = value => {
-    const { column, handleChange, tableName } = this.props;
+  componentDidUpdate(prevProps, prevState) {
```
fboulnois, line 40: Can you add some comments in the code on why these functions are necessary? At a high level I understand that by default the reset button is disabled, but when a filter is applied, this updates the filterMap which in turn updates the state of the reset button to enabled. Is this correct?

src/components/dataset/query/sidebar/QueryViewSidebarItem.jsx
```diff
@@ -39,37 +62,56 @@ export class QueryViewSidebarItem extends React.PureComponent {
...
-            tableName={tableName}
-            token={token}
-          />
-        );
-      default:
-        return <div />;
-    }
+    const { disableButton, filterMap } = this.state;
+    const item =
+      column.datatype === 'string' ? (
```
fboulnois: If I understand correctly, if the column returns strings, you get a filter where the rows are strings, otherwise if it's numeric you get a filter where the rows are selectable ranges?

fboulnois (commented):
This approach looks reasonable to me! A few questions and comments:

src/components/dataset/query/sidebar/QueryViewSidebarItem.jsx
mackenziekerwin: Yes, that's right! However, at the design meeting yesterday we decided to omit the reset button, so I will update this and add some comments.

src/components/dataset/query/sidebar/QueryViewSidebarItem.jsx
mackenziekerwin: Yes! The datatype of the column determines the type of input component used to filter the data. The user can select specific string values using checkboxes or entering them into a text box, whereas numeric values are filtered with a range slider.

mackenziekerwin:
exclude-values
New changes:
- Lists of filter values can be pasted
- Values are rendered under the text box in a list
- Toggle whether values are to be included/excluded

src/components/dataset/query/sidebar/filter/FreetextFilter.jsx
```diff
@@ -21,36 +36,88 @@ export class FreetextFilter extends React.PureComponent {
...
+    if (value.includes(',')) {
+      selections = value.split(',');
+    } else if (value.includes(' ')) {
+      selections = value.split(' ');
+    } else {
+      return;
+    }
+    const trimmed = _.map(selections, _.trim);
+    const nonEmpty = _.filter(trimmed, (t) => t !== '');
+    handleChange(nonEmpty);
```
snf2ye, line 52: I think this can be simplified a bit:  Using regex should return a trimmed array of strings split at " " and ",". Then you can just check if the length of the array is greater than 1: 
`const selections = value.split(/[ ,]+/)`

src/components/dataset/query/sidebar/QueryViewSidebarItem.jsx
```diff
@@ -17,60 +35,98 @@ export class QueryViewSidebarItem extends React.PureComponent {
...
       column,
       dataset,
-      filterData,
       filterStatement,
       joinStatement,
-      handleFilters,
       tableName,
       token,
     } = this.props;
-    switch (column.datatype) {
```
snf2ye, line 42: Is there a reason why you're not using the switch statement anymore? To me, the switch statement feels a bit easier to read.

snf2ye (commented):
Looks good to me, just a couple of thoughts! I haven't worked with lodash, so this was a good learning experience to read though! Thanks! 😄

src/components/dataset/query/sidebar/filter/FreetextFilter.jsx
mackenziekerwin: I like that a lot better. Thanks for the tip!!

src/components/dataset/query/sidebar/QueryViewSidebarItem.jsx
mackenziekerwin: I couldn't find a way to conditionally assign a component to `item` using a switch, and then I forgot to look into it further (my bad). I just did a quick search and I found I could do this:
```
    const item = ((datatype) => {
      switch (datatype) {
        case 'string':
          return <CategoryWrapper ... />;
        case 'float':
        case 'integer':
          return <RangeFilter ... />;
        default:
          return <div />;
      }
    })(column.datatype);
```
So I can make that edit since I agree it's easier to read.

src/components/dataset/query/sidebar/filter/FreetextFilter.jsx
```diff
@@ -43,20 +43,11 @@ export class FreetextFilter extends React.PureComponent {
...
-      selections = value.split(' ');
-    } else {
-      return;
-    }
-    const trimmed = _.map(selections, _.trim);
-    const nonEmpty = _.filter(trimmed, (t) => t !== '');
-    handleChange(nonEmpty);
+    event.preventDefault();
+    const text = event.clipboardData.getData('text');
+    handleChange(text.split(/[ ,\n]+/));
```
snf2ye: I *think* you still need to filter out the empty strings (like you did in the 'nonEmpty' variable above), for the case that someone ends with a comma or space.

myessail: +1, might be good to pull this regex out to a const and give it a name so it's a bit more readable as well

myessail (approved):

snf2ye (approved):

PR #491, opened 2020-06-04
DR-1056 asset dropdown

mackenziekerwin:
create-snapshot
The UI includes a dropdown to select an asset, which is necessary for snapshot creation. We had previously hardcoded the first asset in the schema to be used every time.

### DIFF
--- a/src/components/dataset/query/JadeDropdown.jsx
+++ b/src/components/dataset/query/JadeDropdown.jsx
@@ -14,13 +14,14 @@ const styles = () => ({
 export class JadeDropdown extends React.PureComponent {
   static propTypes = {
     classes: PropTypes.object,
+    name: PropTypes.string,
     onSelectedItem: PropTypes.func,
     options: PropTypes.array,
     value: PropTypes.string,
   };
 
   render() {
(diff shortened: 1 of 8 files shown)


RoriCremer (approved):

src/components/dataset/query/JadeDropdown.jsx
```diff
@@ -0,0 +1,51 @@
...
+    onSelectedItem: PropTypes.func,
+    options: PropTypes.array,
+    value: PropTypes.string,
+  };
+
+  render() {
+    const { classes, options, onSelectedItem, value, name } = this.props;
+
+    return (
+      <form autoComplete="off">
```
fboulnois, line 18: It's not clear to me what this is rendering. Could you comment on this?

fboulnois (approved):
lgtm, one question:

src/components/dataset/query/JadeDropdown.jsx
mackenziekerwin: It renders a dropdown component, like the "Select Asset" box in the gif. I added a comment to the render method!

PR #520, opened 2020-06-15
DR-1060 store cleanup

mackenziekerwin:
This PR addresses some tech debt surrounding the Redux store. A 'snapshot' reducer previously acted as a form for storing fields necessary to create a snapshot, but those fields have been moved to the 'snapshots' state so they can be displayed in components such as the popup modal and the sidebar panels.


RoriCremer (approved):

src/reducers/query.js
```diff
@@ -75,6 +75,7 @@ export default {
       },
       [LOCATION_CHANGE]: (state, action) => {
         if (!action.payload.location.pathname.includes('/query')) {
+          // michael can you help us with this
```
fboulnois: Do you still need help on this? :)

fboulnois (commented):
Quick question, otherwise nice cleanup, lgtm:

src/reducers/query.js
mackenziekerwin: Yes, we needed @myessail's attention so we could ask about that! I'm going to merge for now and we can make it a part of the next PR.

PR #551, opened 2020-06-24
DR-1023 update snapshot creation modal

mackenziekerwin:
confirmation-dialog
This PR makes several updates to the dialog that opens when a snapshot is created.
Changes:
- Loading dialog opens on top of the Query Builder while a snapshot is being created
- If snapshot creation fails, dialog closes on the Query Builder, revealing the error box
- If snapshot is successfully created, page navigates to the snapshot list and dialog displays a snapshot summary
- Snapshot query statement is fixed to join correctly on the root table of the selected asset
- [DR-909](https://broadworkbench.atlassian.net/browse/DR-909) is fixed so sidebar doesn't hide error box

### DIFF
--- a/src/containers/App.jsx
+++ b/src/containers/App.jsx
@@ -30,7 +30,7 @@ import SignOutSVG from '../../assets/media/icons/logout-line.svg';
 
 const drawerWidth = 240;
 
-const useStyles = makeStyles(theme => ({
+const useStyles = makeStyles((theme) => ({
   root: {
     display: 'flex',
     fontFamily: theme.typography.fontFamily,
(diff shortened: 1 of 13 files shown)


RoriCremer (approved):

src/containers/App.jsx
```diff
@@ -135,7 +135,7 @@ const useStyles = makeStyles(theme => ({
     position: 'absolute',
     right: 0,
     width: theme.spacing(40),
-    zIndex: '100',
+    zIndex: 1201,
```
fboulnois, line 138: This seems like a pretty large z-index, any particular reason why? Also, should it numeric or a string?

fboulnois (approved):
Great job! One question:

src/containers/App.jsx
mackenziekerwin: It needs to be higher than the z-index for the sidebar panel, which is 1200 (default Material UI classes)! I think either a number or string works but I made it a number to be consistent with the other CSS rules.

PR #552, opened 2020-06-29
DR-783 type decimals in range filter

mackenziekerwin:
range-slider
The TextField part of the range filter is tricky because the value gets inputted as text but stored as a number for our filtering purposes. We currently parse out the number as the text gets entered, which is problematic for decimals, since `parseFloat('0.')` returns `0`.

It turns out that filters on our columns of type 'integer' or 'float' don't even need to be stored in the `filterData` as numbers. We get the min and max values from BigQuery as strings, parse out the numbers, then put the filters back into another query as a string. This PR takes out that middle step and keeps the range values as strings, allowing the user to freely enter text into the range filters. Number validation is left at a higher level, and the "apply filters" button disables for bad inputs.


fboulnois (approved):
great job! lgtm

myessail (approved):

PR #554, opened 2020-07-01
DR-1000 Autocomplete lag in filter panels

myessail:


src/modules/bigquery.js
```diff
@@ -159,6 +160,40 @@ export default class BigQuery {
...
+    dataset,
+    tableName,
+    token,
+    filterStatement,
+    joinStatement,
+  ) => {
+    const url = `https://bigquery.googleapis.com/bigquery/v2/projects/${dataset.dataProject}/queries`;
+    const filterOrEmpty = _.isEmpty(filterStatement)
+      ? `WHERE ${columnName} LIKE '%${currText}%'`
+      : `WHERE ${filterStatement} AND ${columnName} LIKE '%${currText}%'`;
```
fboulnois: Will this kick off for every character entered in the filter panel?

fboulnois (commented):
One question:

src/modules/bigquery.js
myessail: nope, i'm calling `getAutocompleteForColumnDebounced` in the component which will fire this off after the user stops typing for 500ms

fboulnois: Nice, that's awesome.

fboulnois (approved):
lgtm

src/modules/bigquery.js
```diff
@@ -159,6 +160,40 @@ export default class BigQuery {
...
+    const joinOrEmpty = _.isEmpty(joinStatement) ? '' : joinStatement;
```
mackenziekerwin: Is this just equal to `joinStatement`? I believe it is a string so in both cases `joinOrEmpty` and `joinStatement` are equal

src/modules/bigquery.js
mackenziekerwin: I think when `filterStatement` is not empty it already starts with "WHERE". Would this cause a parsing error for BigQuery?

mackenziekerwin (approved):
Nice job, I just have some quick comments:

src/modules/bigquery.js
myessail: not quite, if there is no `joinStatement` (it hasn't been created in the store), it'll get put in as `undefined`

mackenziekerwin: Ok, got it!!

PR #555, opened 2020-07-02
DR-929 properly align columns in dataset tables

mackenziekerwin:
Before:
Screen Shot 2020-07-02 at 12 59 14 PM
After:
Screen Shot 2020-07-02 at 12 59 41 PM

Fixes a bug that omits table cells when the value for that column is `null`. Also fixes null integer and float values that are instead being rendered as `0`.


fboulnois (approved):
lgtm

myessail (approved):

PR #559, opened 2020-07-08
DR-913 Back button is broken after navigating to dataset page

myessail:
So, we had some code that was overriding history.listen and adding some extra properties that we're not actually using. On chrome, this is causing 5 entries to be added to the history for every link you click -- so the back button wasn't broken, there were just extra pages. I think all of these entries had different key values, and somehow firefox was able to determine that they were all actually the same page which is why it isn't broken there.

We aren't actually using any of this extra stuff so I'm just going to get rid of it.

### DIFF
--- a/src/modules/hist.js
+++ b/src/modules/hist.js
@@ -1,24 +1,6 @@
-// @flow
 import { createBrowserHistory } from 'history';
 
-import qs from 'qs';
-
 const history = createBrowserHistory();
-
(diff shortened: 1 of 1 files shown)


snf2ye (approved):

mackenziekerwin (approved):

fboulnois (approved):
makes sense, lgtm

PR #560, opened 2020-07-08
DR-1072 format timestamps as a readable date format

mackenziekerwin:
Columns of datatype 'TIMESTAMP' were being formatted as floats instead of dates, so this PR adds formatting to have them read like a date.


myessail (approved):
Thank you MacKenzie, very cool!

fboulnois (approved):
Makes sense to me!

PR #564, opened 2020-07-24
DR-940 error/delay notifications

mackenziekerwin:
error-toast
This PR makes changes to how we display errors and indicate slow behavior in the UI. Currently, we keep a list of active notifications in the store and use actions to hide and show them in a familiar "toast" component. To make these errors more consistent with the different interfaces across Terra, this PR adopts the `react-notifications-component` package that is currently used in the Terra UI. This library allows our errors to behave in a less confusing way--dismissible _and_ able to disappear after a set number of seconds.
Additionally, this PR adds a message to the loading screen in a dataset when a BigQuery query has been polling for longer than 10 seconds.


src/components/table/JadeTable.jsx
```diff
@@ -119,7 +123,7 @@ export class JadeTable extends React.PureComponent {
   };
 
   render() {
-    const { classes, queryResults, columns, rows, polling } = this.props;
+    const { classes, queryResults, columns, rows, polling, delay } = this.props;
```
RoriCremer, line 126: Not specific to this pr, but could we consider a norm on the UI to order these alphabetically as we do with the static propTypes list? I think it would help with readability

mackenziekerwin: sure, we can start fixing those from now on

RoriCremer (approved):

fboulnois (commented):
We can probably `git ignore` the .DS_Store files. Otherwise I think it seems reasonable.

src/components/table/JadeTable.jsx
myessail: i'll see if i can add something in eslint for it!

src/sagas/repository.js
```diff
@@ -38,7 +41,7 @@ export function* authGet(url, params = {}) {
       headers: { 'Content-Type': 'application/json', Authorization: `Bearer ${token}` },
     });
   }
-  return false;
+  throw timeoutMsg;
```
myessail, line 44: 🎉

myessail (approved):

PR #566, opened 2020-07-29
DR-945 Snapshot details page lists too many datasets

mackenziekerwin:
The details page for a given snapshot includes a table that should list all the datasets included in the snapshot. Instead, this table lists every dataset in the repo.

It turns out that the snapshot details page is passing down the rows to be included in the dataset table, but those rows are being overwritten with the `<DatasetTable/>`'s own `dispatch(getDatasets())`. So, I've taken this method out of the `<DatasetTable/>` component so the rows can be passed down from the parent component. I've also updated all references of `<DatasetTable/>` (and `<SnapshotTable/>`, to be consistent) to behave the same way.

This also meant I had to relocate the methods for filtering and sorting dataset/snapshot tables, and I took this functionality out of the summary tables.

### DIFF
--- a/src/components/SnapshotDetailView.jsx
+++ b/src/components/SnapshotDetailView.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import _ from 'lodash';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { withStyles } from '@material-ui/core/styles';
@@ -13,7 +14,6 @@ import {
 import DetailViewHeader from './DetailViewHeader';
 
 import DatasetTable from './table/DatasetTable';
-import _ from 'lodash';
 
 const styles = (theme) => ({
   wrapper: {
--- a/src/components/SnapshotDetailView.jsx
+++ b/src/components/SnapshotDetailView.jsx
@@ -13,6 +13,7 @@ import {
 import DetailViewHeader from './DetailViewHeader';
 
 import DatasetTable from './table/DatasetTable';
+import _ from 'lodash';
 
 const styles = (theme) => ({
   wrapper: {
@@ -53,6 +54,13 @@ const styles = (theme) => ({
 });
 
 export class SnapshotDetailView extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = {
+      filteredDatasets: null,
+    };
+  }
+
   static propTypes = {
     classes: PropTypes.object.isRequired,
     dispatch: PropTypes.func.isRequired,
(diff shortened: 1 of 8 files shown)


src/components/SnapshotDetailView.jsx
```diff
@@ -89,8 +97,18 @@ export class SnapshotDetailView extends React.PureComponent {
     dispatch(removeCustodianFromSnapshot(snapshot.id, removeableEmail));
   };
 
+  handleFilterDatasets = (limit, offset, sort, sortDirection, searchString) => {
```
mackenziekerwin, line 100: Note: `handleFilterDatasets` for `SnapshotDetailView` is different from that in `DatasetView` because each snapshot has its own list of datasets as part of the snapshot object, so filtering and sorting are just manipulations on that array.

src/components/table/LightTableHead.jsx
```diff
@@ -49,19 +50,23 @@ export class LightTableHead extends React.PureComponent {
...
-                <Tooltip
-                  title="Sort"
-                  placement={col.numeric ? 'bottom-end' : 'bottom-start'}
-                  enterDelay={300}
-                >
-                  <TableSortLabel
-                    active={orderBy === col.property}
-                    direction={orderDirection}
-                    onClick={this.createSortHandler(col.property)}
+                {summary ? (
```
mackenziekerwin, line 53: This is where the tables allow sorting on the columns. For summary tables (like in the Home view), the columns are not clickable buttons.

myessail (approved):

RoriCremer (approved):

src/components/SnapshotDetailView.jsx
```diff
@@ -89,8 +97,18 @@ export class SnapshotDetailView extends React.PureComponent {
...
+    const { snapshot } = this.props;
+    let datasets = snapshot.source.map((s) => s.dataset);
+    datasets = datasets.filter((d) => d.name.toLowerCase().includes(searchString.toLowerCase()));
+    datasets = _.orderBy(datasets, sort, sortDirection);
+    datasets = _.take(_.drop(datasets, offset), limit);
```
RoriCremer: would this be more readable as a chain rather than a reassignment?

mackenziekerwin: It might be most readable to separate each method out on a new line, but instead of reassigning, give a more descriptive label to each array: 
```
const filtered = datasets.filter((d) => d.name.toLowerCase().includes(searchString.toLowerCase()));
const sorted = _.orderBy(filtered, sort, sortDirection);
const paged = _.take(_.drop(sorted, offset), limit);
```

PR #582, opened 2020-08-03
DR-1150 Script to ingest datasets from one environment to another

myessail:


tools/ingest.js
```diff
@@ -0,0 +1,238 @@
...
+      })
+      .option('verbose', {
+        alias: 'v',
+        desc: 'print extra information',
+        default: false,
+        type: 'boolean',
+      }),
+);
+
+const CORE_BILLING_ACCOUNT_ID = '00708C-45D19D-27AAFA';
```
fboulnois, line 80: Will this vary between environments? Also, say an HCA person wants to transfer data from dev to prod, who should pay for that transfer?

tools/ingest.js
```diff
@@ -0,0 +1,238 @@
...
+
+if (argv.verbose) {
+  axios.interceptors.request.use((request) => {
+    console.log('Starting Request', request);
+    return request;
+  });
+}
+
+const agent = new https.Agent({
+  // rejectUnauthorized: false,
```
fboulnois, line 90: Just checking that this is the desired value here!

tools/ingest.js
```diff
@@ -0,0 +1,238 @@
...
+        console.error(stderr);
+      }
+      resolve(data);
+    });
+  });
+}
+
+function switchAccount(email) {
+  return new Promise((resolve) => {
+    call(`gcloud config set account ${email} && gcloud auth print-access-token`).then((data) => {
```
fboulnois, line 110: Hrm, do we need to do some extra validation here to make sure something like `person@gmail.com; cat /etc/password` can't be passed in here?

fboulnois (commented):
A few questions, otherwise lgtm!

snf2ye (approved):
Thanks for walking through this today and answering questions!

PR #604, opened 2020-08-06
Fix bug preventing seamless switch between datasets

mackenziekerwin:
When you open one dataset, then open another, the UI continues to display the table from the previous dataset. This is because when there is already a dataset in the store, the page renders before the new dataset gets fetched. There is a method `hasDataset()` that delays rendering until there is a dataset in the store, so I've added a condition that ensures that the stored dataset is the correct one--not the one that was previously opened.


src/components/dataset/query/QueryView.jsx
```diff
@@ -125,8 +125,9 @@ export class QueryView extends React.PureComponent {
   };
 
   hasDataset() {
-    const { dataset } = this.props;
-    return dataset && dataset.schema;
+    const { dataset, match } = this.props;
+    const datasetId = match.params.uuid;
```
RoriCremer, line 129: will match always have a params val?

RoriCremer (approved):

src/components/dataset/query/QueryView.jsx
mackenziekerwin: oh, I have no idea... I would think so, since the UUID is part of the path for each dataset. It's being used just like this in componentDidMount

fboulnois (approved):
neat!

PR #607, opened 2020-08-10
added sem var tag to image

smark88:


fboulnois (approved):
lgtm, the e2e tests failed in the cleanup step due to a blip in GCP service.

PR #609, opened 2020-08-14
DR-1267 Feature flag infrastructure

myessail:
This gets the SAM groups that the logged-in user is a part of, and filters out the ones prefixed with `jade-feature` to get a list of our features that they have access to.

### DIFF
--- a/src/sagas/repository.js
+++ b/src/sagas/repository.js
@@ -431,6 +431,23 @@ export function* countResults({ payload }) {
 }
 
 /**
+ * feature flag stuff
+ */
+
+export function* getFeatures() {
+  try {
+    const url = 'https://sam.dsde-dev.broadinstitute.org/api/groups/v1';
+    const response = yield call(authGet, url);
+    yield put({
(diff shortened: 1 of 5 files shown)


src/sagas/repository.js
```diff
@@ -430,6 +430,23 @@ export function* countResults({ payload }) {
   }
 }
 
+/**
+ * feature flag stuff
+ */
+
+export function* getFeatures() {
+  try {
+    const url = 'https://sam.dsde-dev.broadinstitute.org/api/groups/v1';
```
RoriCremer, line 439: worth double checking, but using dev sam makes sense to me for now

RoriCremer (approved):

mackenziekerwin (approved):
yeet

fboulnois (approved):
lgtm

src/sagas/repository.js
myessail: yep, checked with @fboulnois and it seems like we're just using dev for now

PR #614, opened 2020-08-18
DR-1252 Landing page for new search ui

mackenziekerwin:
Screen Shot 2020-08-18 at 2 40 55 PM

- Added a tab to the UI that is rendered only if the logged in user has access to the new search page
- Created a page for the data explorer with a general skeleton for components to be added as they are created

### DIFF
--- a/src/components/search/DataExplorerView.jsx
+++ b/src/components/search/DataExplorerView.jsx
@@ -2,22 +2,18 @@ import React from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 import { withStyles } from '@material-ui/core/styles';
+import { Divider } from '@material-ui/core';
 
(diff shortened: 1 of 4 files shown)


src/components/search/DataExplorerView.jsx
```diff
@@ -0,0 +1,45 @@
...
+        <Divider />
+        <div className={classes.section}>This is where the applied filters will go</div>
+        <Divider />
+        <div className={classes.section}>This is where the dataset results will go</div>
+      </div>
+    );
+  }
+}
+
+function mapStateToProps(state) {
```
myessail, line 42: you can just put the eslint ignore comment here for now, we'll use it later

myessail (approved):
looks good 👍

fboulnois (approved):
lgtm!

PR #644, opened 2020-08-28
DR-1283 First half of 'Snapshot Full Dataset' flow

mackenziekerwin:
full-snapshot
Opens a dialog to add details for a new snapshot when the "Snapshot Full Dataset" button is clicked. The "Save Snapshot" button, which will take the dialog to the next step in the flow, is not hooked up yet. The cancel button closes the dialog and discards the snapshot.


myessail (approved):

snf2ye (approved):

fboulnois (commented):
belated lgtm 👍

PR #666, opened 2020-09-08
Uneject the UI

myessail:
Michael, why???

From the create-react-app documentation:

> Ejecting lets you customize anything, but from that point on you have to maintain the configuration and scripts yourself. This can be daunting if you have many similar projects. In such cases instead of ejecting we recommend to fork react-scripts and any other packages you need. This article dives into how to do it in depth.

We currently don't have any custom scripts, so there's no point in having done this. And, I don't want to maintain these basically-vanilla scripts on my own if the mighty power of Facebook will do it for me.

Technically, there's no operation to "undo" ejecting, so what we did was make a new project with create-react-app and copy over the things that we needed, and then cleaned up the package.json to get rid of those extra dependencies. Building and running the UI now takes under 30 seconds and our dependencies aren't an absolute mess anymore.

Co-Authored-By: Isaac Zarsky <22642695+zarsky-broad@users.noreply.github.com>


src/components/dataset/query/sidebar/filter/CategoryWrapper.jsx
```diff
@@ -75,6 +75,7 @@ export class CategoryWrapper extends React.PureComponent {
   transformResponse = (response) => {
     const counts = {};
     if (response) {
+      // eslint-disable-next-line
```
snf2ye, line 78: What is the linter mad about here?

snf2ye (commented):
Do you need to update any of the readme for setting up the project? 

One other question below, but otherwise looks good to me!

public/index.html
```diff
@@ -0,0 +1,276 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1.0, user-scalable=0">
+    <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
+    <meta property="og:title" content="Jade Data Repository" />
```
RoriCremer, line 7: Nit: Should this be Terra not Jade?

public/index.html
```diff
@@ -0,0 +1,276 @@
...
+    <meta property="og:description" content="A place to store and share health data" />
+    <meta property="og:image" content="https://cdn-images-1.medium.com/max/1600/1*SABvfLYK2RVnxKyerjpOdQ.png" />
+    <meta property="og:type" content="website" />
+    <meta name="theme-color" content="#000000">
+    <meta name="apple-mobile-web-app-title" content="Jade Data Repo">
+    <meta name="apple-mobile-web-app-capable" content="yes">
+    <meta name="apple-mobile-web-app-status-bar-style" content="#333333">
+    <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico">
+    <link rel="icon" type="image/png" href="%PUBLIC_URL%/favicon-16x16.png" sizes="16x16">
+    <link rel="icon" type="image/png" href="%PUBLIC_URL%/favicon-32x32.png" sizes="32x32">
```
RoriCremer, line 18: Nit: Do we need all three?

RoriCremer (approved):

src/components/dataset/details/DatasetInfoCard.jsx
```diff
@@ -68,7 +68,13 @@ export class DatasetInfoCard extends React.PureComponent {
         </div>
         <div className={classes.flex}>
           <div className={classes.jadeLink}>
-            <a href={bigQueryUrl} target="_blank">
+            <a
+              href={bigQueryUrl}
+              // rel=noopener mitigates security risk here
```
fboulnois, line 73: 👍

fboulnois (approved):
overall lgtm, -20k lines, nice work!

src/components/dataset/query/sidebar/filter/CategoryWrapper.jsx
myessail: it wants a return statement in the arrow function, which i don't really care about here

PR #700, opened 2020-09-16
DR-1282 Dashboard layout

myessail:
image
Just some basic layout stuff :)


src/components/dataset/details/DatasetDetailView.jsx
```diff
@@ -1,81 +1,105 @@
...
+  infoColumnPanel: {
+    flexGrow: 1,
+  },
+  mainColumn: {
+    display: 'flex',
+    flexDirection: 'column',
+    paddingLeft: '0.5rem',
+  },
+  mainColumnRow1: {},
+  mainColumnRow2: {},
```
RoriCremer: is there a reason to keep the empty ones?

src/components/dataset/details/DatasetDetailView.jsx
```diff
@@ -1,81 +1,105 @@
...
-            datasetPolicies={datasetPolicies}
-            openSnapshotCreation={this.openSnapshotCreation}
-          />
-          <CreateFullSnapshotView
-            open={creatingSnapshot}
-            openSnapshotCreation={this.openSnapshotCreation}
-          />
+  return datasetPolicies && dataset && dataset.id === uuid ? (
+    <Fragment>
+      <h2 style={{ margin: '0.5rem' }}>{dataset.name}</h2>
```
RoriCremer: why not use a class? if because of the h2, do we use h2 a lot? and if yes, than is it worth pulling it (h2 styling and the inline) out into a full theme?

RoriCremer (approved):

src/components/dataset/details/DatasetDetailView.jsx
```diff
@@ -1,81 +1,105 @@
-import React from 'react';
+import _ from 'lodash';
```
fboulnois: Are these linting errors real or spurious?

src/components/dataset/details/VisualizeRelationshipsPanel.jsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Paper, Typography, Button } from '@material-ui/core';
```
fboulnois: Same question here?

src/routes/Private.jsx
```diff
@@ -18,6 +18,9 @@ import DataExplorerView from '../components/search/DataExplorerView';
 const styles = (theme) => ({
   wrapper: {
     fontFamily: theme.typography.fontFamily,
+    height: 'calc(100% - 64px)',
```
fboulnois, line 21: This looks pretty unusual!

src/components/dataset/details/DatasetDetailView.jsx
nmalfroy: Actually, given that we're using matarial UI, any reason we wouldn't use the `Typography` component? https://material-ui.com/components/typography/#typography
 That has the advantage of plugging nicely into the theme

src/routes/Private.jsx
nmalfroy: What is our browser support matrix?  This will work on modern browsers (and is admittedly, super handy) but won't work on ...sigh...older IE.

Also, it strikes me as strange to be doing layout stuff in a class that seems dedicated to routing

src/components/dataset/details/DatasetDetailView.jsx
```diff
@@ -1,81 +1,105 @@
...
-            openSnapshotCreation={this.openSnapshotCreation}
-          />
-          <CreateFullSnapshotView
-            open={creatingSnapshot}
-            openSnapshotCreation={this.openSnapshotCreation}
-          />
+  return datasetPolicies && dataset && dataset.id === uuid ? (
+    <Fragment>
+      <h2 style={{ margin: '0.5rem' }}>{dataset.name}</h2>
+      <div className={classes.root}>
```
nmalfroy, line 69: Why not use `Grid` components here?  I've found those make the pseudo-HTML a lot more readable since you can clearly see what the layout elements are.

src/components/dataset/details/DatasetDetailView.jsx
```diff
@@ -1,81 +1,105 @@
...
+          </div>
+        </div>
+        <div className={classes.mainColumn}>
+          <div className={classes.mainColumnRow1}>
+            <div className={classes.datasetInfoContainer}>
+              <div className={classes.headerText}>Dataset Information</div>
+            </div>
+          </div>
+          <div className={classes.mainColumnRow2}>
+            <DatasetInfoCard dataset={dataset} datasetPolicies={datasetPolicies} />
```
nmalfroy: You probably don't want to be passing these in.  You should read these from global state in the DatasetInfoCard component

src/components/dataset/details/VisualizeRelationshipsPanel.jsx
```diff
@@ -0,0 +1,47 @@
...
+    height: '100%',
+  },
+  headerText: {
+    fontWeight: theme.typography.bold,
+    textTransform: 'uppercase',
+  },
+  flex: {
+    display: 'flex',
+    justifyContent: 'space-between',
+    margin: `${theme.spacing(2)}px 0px`,
```
nmalfroy: this can (probably should?) be written as `margin: [theme.spacing(2), 0]`

src/components/dataset/details/VisualizeRelationshipsPanel.jsx
```diff
@@ -0,0 +1,47 @@
...
+  },
+  jadeLink: {
```
nmalfroy: I wonder, if we need to do this everywhere, should this be put into a `JadeLink` component?

src/components/dataset/details/DatasetDetailView.jsx
nmalfroy: those look legit.  There may be a way to get intellij to run "Organize Imports" on save, which would correct that for you

src/routes/Private.jsx
myessail: pushing this to a new ticket to fix the underlying issue!

src/components/dataset/details/DatasetInfoCard.jsx
```diff
@@ -91,4 +92,9 @@ export class DatasetInfoCard extends React.PureComponent {
   }
 }
 
-export default withStyles(styles)(DatasetInfoCard);
+const mapStateToProps = ({ datasets: { dataset, datasetPolicies } }) => ({
+  dataset,
```
nmalfroy, line 96: :+1:

nmalfroy (commented):
looks good!

PR #702, opened 2020-09-21
DR-1335 Move layout stuff out of routing layer

myessail:
Keeping some layout stuff in `Private.jsx` because those tabs are specific to logged-in folks, but this takes care of the weird magic numbers and makes the divs all have the correct (intuitive) heights.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -23,254 +23,256 @@
   </head>
   <body>
     <noscript><div style="align-items: center; display: flex; font-size: 30px; justify-content: center; min-height: 100vh;">This website need Javascript. Please turn it on.</div></noscript>
-    <div id="react" style="min-height: 100vh;">
-      <svg xmlns="http://www.w3.org/2000/svg" version="1.1">
-        <defs>
-          <filter id="gooey">
-            <feGaussianBlur in="SourceGraphic" stdDeviation="10" result="blur"></feGaussianBlur>
-            <feColorMatrix in="blur" mode="matrix" values="1 0 0 0 0  0 1 0 0 0  0 0 1 0 0  0 0 0 18 -7" result="goo"></feColorMatrix>
-            <feBlend in="SourceGraphic" in2="goo"></feBlend>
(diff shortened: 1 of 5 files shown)


src/containers/App.jsx
```diff
@@ -22,11 +22,11 @@ import Private from 'routes/Private';
 import NotFound from 'routes/NotFound';
 import Logo from 'components/Logo';
 
-import { logOut } from 'actions/index';
 import RoutePublic from 'components/RoutePublic';
 import RoutePrivate from 'components/RoutePrivate';
 import { ReactComponent as CarrotSVG } from 'media/icons/angle-line.svg';
 import { ReactComponent as SignOutSVG } from 'media/icons/logout-line.svg';
+import { logOut } from 'actions';
```
RoriCremer, line 30: how was this working before?

RoriCremer (approved):

public/index.html
```diff
@@ -23,254 +23,256 @@
...
+          body {
+            background: linear-gradient(to bottom, #00b4d5, #00191e);
           }
-        }
 
-        @-webkit-keyframes animate-to-3 {
-          25%, 75% {
-            -webkit-transform: translateX(1.5rem) scale(0.75);
-            transform: translateX(1.5rem) scale(0.75);
+          .blob {
```
nmalfroy, line 55: It feels strange that we could keep CSS in the index.  Feels like this is stuff that should live in a root component

nmalfroy (approved):
Looks good.  Eventually, I'd love to move the css out of index.html and into a root component

src/containers/App.jsx
```diff
@@ -125,7 +126,9 @@ const useStyles = makeStyles((theme) => ({
   },
   content: {
     flexGrow: 1,
-    height: '100vh',
+    paddingTop: 64,
```
fboulnois, line 129: 👍

fboulnois (approved):
lgtm

PR #711, opened 2020-09-29
DR-1284 "snapshot full dataset" modal sharing view

myessail:
snapfull

### DIFF
--- a/src/components/dataset/details/CreateFullSnapshotNamingView.jsx
+++ b/src/components/dataset/details/CreateFullSnapshotNamingView.jsx
@@ -1,4 +1,4 @@
-import React, { Fragment } from 'react';
+import React, { Fragment, useState } from 'react';
 import PropTypes from 'prop-types';
 import {
   Button,
@@ -9,6 +9,7 @@ import {
   Typography,
 } from '@material-ui/core';
 import { withStyles } from '@material-ui/core/styles';
+import { validateSnapshotName } from '../../../libs/utils';
 
 const styles = (theme) => ({
   root: {
@@ -31,6 +32,13 @@ const CreateFullSnapshotNamingView = ({
   setIsSharing,
   setName,
 }) => {
+  const [hasError, setHasError] = useState(false);
+
+  const validateNameAndProgress = (snapshotName) => {
+    // eslint-disable-next-line no-unused-expressions
+    validateSnapshotName(snapshotName) ? setIsSharing(true) : setHasError(true);
+  };
+
   return (
     <Fragment>
       <DialogTitle>Name and Describe Snapshot</DialogTitle>
@@ -45,6 +53,11 @@ const CreateFullSnapshotNamingView = ({
           onChange={(event) => setName(event.target.value)}
           value={name}
         />
+        {hasError && (
+          <Typography variant="subtitle2" color="error">
+            The name you entered is invalid. Please make sure to enter an alphanumeric name that is less than 64 characters.
+          </Typography>
+        )}
         <Typography variant="subtitle2">Description</Typography>
         <TextField
           id="snapshotDescription"
@@ -60,7 +73,7 @@ const CreateFullSnapshotNamingView = ({
       </DialogContent>
       <DialogActions className={classes.buttonContainer}>
         <Button onClick={onDismiss}>Cancel</Button>
-        <Button variant="contained" disabled={name === ''} onClick={setIsSharing}>
+        <Button variant="contained" disabled={name === ''} onClick={validateNameAndProgress}>
           Save snapshot
         </Button>
       </DialogActions>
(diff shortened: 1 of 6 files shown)


src/components/dataset/details/CreateFullSnapshotNamingView.jsx
```diff
@@ -0,0 +1,81 @@
...
+          fullWidth
+          className={classes.textField}
+          onChange={(event) => setDescription(event.target.value)}
+          value={description}
+        />
+      </DialogContent>
+      <DialogActions className={classes.buttonContainer}>
+        <Button onClick={onDismiss}>Cancel</Button>
+        <Button variant="contained" disabled={name === ''} onClick={setIsSharing}>
+          Save snapshot
```
RoriCremer, line 78: Do we also want this to be disabled if the snapshot name is already in use?
(I guess that is more validation than disabling)

RoriCremer (approved):

src/components/dataset/details/CreateFullSnapshotNamingView.jsx
```diff
@@ -0,0 +1,81 @@
...
+          rows={5}
+          fullWidth
+          className={classes.textField}
+          onChange={(event) => setDescription(event.target.value)}
+          value={description}
+        />
+      </DialogContent>
+      <DialogActions className={classes.buttonContainer}>
+        <Button onClick={onDismiss}>Cancel</Button>
+        <Button variant="contained" disabled={name === ''} onClick={setIsSharing}>
```
nmalfroy: Is that really the only requirement?  What if it's `'    '` or do we allow slashes etc?  Is there a max length?  Looks like the server has a bunch of validation, e.g.:
```
pattern: '^[a-zA-Z0-9][_a-zA-Z0-9]*$'
minLength: 1
maxLength: 63
```
I'd:
a) rewrite this check with this in mind
b) move it to a function
c) Have it return what is wrong with the name in a human readable form and show that at least as a tooltip if not error text (is there room for that in the designs?

src/components/dataset/details/CreateFullSnapshotView.jsx
```diff
@@ -1,88 +1,35 @@
...
-  constructor(props) {
-    super(props);
-    this.state = {
-      name: '',
-      description: '',
-    };
-  }
+  return (
+    <Dialog open={open} onClose={onDismiss} fullWidth>
+      {isSharing ? (
```
nmalfroy, line 14: I'd wrap this whole thing with a check for `open` since, IIRC, even when it's closed, it tries to render the dialog contents. Might be worth double checking (throw a `console.log` in `CreateFullSnapshotNamingView.render` to make sure it's not hitting that when it's not showing)

src/components/dataset/details/CreateFullSnapshotNamingView.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+const CreateFullSnapshotNamingView = ({
+  classes,
+  description,
+  onDismiss,
+  name,
+  setDescription,
+  setIsSharing,
+  setName,
+}) => {
```
nmalfroy, line 34: Linter's not happy here. Maybe take out the return so we don't get spammed with this in other PRs

nmalfroy (changes requested):
Looks good.  I think the validator needs to be fixed and maybe check you're not rendering the create modal unnecessarily

src/components/dataset/details/CreateFullSnapshotView.jsx
myessail: verified that it's not trying to render, so i'm going to leave as-is

src/libs/utils.js
```diff
@@ -8,3 +8,7 @@ import { useEffect } from 'react';
 export const useOnMount = (fn) => {
   useEffect(fn, []);
 };
+
+export const validateSnapshotName = (name) => {
+  return name.length > 1 && name.length <= 63 && name.match('^[a-zA-Z0-9][_a-zA-Z0-9]*$');
```
nmalfroy: this should be `name.length >= 1`

src/components/dataset/details/CreateFullSnapshotNamingView.jsx
```diff
@@ -60,7 +73,7 @@ const CreateFullSnapshotNamingView = ({
       </DialogContent>
       <DialogActions className={classes.buttonContainer}>
         <Button onClick={onDismiss}>Cancel</Button>
-        <Button variant="contained" disabled={name === ''} onClick={setIsSharing}>
+        <Button variant="contained" disabled={name === ''} onClick={validateNameAndProgress}>
```
nmalfroy, line 77: disabled should probably be using the new function

nmalfroy (changes requested):
Couple small things

nmalfroy (approved):

PR #717, opened 2020-10-07
fix scrolling

myessail:


fboulnois (approved):
changes lgtm

src/routes/Private.jsx
```diff
@@ -98,17 +99,19 @@ class Private extends React.Component {
...
-                    {features.searchui && (
-                      <Route exact path="/explorer" component={DataExplorerView} />
-                    )}
-                    <Route exact path="/datasets" component={DatasetsView} />
-                    <Route exact path="/datasets/:uuid/query" component={QueryView} />
-                    <Route exact path="/datasets/:uuid/details" component={DatasetDetailView} />
-                    <Route exact path="/snapshots" component={SnapshotView} />
-                    <Route exact path="/snapshots/details/:uuid" component={SnapshotDetailView} />
-                  </Switch>
+                  <div className={classes.component}>
```
nmalfroy, line 102: More CSS in the routing layer? We sure we need this?

nmalfroy (commented):
Want to just confirm my question

src/routes/Private.jsx
myessail: yeah, this component class was previously being applied to the `Switch` which doesn't actually work, but we want the overflow property to be set here instead of doing it on every new component

nmalfroy: Ok, probably at some point we should have a wrapping component wraps all pages for these kinds of things

nmalfroy (approved):

PR #739, opened 2020-10-23
DR-1445: Update skaffold template to work with modular helm chart

snf2ye:
Included Changes:
- Updating chart versions
- Moving to modular chart model

Test strategy: I was able to successfully deploy using `skaffold run` using the skaffold.yaml generated by this template.


nmalfroy (approved):
LGTM.  Longer term, do we have a sense for how to keep these in sync?  It's a bummer to have all of these definitions floating all over the place

fboulnois (approved):
lgtm, I agree with @nmalfroy 's assessment.

snf2ye:
> LGTM. Longer term, do we have a sense for how to keep these in sync? It's a bummer to have all of these definitions floating all over the place

I don't know, but I think it would be good to come up with a plan!

PR #751, opened 2020-11-09
DR-1360 Show arrays in the data browsing table

nmalfroy:
The behavior is mostly unchanged but now we:
- Don't distinct rows in the tables
- Show the array values.  Note: I built the display out of elements so that we could style more easily moving forward
- You can't sort of array values
- You currently can't filter or create snapshots using array filters.  Going to do that in a separate ticket.

Note: all values are now passed around as arrays but whether or not to display as an array is determined by the datatype returned by BigQuery.

Can you see which one is the array field :)
image

### DIFF
--- a/src/components/table/JadeTable.jsx
+++ b/src/components/table/JadeTable.jsx
@@ -161,17 +161,22 @@ export class JadeTable extends React.PureComponent {
               />
               {!polling && (
                 <TableBody data-cy="tableBody">
-                  {rows.map((row) => {
+                  {rows.map((row, i) => {
                     const drId = row.datarepo_id;
 
                     return (
                       <TableRow hover tabIndex={-1} key={drId}>
                         {columns.map((column) => {
                           const value = this.handleArrayValues(row[column.id], column);
-
                           return (
                             !_.isUndefined(value) && (
-                              <TableCell key={`${column.id}-${drId}`} className={classes.cell}>{value}</TableCell>
+                              <TableCell
+                                key={`${column.id}-${drId}`}
+                                className={classes.cell}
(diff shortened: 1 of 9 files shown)


src/components/table/JadeTable.jsx
```diff
@@ -115,9 +120,26 @@ export class JadeTable extends React.PureComponent {
...
+      returnValue.push(
+        ...value.flatMap((v, i) => [
+          v,
+          i < value.length - 1 ? (
+            <span key={`sep-${i}`}>
+              ,<br />
+            </span>
+          ) : undefined,
+        ]),
+      );
```
myessail, line 138: not sure if i completely understand what this is doing, is this just adding commas until the end of the array?

myessail (approved):

src/components/table/JadeTable.jsx
nmalfroy: Yeah...adding commas to everything except the last element

src/components/table/JadeTable.jsx
```diff
@@ -32,6 +34,9 @@ const styles = (theme) => ({
   spinner: {
     margin: 'auto',
   },
+  cell: {
+    ...ellipsis,
+  },
```
RoriCremer, line 39: is the full set of values in a popover?

RoriCremer (approved):

src/components/table/JadeTable.jsx
nmalfroy: I don't know if it actually even causes ellipsis but is seems to give the table a good hint to not have such a wonky layout on large values

PR #753, opened 2020-11-09
DR-1442 Snapshot area

myessail:
dashboardsnapshot

Area under "snapshots" as well as some general page cleanup/spacing/etc to make it look pretty


src/components/dataset/details/DatasetDetailView.jsx
```diff
@@ -62,31 +107,55 @@ const DatasetDetailView = ({
     dispatch(getDatasetPolicy(uuid));
   });
 
+  let snapshotCards = [];
+  snapshotCards.push(<NewSnapshotButton datasetId={dataset.id} />);
```
nmalfroy: I think you're gonna need a `key`

src/components/dataset/details/DatasetDetailView.jsx
```diff
@@ -62,31 +107,55 @@ const DatasetDetailView = ({
...
+
+  snapshotCards = snapshotCards.concat(
+    fakeSnapshots.map((snapshot) => {
+      return <SnapshotInfoCard snapshot={snapshot} />;
```
nmalfroy: you're gonna need a `key` here too

src/components/dataset/details/DatasetDetailView.jsx
```diff
@@ -62,31 +107,55 @@ const DatasetDetailView = ({
...
+    }),
+  );
+
+  console.log(
```
nmalfroy: I don't think you want this log statement to stay, right?

src/modules/theme.js
```diff
@@ -29,6 +29,55 @@ export default createMuiTheme({
       '"Segoe UI Symbol"',
     ].join(','),
     bold: '600',
+    h1: {
```
nmalfroy, line 32: 👍

nmalfroy (changes requested):
Looks like you've got some linting stuff still failing and I think you'll need to add the keys (I think the linter is complaining about that too). But it's looking good!

src/components/dataset/details/DatasetDetailView.jsx
```diff
@@ -32,20 +39,58 @@ const styles = (theme) => ({
   mainColumn: {
     display: 'flex',
     flexDirection: 'column',
-    paddingLeft: '0.5rem',
+    marginLeft: 40,
```
RoriCremer, line 42: dont we have themes for these numbers so we stay consistent?

RoriCremer (approved):

src/components/dataset/details/DatasetDetailView.jsx
```diff
@@ -32,20 +39,58 @@ const styles = (theme) => ({
...
+    marginBottom: '1rem',
+  },
+  snapshotCardsContainer: {
+    display: 'grid',
+    gridTemplateColumns: 'repeat(auto-fill, minmax(360px, 32%))',
+    gridGap: '1rem',
   },
 });
 
+const fakeSnapshots = [
```
RoriCremer, line 61: my gut reaction is that I want this in a file exclusively for testing and not here---although it looks like maybe you are planning to remove this entirely

src/components/dataset/details/DatasetDetailView.jsx
myessail: for some of these numbers yes! i've been using rem instead of spacing lately since that's a more general standard, but there are some places that we don't want to scale if we resize things so we keep using pixels

src/components/dataset/details/DatasetDetailView.jsx
myessail: yep, going to be removed once we get an endpoint for this data

nmalfroy (approved):
lgtm!

nmalfroy:
Ah the test has a sad

nmalfroy:
Is that because you changed some things to use `Typography`?

PR #813, opened 2021-01-13
Add multi-stage build for Dockerfile

fboulnois:
Includes a build stage in the Dockerfile so a developer does not first need to manually build the jade-data-repo-ui. This will be useful for the appsec team when they implement the Trivy action.


smark88 (approved):

fboulnois:
I think the test failures are unrelated due to my changes due to https://github.com/DataBiosphere/jade-data-repo-ui/pull/815

nmalfroy (approved):
I get keeping the functionality the same but it's a little weird that we build and run as park of docker run...Seems like we should build as a build step then save copy that result to build the docker image

nmalfroy:
Caught up off line.  lgtm!

PR #817, opened 2021-01-14
Partially revert "Add multi-stage build for Dockerfile (#813)"

fboulnois:
Only remove the `npm` steps

This partially reverts commit a35b76819f4f05d77776b5702afd7e960c50dbe8 .

### DIFF
--- a/.github/workflows/dev-image-update.yaml
+++ b/.github/workflows/dev-image-update.yaml
@@ -57,6 +57,11 @@ jobs:
         uses: "broadinstitute/github-action-get-previous-tag@master"
         env:
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Build nmp build for new ui container
+        run: |
+          git pull
+          GCR_TAG=$(git rev-parse --short HEAD)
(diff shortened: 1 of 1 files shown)


smark88 (approved):

nmalfroy (approved):

.github/workflows/dev-image-update.yaml
```diff
@@ -57,6 +57,11 @@ jobs:
         uses: "broadinstitute/github-action-get-previous-tag@master"
         env:
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Build nmp build for new ui container
+        run: |
+          git pull
+          GCR_TAG=$(git rev-parse --short HEAD)
```
pshapiro4broad, line 63: Can you move this into `Pull down new tags`? I don't think `git pull` or `$GITHUB_ENV` are still needed, so the only thing left from this task is the `git rev-parse`, which you could do in that task instead.

pshapiro4broad: Hmm, apparently `GITHUB_ENV` is a github actions thing, so maybe it's still needed. Pretty sure the `git pull` can go though. Either way this is no longer a `build` task.

PR #834, opened 2021-02-02
DR-1606: first pass at making a yq job to bump the default ui helm charts tag value

smark88:


.github/workflows/helmtagbump.yaml
```diff
@@ -71,3 +71,68 @@ jobs:
...
+      - name: "bump ui chart appVersion"
+        uses: docker://mikefarah/yq:3
+        with:
+          args: yq w -i datarepo-helm/charts/datarepo-ui/Chart.yaml appVersion ${{ steps.uiprevioustag.outputs.tag }}"
+      - name: "get new chart verion"
+        id: new_version
+        run: |
+          version=$(docker run --rm -v ${PWD}:/workdir mikefarah/yq:3 yq r datarepo-helm/charts/datarepo-ui/Chart.yaml 'version')
+          a=(`echo $version | sed 's/\./ /g'`)
+          ((a[2]++))
```
fboulnois, line 106: Does this bump major, minor or patch?

.github/workflows/helmtagbump.yaml
```diff
@@ -71,3 +71,68 @@ jobs:
...
+          printf "increase default ui version: ${a[2]}\n\n"
+          new_version="${a[0]}.${a[1]}.${a[2]}"
+          echo "::set-output name=new_version::$new_version"
```
fboulnois, line 109: Hrm, I wonder how robust this mechanism is.

fboulnois (commented):
Two main questions:

.github/workflows/helmtagbump.yaml
smark88: minor

```((a[2]++))``` is third octet

File: .github/workflows/helmtagbump.yaml (hunk 2 above)
Comment (smark88) [reply .github/workflows/helmtagbump.yaml:109]: currently use it here https://github.com/broadinstitute/datarepo-umbrella-release-action/blob/master/src/main.sh

File: .github/workflows/helmtagbump.yaml [hunk 3]
```diff
@@ -71,3 +71,68 @@ jobs:
...
+          token: ${{ secrets.HELM_REPO_TOKEN }}
+          path: datarepo-helm
+      - name: "ui chart default tag find and replace"
+        uses: docker://mikefarah/yq:3
+        with:
+          args: yq w -i datarepo-helm/charts/datarepo-ui/values.yaml image.tag ${{ steps.uiprevioustag.outputs.tag }}"
+      - name: "bump ui chart appVersion"
+        uses: docker://mikefarah/yq:3
+        with:
+          args: yq w -i datarepo-helm/charts/datarepo-ui/Chart.yaml appVersion ${{ steps.uiprevioustag.outputs.tag }}"
```
snf2ye, line 100: I'm a little confused -  there are two versions in the Chart.yaml, right? "version" and "appVersion". Why do they match right now? they're both 0.0.15. https://github.com/broadinstitute/datarepo-helm/blob/master/charts/datarepo-ui/Chart.yaml#L17

smark88:
```
version ==  helm chart version 
and 
appVersion == supposed to be application version 
```
but since we never previously updated the default application version in the values.yaml i never saw a need to change it, till now


more info here https://helm.sh/docs/topics/charts/

snf2ye (approved):

fboulnois (approved):
lgtm

PR #839, opened 2021-02-04
DR-1586 Make sure that we properly handle the case where a user doesn't have permissions to view policies

nmalfroy:
...e.g. view who belongs to a specific policy.

Note: this is for the details side bar on the snapshot view

This also:
- Makes sure that we can add/remove stewards
- Refactors the code to make it easier and consistent to add/remove snapshot roles
- Does a drive by fix which resets the UI state when you log out (so that when you log in as someone else, it doesn't show you your previous snapshots and datasets while it loads)


src/components/DetailViewHeader.jsx
```diff
@@ -29,26 +29,28 @@ const styles = (theme) => ({
 
 export class DetailViewHeader extends React.PureComponent {
   static propTypes = {
-    addCustodian: PropTypes.func,
+    addSteward: PropTypes.func,
     addReader: PropTypes.func,
```
myessail, line 32: if you wanna fix these linting warnings since you're touching this file, i would appreciate 🙂

myessail (approved):
💪

src/components/DetailViewHeader.jsx
nmalfroy: ah touché.  Will do

fboulnois (approved):
Seems pretty clear, lgtm

PR #846, opened 2021-02-19
DR-1623 Suppress errors when trying to talk to Sam

nmalfroy:
Longer term, we should not always point to dev Sam but to the Sam that matches with the environment we are running in

### DIFF
--- a/src/sagas/repository.js
+++ b/src/sagas/repository.js
@@ -441,6 +441,7 @@ export function* countResults({ payload }) {
 
 export function* getFeatures() {
   try {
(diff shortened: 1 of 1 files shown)


pshapiro4broad (approved):
This mitigates the error but doesn't fix the underlying problem, which is what DR-1623 describes. Will there be another ticket to fix the underlying problem, or do you plan to leave DR-1623 open?

nmalfroy:
I'm planning on leaving it open

samanehsan (approved):

PR #856, opened 2021-02-24
DR-1510: switch to new action and broadbot token

snf2ye:
Successful run: https://github.com/DataBiosphere/jade-data-repo-ui/runs/1973017072?check_suite_focus=true
Merge into datarepo-helm-definitions: https://github.com/broadinstitute/datarepo-helm-definitions/commit/9baee6ff225a0203c7a29aad319740cf0ae2350a
Merge into datarepo-helm: https://github.com/broadinstitute/datarepo-helm/commit/26482131dbdd4093de78e970c1bbcd095aa5ea2a

Adding small additional change that has already been merged into jade-data-repo to prevent dev-image-update from running when broadbot initiates the change. It will prevent the following failure: 
Screen Shot 2021-02-25 at 12 34 43 PM


fboulnois (approved):
lgtm

.github/workflows/helmtagbump.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: 'broadinstitute/datarepo-helm-definitions'
-          token: ${{ secrets.HELM_REPO_TOKEN }}
+          token: ${{ secrets.BROADBOT_TOKEN}}
           path: datarepo-helm-definitions
       - name: "integration-1 find and replace"
```
pshapiro4broad, line 25: Not sure if there's a way to do this, but I think it would be great if we could have an action that did the checkout, patch and merge in one step, to avoid duplicate everywhere we do this. It looks like everything is boilerplate except for the patch step. If the patch was done as a bash script, that could be one way to do it.

snf2ye:
Test run:
-> Original action: https://github.com/DataBiosphere/jade-data-repo-ui/actions/runs/600208822
-> datarepo-helm actions down the pipeline passed:
image

.github/workflows/helmtagbump.yaml
snf2ye: Yes, this is definitely something I'd like to work on.  Additionally, our "helmtagbump.yaml" here in the UI almost identical to our "helmtagbumper.yaml" in the API - there must be a way for us to reduce all of this duplicated code.

The easiest path forward that I see is to create a new action in datarepo-actions (like we just did for the "merger" action) that we could group some of these steps together and we could greatly reduce the number of steps and the duplicated code. 

I think there is also an option for a "[workflow template](https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization)", but I haven't figured out a good use case for this.  I also think we'd have to share it with all of the DataBiosphere org.

smark88 (approved):

PR #858, opened 2021-02-26
DR-1510: Point to new version of datarepo-actions

snf2ye:


fboulnois (approved):
lgtm!

samanehsan (approved):
LGTM! Can you remind me why we need to bump the version here as well? I thought there was only an issue with the perf tests but I am probably missing something 🤔

smark88 (approved):

snf2ye:
> LGTM! Can you remind me why we need to bump the version here as well? I thought there was only an issue with the perf tests but I am probably missing something 🤔

This is just bumping the version for datarepo-actions after my changes a couple of days ago. I didn't include these version bumps in my last PR just b/c I wanted to merge in small chunks at a time. But, I see where it's confusing - I combined this version bump and the perf test fix in the same PR in jade-data-repo.

PR #909, opened 2021-04-05
DR-1731: add scheduled job

snf2ye:


pshapiro4broad (approved):
If this fails, will it block promotion of the build? Or will that be handled elsewhere?

That is, in order for this result to be respected, does it require someone seeing the failure and taking action, or is there an automation set up so user intervention is not required?

fboulnois (commented):
This can be useful to check that there are no regressions in the UI code. However, it would not be able to catch the type of issue that we saw today as that lives in the helm layer.

fboulnois (approved):
caught up offline, lgtm 👍

PR #924, opened 2021-04-15
DR-1739: "Wait for Deployment" github action

snf2ye:


.github/workflows/test-e2e.yml
```diff
@@ -89,10 +90,10 @@ jobs:
...
           helm_oidc_proxy_chart_version: 0.0.22
           helm_imagetag_update: 'ui'
-      - name: wait for status checks
-        uses: broadinstitute/datarepo-actions/actions/main@0.57.0
-        with:
-          actions_subcommand: 'waitfordeployment'
+      - name: "Wait for deployment to come back online"
+        uses: broadinstitute/datarepo-actions/actions/wait-for-deployment@0.59.0
+        env:
+          DEPLOYMENT_TYPE: 'ui'
```
fboulnois, line 96: I am guessing the other envs don't need to be set since it's the UI?

snf2ye: I think I do set this for the other actions - I didn't define a default.

snf2ye: But, we don't need to define the "desired githash" for this since we're not waiting for a specific version to deployed. Although, one day we might want to add that.

.github/workflows/dev-image-update.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
           token: ${{ secrets.BROADBOT_TOKEN }}
           path: datarepo-helm-definitions
       - name: "Bump the tag to a new version"
-        uses: broadinstitute/datarepo-actions/actions/main@0.57.0
+        uses: broadinstitute/datarepo-actions/actions/main@0.59.0
```
pshapiro4broad, line 46: Is it worth updating version if we know it has no effect? Better to keep them all the same, I guess?

pshapiro4broad (approved):

samanehsan (approved):
LGTM 🌦️

fboulnois (approved):
lgtm

.github/workflows/dev-image-update.yaml
snf2ye: That's been my approach - I'd rather keep everything on the same version. We've run into issues before when some actions get behind in version and then we don't realize we're introducing errors until we actually need to make a change. It gets messy.

PR #926, opened 2021-04-21
DR-1700 Notify users if TDR is down or unavailable

tlangs:
A TDR error page! Right now, TDR just checks that the server is up and functional when the page is first loaded. It doesn't continually check yet. That should probably be another ticket. 

# Page showing dependencies are down
image

# Page showing TDR is down
image

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -85,14 +85,13 @@ function bootstrap() {
         resolve();
       })
       .catch((error) => {
-        // The React proxy returns a 504 if the '/status' endpoint can't be reached.
-        if (error.response.status === 504) {
+        // The API returns a 503 UNAVAILABLE from the  '/status' endpoint if dependencies are down.
+        if (error.response === undefined || error.response.status !== 503) {
           store.dispatch({
             type: ActionTypes.GET_SERVER_STATUS_DOWN,
             status: {
               tdrOperational: false,
               apiIsUp: false,
-              serverStatus: error.response.data,
             },
           });
         } else {
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -138,6 +138,7 @@ function render(Component) {
   }
 }
 
+console.log('foobar');
 bootstrap()
   .then(() => render(App))
   .catch((err) => {
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -85,7 +85,7 @@ function bootstrap() {
         resolve();
       })
       .catch((error) => {
-        console.error('The server is down!');
+        // The React proxy returns a 504 if the '/status' endpoint can't be reached.
         if (error.response.status === 504) {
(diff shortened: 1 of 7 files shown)


src/index.jsx
```diff
@@ -31,35 +42,71 @@ function getConfig() {
...
+              })
+              .catch(reject);
           })
           .catch(reject);
+        resolve();
       })
-      .catch(reject);
+      .catch((error) => {
+        console.error('The server is down!');
+        if (error.response.status === 504) {
```
samanehsan: What if we get other 500 errors from the API?

src/index.jsx
```diff
@@ -31,35 +42,71 @@ function getConfig() {
...
+              apiIsUp: false,
+              serverStatus: error.response.data,
+            },
+          });
+        } else {
+          store.dispatch({
+            type: ActionTypes.GET_SERVER_STATUS_FAILURE,
+            status: {
+              tdrOperational: false,
+              apiIsUp: true,
```
samanehsan, line 102: I'm wondering what situation this would happen in. I thought if there was any failure status from the API, this would be false.

tlangs: If the TDR api is up, but SAM is not, then the API is running but TDR isn't operational. We can use that information to give a more detailed message about whats going on vs just "the server is down".

src/index.jsx
tlangs: The 504 isn't from the TDR API, but from the React proxy. It always returns a 504 if it can't reach the endpoint. I've added a comment to explain.

snf2ye (approved):
It's been a while since I've looked at UI code, but LGTM! If not too much trouble, could you also provide a screen shot for the error page?

src/index.jsx
samanehsan: Oh interesting! Thanks for clarifying that

src/index.jsx
samanehsan: Hmmm so I'm looking at the [status endpoint code](https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/java/bio/terra/service/configuration/StatusService.java#L46), and the Data Repo API status is only "ok" if SAM and Postgres are up, otherwise the status endpoint returns a 503 status code. Does that mean the react proxy would return a 504 (so it falls under the first condition)?

tlangs:
@snf2ye I've added screenshots to the description!

src/index.jsx
samanehsan: To document what we discussed at stand-up, the first condition occurs when the api pod is unavailable (so there is no response at all), while this one occurs if we can reach the status endpoint, but there's a 503 response.

samanehsan (approved):
LGTM ☀️

PR #958, opened 2021-05-24
[DR-1826] Add region to snapshot table

samanehsan:
### Home tab:
Screen Shot 2021-05-24 at 12 32 19 PM

### Snapshots tab:
Screen Shot 2021-05-24 at 12 32 25 PM

### DIFF
--- a/src/components/table/SnapshotTable.jsx
+++ b/src/components/table/SnapshotTable.jsx
@@ -48,10 +48,10 @@ class SnapshotTable extends React.PureComponent {
         render: (row) => moment(row.createdDate).fromNow(),
       },
       {
-        label: 'Storage',
+        label: 'Storage Regions',
         property: 'storage',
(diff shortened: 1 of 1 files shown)


tlangs (commented):
Is there a reason `Datasets` has `Storage Regions`, but snapshots only has `Storage`?

samanehsan:
@tlangs good point! Those should be consistent with each other so I'll change storage --> storage regions

tlangs (approved):

samanehsan:
@tlangs I just fixed it!

nmalfroy (approved):
Same point as Trevyn.  Also, is there a ticket to link this to?

samanehsan:
Yes! It's https://broadworkbench.atlassian.net/secure/RapidBoard.jspa?rapidView=44&projectKey=DR&modal=detail&selectedIssue=DR-1826

PR #963, opened 2021-06-24
DR-1819 Correct the pagination to display accurate totals

tlangs:
By including the filteredTotal as well as the total count for enumerate endpoints, the UI can tell users how many total elements there are, as well as how many match their filter. It also allows us to fix a bug in the pagination where the totals didn't add up.

Screen Shot 2021-06-23 at 4 06 10 PM
Screen Shot 2021-06-23 at 4 05 57 PM

Screen Shot 2021-06-23 at 4 06 44 PM
Screen Shot 2021-06-23 at 4 06 25 PM


nmalfroy (approved):
looks good!  Just make sure that the API is deployed before we commit this

samanehsan (approved):
Looks good! Thanks for attaching screenshots

PR #995, opened 2021-09-09
DR-1843 Add Cloud Platform Column to datasets and snapshots

tlangs:
Added Cloud Platform column in tables for datasets and snapshots.
Screen Shot 2021-09-09 at 1 40 56 PM

### DIFF
--- a/src/components/table/DatasetTable.jsx
+++ b/src/components/table/DatasetTable.jsx
@@ -71,6 +71,11 @@ class DatasetTable extends React.PureComponent {
         property: 'storage.array',
         render: (row) => Array.from(new Set(row.storage.map((s) => s.region))).join(', '),
       },
+      {
+        label: 'Cloud Platform',
+        property: 'storage.array',
+        render: (row) => Array.from(new Set(row.storage.map((s) => s.cloudPlatform))).join(', '),
+      },
     ];
     return (
       <LightTable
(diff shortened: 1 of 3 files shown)


src/components/table/DatasetTable.jsx
```diff
@@ -71,6 +71,11 @@ class DatasetTable extends React.PureComponent {
         property: 'storage.array',
         render: (row) => Array.from(new Set(row.storage.map((s) => s.region))).join(', '),
       },
+      {
+        label: 'Cloud Platform',
+        property: 'storage.array',
+        render: (row) => Array.from(new Set(row.storage.map((s) => s.cloudPlatform))).join(', '),
```
pshapiro4broad, line 77: Do we support datasets or snapshots with more than one cloud platform?

pshapiro4broad (approved):
Seems redundant since we already have storage regions which is cloud specific. But I guess this makes it explicit.

src/components/table/DatasetTable.jsx
tlangs: Currently, yes. Since we've been slowly porting functionality from GCP to Azure, we could have datasets with both cloud providers. Once we finish up the azure work, there won't be any combined cloud platform datasets.

tlangs:
I think some of the region names can be very similar so having an explicit identifier for cloud platform is useful.

samanehsan (approved):
Lgtm! ⛅

PR #1003, opened 2021-09-23
DR-2109 Make Welcome Page compliant and some UI refresh

nmalfroy:
Commit by commit ought to tell the story but this makes changes to bring the UI up to match with Terra

**Front page**
image

- Fixed up layout
- Descriptive text
- Links to doc
- Footer with links to ToS, Security, Github, Copyright
- Newer background

**General updates**
image
- Update toolbar
- Remove animations
- Widen content

**Bug fix**
- Make sure that we don't load the google login button until the client id is loaded from config


snf2ye (approved):

src/components/WelcomeView.jsx
```diff
@@ -81,53 +122,101 @@ export class WelcomeView extends React.PureComponent {
...
+            </p>
+            <div id="signin-button-container" />
+            <div className={classes.newUser}>New User?</div>
+            <div className={classes.header}>Terra Data Repository requires a Google account.</div>
+            <p>
+              Terra Data Repository uses your Terra account which uses a Google account. Once you
+              have signed in and completed the user profile registration step, you can start using
+              Terra and Terra Data Repository.
+            </p>
+            <a href="https://app.terra.bio/" className={classes.terraLink}>
```
snf2ye: Should these URLs change per environment?

nmalfroy: Yeah, that's a good point.  I was debating since I'm actually not sure if these links exist in all environments.  I'll check

tlangs (approved):

PR #1010, opened 2021-10-08
[DR-2137] Check billing profile access when creating snapshots

samanehsan:
**Changes:**
1. "Create snapshot" is disabled and the "share" panel is not displayed if the logged in user does not have GET access to the billing profile:
Screen Shot 2021-10-07 at 6 09 39 PM

 
2. "Create snapshot" button is blue when activated
Screen Shot 2021-10-07 at 6 07 30 PM

### DIFF
--- a/src/actions/app.js
+++ b/src/actions/app.js
@@ -9,8 +9,8 @@ import { createActions } from 'redux-actions';
 import { ActionTypes } from 'constants/index';
 
 export const { getBillingProfileById } = createActions({
-   [ActionTypes.GET_BILLING_PROFILE_BY_ID]: (profile) => profile,
-   [ActionTypes.GET_BILLING_PROFILE_BY_ID_SUCCESS]: (profile) => profile,
+  [ActionTypes.GET_BILLING_PROFILE_BY_ID]: (profile) => profile,
+  [ActionTypes.GET_BILLING_PROFILE_BY_ID_SUCCESS]: (profile) => profile,
 });
 
 export const { createSnapshot } = createActions({
--- a/src/actions/app.js
+++ b/src/actions/app.js
@@ -8,6 +8,11 @@ import { createActions } from 'redux-actions';
 
 import { ActionTypes } from 'constants/index';
 
+export const {getBillingProfileById} = createActions({
+   [ActionTypes.GET_BILLING_PROFILE_BY_ID]: (profile) => profile,
+   [ActionTypes.GET_BILLING_PROFILE_BY_ID_SUCCESS]: (profile) => profile,
+});
+
 export const { createSnapshot } = createActions({
   [ActionTypes.CREATE_SNAPSHOT]: () => ({}),
   [ActionTypes.CREATE_SNAPSHOT_JOB]: (snapshot) => snapshot,
--- a/src/actions/app.js
+++ b/src/actions/app.js
@@ -8,7 +8,7 @@ import { createActions } from 'redux-actions';
 
 import { ActionTypes } from 'constants/index';
 
-export const {getBillingProfileById} = createActions({
+export const { getBillingProfileById } = createActions({
    [ActionTypes.GET_BILLING_PROFILE_BY_ID]: (profile) => profile,
    [ActionTypes.GET_BILLING_PROFILE_BY_ID_SUCCESS]: (profile) => profile,
 });
--- a/src/components/dataset/query/QueryView.jsx
+++ b/src/components/dataset/query/QueryView.jsx
@@ -65,6 +65,7 @@ export class QueryView extends React.PureComponent {
     orderBy: PropTypes.string,
     queryResults: PropTypes.object,
     profile: PropTypes.object,
+    table: PropTypes.object,
   };
 
   componentDidMount() {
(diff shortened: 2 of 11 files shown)


src/actions/app.js
```diff
@@ -8,6 +8,11 @@ import { createActions } from 'redux-actions';
 
 import { ActionTypes } from 'constants/index';
 
+export const { getBillingProfileById } = createActions({
+  [ActionTypes.GET_BILLING_PROFILE_BY_ID]: (profile) => profile,
```
nmalfroy, line 12: God, I just want to auto gen all of this stuff...some day!

src/components/dataset/query/QueryView.jsx
```diff
@@ -77,8 +82,21 @@ export class QueryView extends React.PureComponent {
...
+    if (
+      this.hasDataset() &&
+      dataset.defaultProfileId &&
+      canLink === false &&
+      Object.keys(prevProps.profile).length === 0
+    ) {
+      dispatch(getBillingProfileById(dataset.defaultProfileId));
+    }
+
+    if (profile.id && canLink === false) {
```
nmalfroy: Do you need the `canLink === false`?  I don't know if set state triggers an event if the value doesn't change.  Also, I think it's more idiomatic to use `!canLink`

src/components/dataset/query/QueryView.jsx
```diff
@@ -77,8 +82,21 @@ export class QueryView extends React.PureComponent {
...
+    const { dataset, dispatch, filterStatement, joinStatement, orderBy, profile } = this.props;
+    const { selected, canLink } = this.state;
+
+    if (
+      this.hasDataset() &&
+      dataset.defaultProfileId &&
+      canLink === false &&
+      Object.keys(prevProps.profile).length === 0
+    ) {
+      dispatch(getBillingProfileById(dataset.defaultProfileId));
```
nmalfroy: It feels a little weird to be triggering the dispatch from the component. Could we have something that triggers this just based on the dataset being fetched successfully?

src/components/dataset/query/sidebar/panels/FilterPanel.jsx
```diff
@@ -159,9 +167,12 @@ export class FilterPanel extends React.PureComponent {
       joinStatement,
       selected,
       handleCreateSnapshot,
+      canLink,
     } = this.props;
     const { searchString, openFilter } = this.state;
     const filteredColumns = table.columns.filter((column) => column.name.includes(searchString));
+    const billingErrorMessage =
+      'You cannot create a snapshot because you do not have access to the dataset billing profile.';
```
nmalfroy: I think `dataset's`

nmalfroy (commented):
Looks good!  I just had a quick question.  Maybe we can parking lot

src/components/dataset/query/QueryView.jsx
samanehsan: I was thinking we wouldn't need to change the state if it is already true, but it looks like it won't re-trigger the component update anyways if the value doesn't change!

nmalfroy: Right on!  I think that's what happens when you use a PureComponent (which ends up being a nice perf optimization)

src/components/dataset/query/QueryView.jsx
samanehsan: I see what you mean...is there a way to wait for the result of the dataset dispatch? Then I could move this to `componentDidMount`

nmalfroy: I think that's there's an action that triggers when the load is complete.  I think the sagas can link events.  I haven't done this in TDR yet but it could work.  Lemme take a quick look

src/components/dataset/query/QueryView.jsx
nmalfroy: I don't think we do this today but if you check out this:
https://redux-saga.js.org/docs/advanced/Channels/

I think that this is what we want to do.  Basically, watch for the `GET_DATASET_BY_ID_SUCCESS` action and have that trigger fetching the profile

samanehsan: Ooh thanks! I'll take a look at that

nmalfroy:
Note, this would go in repository.js I think

nmalfroy (approved):
thanks for the changes!

src/components/dataset/query/QueryView.jsx
```diff
@@ -206,6 +223,8 @@ function mapStateToProps(state) {
     joinStatement: state.query.joinStatement,
     queryResults: state.query.queryResults,
     orderBy: state.query.orderBy,
+    profile: state.profiles.profile,
+    canLink: state.canLink,
```
myessail: i think you can omit `canLink` here, only things that you're pulling from the global state need to be in `mapStateToProps`

src/sagas/repository.js
```diff
@@ -345,6 +345,32 @@ export function* getDatasetTablePreview({ payload }) {
...
+    });
+  } catch (err) {
+    yield put({
+      type: ActionTypes.EXCEPTION,
+      profile: null,
+    });
+  }
+}
+
+export function* watchRequests() {
```
myessail: i'm not super familiar with `take` and `actionChannel`, but it might be useful to give this a more billing profile specific name unless we plan on adding other polling actions here as well

myessail (approved):
LGTM, few minor comments

PR #1014, opened 2021-10-14
[DR-2152] Add export snapshot link

samanehsan:
Add a link to the "export" button in the pop-up that appears after a successful snapshot create. This is the same link that is on the snapshot detail page.


nmalfroy (approved):
nice!

myessail (approved):

PR #1015, opened 2021-10-14
[DR-2136] Rename confusing names

samanehsan:
**Changes:**
- "Data Snapshot" --> "Select Data"

Screen Shot 2021-10-14 at 4 20 45 PM

- "Release Dataset" --> "Create Snapshot"
Screen Shot 2021-10-14 at 4 07 55 PM

- The "Next" button is now blue when active instead of gray. I wanted to add that here since it's a relatively small change and was pointed out during a brief UI review:
Screen Shot 2021-10-14 at 4 21 03 PM


nmalfroy (approved):
lgtm!

myessail (approved):

PR #1019, opened 2021-11-03
DR-2248 Make UI error toasts more informative

myessail:
errormsg

### DIFF
--- a/src/components/Toast.jsx
+++ b/src/components/Toast.jsx
@@ -37,17 +37,19 @@ export class Toast extends React.PureComponent {
   static propTypes = {
     classes: PropTypes.object.isRequired,
     errorMsg: PropTypes.string,
+    status: PropTypes.string,
   };
 
   render() {
-    const { classes, errorMsg } = this.props;
+    const { classes, errorMsg, status } = this.props;
     return (
       <Card className={classes.card}>
         <CardContent className={classes.content}>
           <div className={classes.cardBody}>
             <ErrorIcon className={classes.icon} alt="logo" />
             <div className={classes.text}>
-              {errorMsg || 'An error occurred, please try again or submit a bug report'}
+              {`Error ${status}: ${errorMsg}` ||
+                'An error occurred, please try again or submit a bug report'}
             </div>
           </div>
         </CardContent>
--- a/src/components/Toast.jsx
+++ b/src/components/Toast.jsx
@@ -42,15 +42,20 @@ export class Toast extends React.PureComponent {
 
   render() {
     const { classes, errorMsg, status } = this.props;
+
+    let errString;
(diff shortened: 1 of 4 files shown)


src/components/Toast.jsx
```diff
@@ -37,17 +37,19 @@ export class Toast extends React.PureComponent {
...
+    const { classes, errorMsg, status } = this.props;
     return (
       <Card className={classes.card}>
         <CardContent className={classes.content}>
           <div className={classes.cardBody}>
             <ErrorIcon className={classes.icon} alt="logo" />
             <div className={classes.text}>
-              {errorMsg || 'An error occurred, please try again or submit a bug report'}
+              {`Error ${status}: ${errorMsg}` ||
+                'An error occurred, please try again or submit a bug report'}
```
nmalfroy: I'm not sure this is right.  I think the || will never resolve since it's happening outside of the string.  I think you need to include it inside of an expression within the string template

src/sagas/repository.js
```diff
@@ -435,7 +435,7 @@ export function* pageQuery({ payload }) {
       results: response,
     });
   } catch (err) {
-    showNotification(err);
+    showNotification(err.response);
```
nmalfroy: With all of these err.responses, do we need to worry about err potentially being null?

nmalfroy (changes requested):
A question and a comment but otherwise looks good

samanehsan (approved):
Lgtm! 🍞

src/sagas/repository.js
```diff
@@ -96,7 +96,11 @@ function* pollJobWorker(jobId, jobTypeSuccess, jobTypeFailure) {
       yield call(pollJobWorker, jobId, jobTypeSuccess, jobTypeFailure);
     }
   } catch (err) {
-    showNotification(err);
+    if (err.response) {
```
nmalfroy: This logic gets repeated a bunch...would it make sense to have it in the `showNotification` function?

src/modules/notifications.jsx
```diff
@@ -1,10 +1,19 @@
 import React from 'react';
 import Toast from 'components/Toast';
 import { store } from 'react-notifications-component';
+import _ from 'lodash';
 
 export function showNotification(err) {
+  let message;
+  const errDetail = _.get(err, 'data.errorDetail');
+  if (_.isUndefined(errDetail) || _.isEmpty(errDetail)) {
```
nmalfroy: huh...does `_.isEmpty` not cover `_.isUndefined`?

nmalfroy (commented):
Might make sense to fold that repeated logic into the showNotification function?

nmalfroy (approved):
beautiful, thanks!

PR #1037, opened 2021-12-03
[DR-2308] Automatically log out users who have been idle for 15 minutes or more

petesantos:
Users who are considered clinical users (as determined by group membership) will now be have a timer running in the background. After fifteen minutes of inactivity the user will be logged out and will need to re-enter their password. At the 3 minute mark the user will see a dialog window with a countdown timer.

This functionality will work across multiple open tabs, and will log the user out of all signed in tabs.

Countdown timer
image

User logged out
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "modernizr": "^3.11.8",
     "moment": "^2.29.1",
     "prop-types": "^15.7.2",
-    "qs": "^6.10.1",
     "react": "^16.14.0",
     "react-dom": "^16.14.0",
     "react-helmet": "^6.1.0",
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "modernizr": "^3.11.8",
     "moment": "^2.29.1",
     "prop-types": "^15.7.2",
+    "qs": "^6.10.1",
     "react": "^16.14.0",
     "react-dom": "^16.14.0",
     "react-helmet": "^6.1.0",
--- a/src/components/IdleStatusMonitor.jsx
+++ b/src/components/IdleStatusMonitor.jsx
@@ -68,6 +68,7 @@ const CountdownModal = withStyles(styles)(({ classes, onSignOut, countdown }) =>
       <DialogContentText>You can extend your session to continue working</DialogContentText>
       <div className={classes.buttonBar}>
         <Button>Extend Session</Button>
+        {/* For this specific dialog, we need to use a mousedown to handle the log out before the timeout is reset */}
         <Button onMouseDown={onSignOut}>Log Out</Button>
       </div>
     </DialogContent>
@@ -99,7 +100,9 @@ const InactivityTimer = ({ id, timeout, countdownStart, onSignOut }) => {
       setLastActive(id, Date.now());
     }
 
-    // Listen on the capture phase of the event, otherwise the order of events in a deliberate log out will be incorrect
+    // Listen on the capture phase of the event, otherwise the order of events in a deliberate log out will be incorrect.
+    // Without capture, if the user manually logs out, the last updated time will be recorded after the click to logout
+    // resulting in a last active time getting set in local storage.
     _.forEach((event) => document.addEventListener(event, updateLastActive, true), targetEvents);
 
     return () => {
@@ -133,7 +136,6 @@ export const LogoutIframe = ({ id, dismissLogout }) => (
     src="https://www.google.com/accounts/Logout"
     onLoad={() => {
       setLastActive(id);
-      setTimeout(() => window.location.reload(), 500);
       dismissLogout();
     }}
   />
--- a/src/components/IdleStatusMonitor.jsx
+++ b/src/components/IdleStatusMonitor.jsx
@@ -1,6 +1,4 @@
 import _ from 'lodash/fp';
-import * as qs from 'qs';
-import { createBrowserHistory } from 'history';
 import React, { useEffect, useState } from 'react';
 import { lastActiveTimeStore } from 'libs/state'; // stub
 import * as Utils from 'libs/utils';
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -49,6 +49,7 @@
     "modernizr": "^3.11.8",
     "moment": "^2.29.1",
     "prop-types": "^15.7.2",
+    "qs": "^6.10.1",
```
petesantos: Handles parsing of query parameters

src/components/IdleStatusMonitor.jsx
```diff
@@ -0,0 +1,226 @@
...
+));
+
+CountdownModal.propTypes = {
+  countdown: PropTypes.number,
+  onSignOut: PropTypes.func,
+};
+
+const InactivityTimer = ({ id, timeout, countdownStart, onSignOut }) => {
+  const { [id]: lastRecordedActivity } = Utils.useStore(lastActiveTimeStore) || {};
+  const [currentTime, setDelay] = Utils.useCurrentTime();
```
petesantos, line 84: This hook allows us to set a polling delay. This component will re-render for each delay tick. The delay changes on whether or not the countdown needs to be rendered.

src/components/IdleStatusMonitor.jsx
```diff
@@ -0,0 +1,226 @@
...
+  const { timedOut, showCountdown, countdown } = getIdleData({
+    currentTime,
+    lastRecordedActivity,
+    timeout,
+    countdownStart,
+  });
+
+  setDelay(showCountdown ? 1000 : Math.max(250, countdown - countdownStart));
+
+  Utils.useOnMount(() => {
```
petesantos, line 94: Global event listeners are added to check for activity

src/components/IdleStatusMonitor.jsx
```diff
@@ -0,0 +1,226 @@
...
+
+    return () => {
+      _.forEach(
+        (event) => document.removeEventListener(event, updateLastActive, true),
+        targetEvents,
+      );
+    };
+  });
+
+  useEffect(() => {
```
petesantos, line 115: Upon a timeout we will sign the user out

src/containers/App.jsx
```diff
@@ -163,9 +164,17 @@ export function App(props) {
     setAnchorEl(null);
   };
 
+  const signOut = () => {
```
petesantos, line 169: Moved from line 208 for reuse. Added a condition to log the user out if they are part of a timeout group (clinical user).

src/libs/browser-storage.js
```diff
@@ -0,0 +1,89 @@
+import _ from 'lodash/fp';
```
petesantos, line 1: This code was pulled from terra-ui. These are abstractions around browser storage designed to prevent issues in the event that storage has filled up. They also provide functionality allowing us to listen to storage changes

src/libs/utils.js
```diff
@@ -9,6 +10,118 @@ export const useOnMount = (fn) => {
   useEffect(fn, []);
 };
 
-export const validateSnapshotName = (name) => {
-  return name.length >= 1 && name.length <= 63 && name.match('^[a-zA-Z0-9][_a-zA-Z0-9]*$');
+export const validateSnapshotName = (name) =>
+  name.length >= 1 && name.length <= 63 && name.match('^[a-zA-Z0-9][_a-zA-Z0-9]*$');
+
+const maybeCall = (maybeFn) => (_.isFunction(maybeFn) ? maybeFn() : maybeFn);
```
petesantos, line 16: The functions below were taken from terraUI as helpers to make the idle status monitor migration easier

src/modules/auth.js
```diff
@@ -26,17 +29,36 @@ function describeUser(user) {
...
+  }
+}
+
 // return a user object that is not tied to the Google API
 export function getUser(options) {
   return new Promise((resolve, reject) => {
     init(options)
       .then((GoogleAuth) => {
         const user = GoogleAuth.currentUser.get();
+        GoogleAuth.currentUser.listen(dispatchCurrentUser);
```
petesantos, line 62: Added a listener to handle auth state changes. This prevents syncing issues between tabs and also improves existing UX and fixes some bugs when signing in or out. When a users signs in, they no longer have to click sign in on every tab. When they sign out of one tab, they will be signed out of all tabs.

src/reducers/user.js
```diff
@@ -28,6 +30,7 @@ export default {
           email: { $set: action.payload.email },
           token: { $set: action.payload.token },
           tokenExpiration: { $set: action.payload.tokenExpiration },
+          id: { $set: action.payload.id },
```
petesantos, line 33: ID is used to set the local storage key for the timeout

snf2ye (approved):
Thanks for the helpful comments. LGTM.

src/components/IdleStatusMonitor.jsx
```diff
@@ -0,0 +1,183 @@
...
+    color: theme?.palette.primary.dark,
+    whiteSpace: 'pre',
+    textAlign: 'center',
+    fontSize: '4rem',
+  },
+});
+
+// Helpers
+const displayRemainingTime = (remainingSeconds) =>
+  _.join(':', [
```
nmalfroy: We have the `moment` library imported.  Feels like we should be using that for format durations/times, etc. (https://momentjs.com/)

src/libs/utils.js
```diff
@@ -9,6 +10,105 @@ export const useOnMount = (fn) => {
...
+
+export const maybeParseJSON = (maybeJSONString) => {
+  try {
+    return JSON.parse(maybeJSONString);
+  } catch {
+    return undefined;
+  }
+};
+
+export const durationToMillis = ({ hours = 0, minutes = 0, seconds = 0 }) =>
```
nmalfroy: Again, I think moment can do that for us

nmalfroy (approved):
Cool, I think you may want to use the moment library for some of the time math / display but looks great!

src/components/IdleStatusMonitor.jsx
petesantos: Switched to `moment` - Thanks!

nmalfroy (approved):
Awesome, thanks for making the changes!

PR #1040, opened 2021-12-10
DR-2312 Pass location along when paginating BQ queries

nmalfroy:
So we don't get 404's when the location isn't in US


samanehsan (approved):
Lgtm! :octocat:

tlangs (approved):
Thanks!

fboulnois (approved):

PR #1045, opened 2022-01-10
DR-2340 Consolidate Dataset and Snapshot tabs into Dashboard tabs

myessail:
tabswitching

### DIFF
--- a/src/components/HomeView.jsx
+++ b/src/components/HomeView.jsx
@@ -11,6 +11,12 @@ import DatasetView from './DatasetView';
 import SnapshotView from './SnapshotView';
 
 const styles = (theme) => ({
+  pageRoot: {
+    padding: '16px 24px',
+    display: 'flex',
+    flexDirection: 'column',
+    height: '100%',
+  },
   header: {
     alignItems: 'center',
     color: theme.typography.color,
@@ -38,16 +44,6 @@ const styles = (theme) => ({
     fontSize: '1.5rem',
     fontWeight: 700,
     paddingBottom: '1rem',
-  },
-  wrapper: {
-    display: 'flex',
-    fontFamily: theme.typography.fontFamily,
-    justifyContent: 'center',
-    padding: theme.spacing(4),
-    margin: theme.spacing(4),
-  },
-  width: {
-    ...theme.mixins.containerWidth,
   },
   tabsRoot: {
     color: '#333F52',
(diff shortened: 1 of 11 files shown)


src/components/HomeView.jsx
```diff
@@ -32,24 +40,40 @@ const styles = (theme) => ({
...
+    transition: 'none',
   },
 });
 
 class HomeView extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      tabValue: '/datasets',
```
nmalfroy: I'm not sure that we want to use component state...shouldn't we just be using the redux store.  Component state should, in my opinion, be used exceedingly rarely

src/routes/Private.jsx
```diff
@@ -115,13 +82,10 @@ class Private extends React.Component {
                   <div className={classes.component}>
                     <Switch>
                       <Route exact path="/" component={HomeView} />
-                      {features.searchui && (
-                        <Route exact path="/explorer" component={DataExplorerView} />
-                      )}
-                      <Route exact path="/datasets" component={DatasetsView} />
+                      <Route exact path="/datasets" component={HomeView} />
+                      <Route exact path="/snapshots" component={HomeView} />
                       <Route exact path="/datasets/:uuid/query" component={QueryView} />
```
nmalfroy, line 90: presumably, we'll want these paths to live under `HomeView` too

src/components/HomeView.jsx
nmalfroy: Can't we get that from the router value from the global store?

src/components/HomeView.jsx
```diff
@@ -32,59 +39,80 @@ const styles = (theme) => ({
...
-    const { dispatch } = this.props;
-    dispatch(getDatasets(5));
-    dispatch(getSnapshots(5));
-  }
+  static prefixMatcher = new RegExp('/[^/]*');
 
   render() {
-    const { classes, datasets, features, snapshots } = this.props;
+    const { classes, location } = this.props;
+    const tabValue = HomeView.prefixMatcher.exec(location.pathname)[0];
```
nmalfroy, line 77: Does this make sense as part of a component? I'm also surprised this isn't something that's part of the router

nmalfroy (approved):
looks good to me! (modulo linting/test failures)

src/components/HomeView.jsx
myessail: I think if we want the tabs in the HomeView then we would need to handle the values here in some way, this is getting the location from the global state but I could try passing it through as a prop from the router if you think that might be a better approach?

src/routes/Private.jsx
myessail: I don't think we need to since these pages aren't under the datasets/snapshots tab in the new mocks

snf2ye (approved):

PR #1048, opened 2022-01-12
DR-2341 Update table stylings to more closely match Terra UI

tlangs:
Updated table stylings to be more in-line with Terra's based on [Lou's design](https://projects.invisionapp.com/share/KQ124BKIZGC6#/screens/461773905). I wasn't able to get all the way there, due to some limitations of MaterialUI, but its pretty close, save the pagination icons and layout. 

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "jade-data-repo-ui",
-  "version": "0.90.0",
+  "version": "0.93.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -2963,12 +2963,6 @@
             "node-releases": "^1.1.66"
           }
         },
-        "caniuse-lite": {
-          "version": "1.0.30001159",
-          "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001159.tgz",
-          "integrity": "sha512-w9Ph56jOsS8RL20K9cLND3u/+5WASWdhC/PPrf+V3/HsM3uHOavWOR1Xzakbv4Puo/srmPHudkmCRWM7Aq+/UA==",
-          "dev": true
-        },
         "electron-to-chromium": {
           "version": "1.3.602",
           "resolved": "https://registry.npmjs.org/electron-to-chromium/-/electron-to-chromium-1.3.602.tgz",
@@ -5510,9 +5504,9 @@
       }
     },
     "caniuse-lite": {
-      "version": "1.0.30001123",
-      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001123.tgz",
-      "integrity": "sha512-03dJDoa4YC4332jq0rqwiM+Hw6tA5RJtrnZKvOQy7ASoIUv8CinkcmGhYpCvCjedvkBQrrKnkcELxrUSW/XwNQ=="
+      "version": "1.0.30001298",
+      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001298.tgz",
+      "integrity": "sha512-AcKqikjMLlvghZL/vfTHorlQsLDhGRalYf1+GmWCf5SCMziSGjRYQW/JEksj14NaYHIR6KIhrFAy0HV5C25UzQ=="
     },
     "capture-exit": {
       "version": "2.0.0",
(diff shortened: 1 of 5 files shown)


src/components/table/LightTable.jsx
```diff
@@ -15,7 +15,7 @@ import LightTableHead from './LightTableHead';
 
 const styles = (theme) => ({
   root: {
-    border: `1px solid ${theme.palette.primary.dark}`,
+    border: `1px solid #e8eaeb`,
```
myessail: if this value isn't already there, we should add it to `theme.js` so we can reuse it elsewhere

src/components/table/LightTable.jsx
```diff
@@ -30,6 +30,20 @@ const styles = (theme) => ({
   row: {
     borderRadius: `${theme.shape.borderRadius}px ${theme.shape.borderRadius}px 0 0`,
   },
+  evenRow: {
+    backgroundColor: 'white',
+  },
+  oddRow: {
+    backgroundColor : 'rgba(233,236,239,0.4)'
```
myessail: just to be consistent with other places in the code, i would probably have these as hex values and then set the `opacity` separately

src/components/table/LightTable.jsx
```diff
@@ -30,6 +30,20 @@ const styles = (theme) => ({
...
+  },
+  paginationButton:{
+    borderRadius: `${theme.shape.borderRadius}px`,
+    margin: '0px 2px',
+    transition: null,
+    padding: '0.25rem',
+    border: '1px solid #4d72aa',
```
myessail: would add this color to `theme.js` as well

src/components/table/LightTableHead.jsx
```diff
@@ -6,17 +6,22 @@ import TableHead from '@material-ui/core/TableHead';
 import TableRow from '@material-ui/core/TableRow';
 import Tooltip from '@material-ui/core/Tooltip';
 import TableSortLabel from '@material-ui/core/TableSortLabel';
+import {Sort} from "@material-ui/icons";
 
 const styles = (theme) => ({
   head: {
     color: theme.palette.primary.dark,
-    backgroundColor: theme.palette.primary.light,
+    backgroundColor: 'rgba(233,236,239,0.4)',
```
myessail: same as previous comments

src/components/table/LightTableHead.jsx
```diff
@@ -6,17 +6,22 @@ import TableHead from '@material-ui/core/TableHead';
...
   },
   cell: {
     color: theme.palette.secondary.dark,
     minWidth: 200,
-    fontWeight: 500,
+    fontSize: '14px',
+    fontWeight: 600,
+    letterSpacing: 0,
+    lineHeight: '16px',
+    border: `1px solid #e8eaeb`
```
myessail: same as previous comments

myessail (changes requested):
looks good so far, i think it might look good to add a bit more padding to the right of the next button 🙂

tlangs:
TIL I learned about `theme.js` 🤣

tlangs:
@myessail how does this look? 
image

package-lock.json
```diff
@@ -2963,12 +2963,6 @@
             "node-releases": "^1.1.66"
           }
         },
-        "caniuse-lite": {
```
nmalfroy, line 2966: I'm surprised to see caniuse bumped automatically like this...not sure what triggered that?

src/components/table/LightTable.jsx
```diff
@@ -15,7 +15,7 @@ import LightTableHead from './LightTableHead';
 
 const styles = (theme) => ({
   root: {
-    border: `1px solid ${theme.palette.primary.dark}`,
+    border: '1px solid #e8eaeb',
```
nmalfroy: I think you want to pull this form theme.js no?

src/components/table/LightTable.jsx
```diff
@@ -172,8 +186,11 @@ export class LightTable extends React.PureComponent {
             />
             <TableBody>
               {rows && rows.length > 0 ? (
-                rows.map((row) => (
-                  <TableRow key={rowKey ? rowKey(row) : row.id} className={classes.row}>
+                rows.map((row, index) => (
+                  <TableRow
+                    key={rowKey ? rowKey(row) : row.id}
+                    className={`${classes.row} ${index % 2 ? classes.oddRow : classes.evenRow}`}
```
nmalfroy: Rather than spacing out classnames, you should use the `clsx` class to include logic like this (checkout `SidebarDrawer.jsx` for an example)

nmalfroy (changes requested):
Looks good!  One left over constant and you the clsx thing

myessail (approved):
LGTM assuming you fix muscles' comments 👍

package-lock.json
tlangs: I think I might have run some sort of npm update or something

nmalfroy (approved):
nice!

PR #1051, opened 2022-01-14
DR-2343 help widget

snf2ye:
Latest should be deployed here: https://jade-sh.datarepo-dev.broadinstitute.org/

image
image


src/components/help/HelpContainer.jsx
```diff
@@ -0,0 +1,123 @@
...
+            <div className={classes.anotherContainer}>
+              <div className={classes.close} onClick={this.handleHelpExitClick}>
+                <ExitSVG />
+              </div>
+              <HelpPanel />
+              <div className={classes.footer}>
+                <p>
+                  <b>Not finding what you are looking for?</b>
+                  <br />
+                  Visit the <a className={classes.jadeLink} href="https://support.terra.bio/hc/en-us/sections/4407099323675-Terra-Data-Repository"  target="_blank" rel="noopener noreferrer">Terra Support Hub</a> or <a className={classes.jadeLink} href="https://support.terra.bio/hc/en-us/requests/new" target="_blank" rel="noopener noreferrer">contact us</a>.
```
snf2ye: I don't know how to fix this lint error while keeping the functionality in place

src/components/help/HelpContainer.jsx
```diff
@@ -0,0 +1,141 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { withStyles } from '@material-ui/core/styles';
+import { ReactComponent as ExitSVG } from 'media/icons/times-light.svg';
+import { ReactComponent as QuestionCircleSVG } from 'media/icons/question-circle-solid.svg';
+import HelpPanel from './HelpPanel';
+
+const styles = (theme) => ({
+  panelContainer: {
+    color: '#333F52',
```
nmalfroy: I think we want colors like this to live in theme.js (if it's not there already)

src/components/help/HelpContainer.jsx
```diff
@@ -0,0 +1,141 @@
...
+    fontFamily: theme.typography.fontFamily,
+    right: '0',
+    'max-width': '815px',
+    height: 'calc(100% - 64px)', // 64px = height of top nav bar
```
nmalfroy: Is that a value we can pull from somewhere

src/components/help/HelpContainer.jsx
```diff
@@ -0,0 +1,141 @@
...
+    position: 'fixed',
+    'background-color': '#E9ECEF',
```
nmalfroy: same thing, we should probably have like a panel background var

src/components/help/HelpContainer.jsx
```diff
@@ -0,0 +1,141 @@
...
+  },
+  helpOpen: {
+    height: '48px',
+    width: '48px',
+    position: 'absolute',
+    right: 0,
+    ...theme.mixins.jadeLink,
+  },
+  helpIcon: {
+    padding: '8px',
```
nmalfroy, line 55: You need:
```
width: '100%',
height: '100%'
```

to fix the issue in FF.  This makes it take the 48px that you set above.

Also, can you add ```cursor: 'pointer'``` here too?

src/components/help/HelpContainer.jsx
```diff
@@ -0,0 +1,141 @@
...
+            <div className={classes.anotherContainer}>
+              <div className={classes.close} onClick={this.handleHelpExitClick}>
+                <ExitSVG />
+              </div>
+              <HelpPanel />
+              <div className={classes.footer}>
+                <p>
+                  <b>Not finding what you are looking for?</b>
+                  <br />
+                  Visit the&nbsp;
```
nmalfroy: I'm curious, do you need these spaces? (fine to leave them, I just wanted to make sure we actually need that spacing)

nmalfroy (commented):
great progress!  couple things to help with the FF issue

src/components/help/HelpContainer.jsx
snf2ye: Definitely - without `&nbsp;` we end up with "Visit theTerra Support Huborcontact us." Plus, it's the only option that makes the linter happy (it doesn't like the trailing spaces).

src/components/help/HelpContainer.jsx
snf2ye: I added a constant in theme to store this.. but I'm not sure that makes sense. I don't see the direct height set anywhere. Thoughts?

nmalfroy (approved):
looks great!

src/components/help/HelpContainer.jsx
```diff
@@ -0,0 +1,108 @@
...
+    width: '100%',
+    height: '100%',
+  },
+});
+
+class HelpContainer extends React.Component {
+  constructor(props) {
+    super(props);
+    this.handleHelpButtonClick = this.handleHelpButtonClick.bind(this);
+    this.handleHelpExitClick = this.handleHelpExitClick.bind(this);
```
myessail: i believe you can get away with not doing this if you write these as arrow functions, ie `handleHelpButtonClick = () => {}` on line 73

myessail (approved):

src/components/help/HelpContainer.jsx
nmalfroy: ah great call!

PR #1057, opened 2022-01-21
DR-2342: Update Search Bar

snf2ye:
I might have started switching components from class to functional 😎 

 Changes deployed here: https://jade-sh.datarepo-dev.broadinstitute.org/datasets


nmalfroy:
Yikes!  24L lines!?  I'm wondering what happened to package.json?

src/components/table/SearchTable.jsx
```diff
@@ -0,0 +1,63 @@
...
+import { withStyles } from '@material-ui/core/styles';
+import InputBase from '@material-ui/core/InputBase';
+import { ReactComponent as SearchSVG } from 'media/icons/search_icon.svg';
+import { fade } from '@material-ui/core/styles/colorManipulator';
+
+const styles = (theme) => ({
+  search: {
+    height: 45,
+    width: '100%',
+    border: '1px solid #AEB3BA',
```
nmalfroy: should these colors be in theme.js?  Not sure if they get used a lot

src/components/HomeView.jsx
```diff
@@ -62,33 +65,53 @@ const styles = (theme) => ({
...
   },
+  titleAndSearch: {
+    display: 'flex',
+    'margin-top': '2em',
+    'margin-bottom': '2em',
+  },
 });
 
 class HomeView extends React.PureComponent {
+  constructor(props) {
```
nmalfroy, line 76: I think this will cause it to never reset the search?  Do you happen to have a version up and running?  I think this needs to get reset when you change locations.  That's one more reason I was thinking it makes sense to have this tracked in the central store.  So we can, for instance, listen on the `location.change` action and reset the search value then.

src/components/table/LightTable.jsx
```diff
@@ -132,14 +114,14 @@ export class LightTable extends React.PureComponent {
...
-    const { handleEnumeration } = this.props;
-    const { page, orderDirection, orderBy, rowsPerPage } = this.state; // limit
-    const offset = page * rowsPerPage;
-    const searchString = event.target.value;
-    this.setState({ searchString }); // filter
-    handleEnumeration(rowsPerPage, offset, orderBy, orderDirection, searchString);
-  };
+  componentDidUpdate(prevProps) {
+    const { handleEnumeration, searchString } = this.props;
+    if (prevProps.searchString !== searchString) {
```
nmalfroy: Again, it feels like we should be handling this through an action (not sure if handleEnumeration does that under the covers...I don't have the code handy right now)

Also...and I think it's likely beyond the scope of this PR...I was wondering if we could see about making this debounce to search rather than on every keystroke...that could end up hammering the API

src/modules/theme.js
```diff
@@ -123,6 +123,7 @@ export default createMuiTheme({
   },
   shape: {
     borderRadius: 5,
+    searchBorderRadius: 35,
```
nmalfroy: This depends on the height of the searchBox...might make sense to set it to `100%` which has the same effect

src/components/table/LightTable.jsx
```diff
@@ -27,201 +27,167 @@ const styles = (theme) => ({
...
-          )}
-        </Paper>
-      </div>
+        : 0,
     );
-  }
+    handleEnumeration(rowsPerPage, offset, orderBy, orderDirection, searchString);
+  }, [searchString, page, rowsPerPage, orderDirection]);
+
+  return (
```
nmalfroy, line 100: Look at you with your fancy functional component :)

src/components/table/SearchTable.jsx
```diff
@@ -9,10 +9,10 @@ const styles = (theme) => ({
   search: {
     height: 45,
     width: '100%',
-    border: '1px solid #AEB3BA',
-    backgroundColor: '#F1F4F8',
+    border: `1px solid ${theme.palette.common.border}`,
```
nmalfroy, line 13: haha, that'll teach me to go commit by commit :)

nmalfroy (commented):
This is nice!  I had a couple of questions that it might make sense to talk about face to face.  Do you happen to have this deployed on your env?  I'd love to check behavior

src/components/HomeView.jsx
snf2ye: It's deployed here - https://jade-sh.datarepo-dev.broadinstitute.org/datasets

Right now, the search string is persistent between the snapshot and dataset page and the results are correspondingly updated when dataset/snapshot tab is toggled. I think this makes sense as the user workflow, but we can discuss as a team & w/ Lou.

snf2ye:
> Yikes! 24L lines!? I'm wondering what happened to package.json?

@nmalfroy  - In my experience, it's fairly normal for the package json lock file to have a lot of changes when `npm install` is run, even if there isn't any changes to package.json. I don't think it's something to worry about, but we can make it a team norm to avoid checking it in if that is preferred.

src/components/table/LightTable.jsx
snf2ye: I would think that depends on whether we are storing the search string in the redux store, right?

Also, please take a look at the latest commit - this changes w/ the functional component and the useEffect hook so that we no longer need to check if the search string is different.

src/components/table/LightTable.jsx
```diff
@@ -45,216 +41,153 @@ const styles = (theme) => ({
...
-                          {col.render ? col.render(row) : row[col.property]}
-                        </TableCell>
-                      ))}
-                    </TableRow>
-                  );
-                })
-              ) : (
-                <TableRow className={classes.row}>
-                  <TableCell colSpan={columns.length}>
-                    No {itemType} have been created yet
```
nmalfroy, line 212: Making a quick note so it's top of mind: if we filter down to no found datasets or snapshots, we show this, instead of `No {itemType} match your filter` (which I could have sworn we fixed when we added the totalRows field but perhaps not)

src/modules/theme.js
snf2ye: 100% for the border-radius? That looks like this:
image

src/components/HomeView.jsx
nmalfroy: Ah cool, thanks, this is helpful.  Interesting...I'm kinda torn.  I kinda like the idea of separate filters but I'd be curious for lou's take

src/modules/theme.js
snf2ye: But, I can pull '35px' out of theme.

nmalfroy: Hahaha yikes!  Clearly that's not the way :) Ah wait, I'm getting mixed up with creating css circles...never mind

src/components/table/LightTable.jsx
snf2ye: ok, fix pushed.

nmalfroy (approved):
Looks great!  I'm curious about if we want independent behavior between datasets and snapshots but that could be a separate ticket

myessail (approved):
LGTM 🙂 i think we might want to only check in `package-lock.json` files if we change `package.json`, can be something we discuss with the team though

snf2ye:
> LGTM 🙂 i think we might want to only check in `package-lock.json` files if we change `package.json`, can be something we discuss with the team though

Okay, thanks - I went ahead and reverted my changes to the package lock file.

PR #1058, opened 2022-01-24
DR-2402: Add excluded label for Fields marked as excluded

snf2ye:
image


nmalfroy (approved):
looks good!  I'd love to revisit the UX in general around but this is muuuuch better

snf2ye:
> looks good! I'd love to revisit the UX in general around but this is muuuuch better

Agreed - it currently does not have a great user experience.

myessail (approved):

PR #1060, opened 2022-01-25
DR-2341 Make dataset and snapshot tables render cloud platform the same way

tlangs:
I forgot to update the SnapshotTable, so I made the DatasetTable render method a utility method and used it for both tables.


src/libs/render-utils.jsx
```diff
@@ -0,0 +1,15 @@
...
+const cloudPlatforms = { gcp: 'Google Cloud Platform', azure: 'Microsoft Azure' };
+/**
+ * Render the cloud platform cell contents of a table
+ * @param row
+ * @returns {JSX.Element}
+ */
+export const renderCloudPlatforms = (row) => (
+  <div>
+    {Array.from(new Set(row.storage.map((s) => cloudPlatforms[s.cloudPlatform]))).map((c) => (
+      <div>{c}</div>
```
nmalfroy: you need a key on the div (any time you're working with arrays in React)

You'll probably notice the console warning

nmalfroy (changes requested):
I think you need a key

snf2ye (approved):

src/libs/render-utils.jsx
tlangs: key?

snf2ye: Oh yeah, I think I ran into that error message.  Fix here - https://github.com/DataBiosphere/jade-data-repo-ui/pull/1057/files#diff-d2950d9d693e1a22b29406a1ec27e9905d0f794a113ca9b9f897f9dfb929996cR81

tlangs: Added

src/libs/render-utils.jsx
```diff
@@ -8,8 +8,10 @@ const cloudPlatforms = { gcp: 'Google Cloud Platform', azure: 'Microsoft Azure'
  */
 export const renderCloudPlatforms = (row) => (
   <div>
-    {Array.from(new Set(row.storage.map((s) => cloudPlatforms[s.cloudPlatform]))).map((c) => (
-      <div>{c}</div>
-    ))}
+    {Array.from(new Set(row.storage.map((s) => cloudPlatforms[s.cloudPlatform]))).map(
+      (c, index) => (
+        <div key={index.toString()}>{c}</div>
```
nmalfroy, line 13: This is fine here (since you're guaranteed that index will have a value) but I often do:
`<div key={`${index}`}...` to avoid NPEs

nmalfroy (approved):
cool, thanks for addressing!

PR #1061, opened 2022-01-26
DR-2342 Bug Fix - searchString undefined causing snapshot detail pages not to load

snf2ye:
Console error message:
image

### DIFF
--- a/src/components/SnapshotDetailView.jsx
+++ b/src/components/SnapshotDetailView.jsx
@@ -135,6 +135,7 @@ export class SnapshotDetailView extends React.PureComponent {
               datasetsCount={snapshot.source.length}
               features={features}
               handleFilterDatasets={this.handleFilterDatasets}
+              searchString=""
             />
           )}
         </div>
--- a/src/components/SnapshotDetailView.jsx
+++ b/src/components/SnapshotDetailView.jsx
@@ -100,7 +100,7 @@ export class SnapshotDetailView extends React.PureComponent {
     const { snapshot } = this.props;
     const datasets = snapshot.source.map((s) => s.dataset);
     const filtered = datasets.filter((d) =>
-      d.name.toLowerCase().includes((searchString || '').toLowerCase(),
+      d.name.toLowerCase().includes((searchString || '').toLowerCase()),
     );
     const sorted = _.orderBy(filtered, sort, sortDirection);
     const paged = _.take(_.drop(sorted, offset), limit);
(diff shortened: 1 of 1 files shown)


src/components/SnapshotDetailView.jsx
```diff
@@ -135,6 +135,7 @@ export class SnapshotDetailView extends React.PureComponent {
               datasetsCount={snapshot.source.length}
               features={features}
               handleFilterDatasets={this.handleFilterDatasets}
+              searchString={''}
```
nmalfroy: I think you can just have
```suggestion
              searchString=""
```

nmalfroy (approved):
Lol...I was literally commenting when you pushed a commit :). Looks good.  I wonder if it'd be also safer to fix where it calls `.toLowerCase` to defend against the npe?

snf2ye:
> Lol...I was literally commenting when you pushed a commit :). Looks good. I wonder if it'd be also safer to fix where it calls `.toLowerCase` to defend against the npe?

@nmalfroy - If we're defining it to start, it really should never become undefined again, but I don't think it hurts to add a null check. I've added a ternary - is that what you were suggesting?

src/components/SnapshotDetailView.jsx
```diff
@@ -100,7 +100,7 @@ export class SnapshotDetailView extends React.PureComponent {
     const { snapshot } = this.props;
     const datasets = snapshot.source.map((s) => s.dataset);
     const filtered = datasets.filter((d) =>
-      d.name.toLowerCase().includes(searchString.toLowerCase()),
+      d.name.toLowerCase().includes(searchString ? searchString.toLowerCase() : ''),
```
nmalfroy: thanks!  I agree right now it can't but if ever the value gets programmatically set to null (e.g. if we move to back it with a store and inadvertently set it it null).  I just like the safety of it.
I typically do something like:
```suggestion
      d.name.toLowerCase().includes((searchString || '').toLowerCase(),
```

or I think you could even do something like:
```suggestion
      d.name.toLowerCase().includes(searchString?.toLowerCase(),
```

snf2ye: Oh right - I think the second (null coalescing)  wouldn't work in our case b/c if searchString is undefined, then it would return undefined and the .includes() call would probably  be unhappy w/ that input. 
The `or` operator should work.

src/components/SnapshotDetailView.jsx
```diff
@@ -100,7 +100,7 @@ export class SnapshotDetailView extends React.PureComponent {
     const { snapshot } = this.props;
     const datasets = snapshot.source.map((s) => s.dataset);
     const filtered = datasets.filter((d) =>
-      d.name.toLowerCase().includes((searchString || '').toLowerCase(),
+      d.name.toLowerCase().includes((searchString || '').toLowerCase()),
```
nmalfroy, line 103: Though I will say, FWIW, you could run into something where you have:
`(false||'').toLowerCase()`
but again, that's if we messed something programatically.

Typescript would be nice for preventing that

myessail (approved):

PR #1063, opened 2022-01-26
DR-2348 Dataset Roles and Members

tlangs:
Updated the info panel for datasets to show all Stewards, Custodians, and Snapshot Creators. Also implemented the help panel within the roles accordion, filled with text from the "Getting Started in TDR" document.

This is more React work than I've done in a while, so please let me know what can be done to improve the code!

Screen Shot 2022-01-26 at 5 10 35 PM
Screen Shot 2022-01-26 at 5 10 44 PM


src/components/dataset/query/sidebar/panels/DatasetAccess.jsx
```diff
@@ -0,0 +1,105 @@
...
+    const snapshotCreators = (snapshotCreatorsObj && snapshotCreatorsObj.members) || [];
+    const stewardsObj = policies.find((policy) => policy.name === 'steward');
+    const stewards = (stewardsObj && stewardsObj.members) || [];
+
+    return (
+      <div className={classes.root}>
+        <div>
+          <Link href="#" className={classes.blueLink} onClick={helpOverlayToggleWithContent}>
+            Learn more
+          </Link>{' '}
```
nmalfroy: I think we typically user `&nbsp;` for whitespace like this

src/components/dataset/query/sidebar/panels/DatasetAccess.jsx
```diff
@@ -0,0 +1,105 @@
...
+      </Typography>
+    </Fragment>
+  );
+
+  render() {
+    const { classes, helpOverlayToggle, policies } = this.props;
+    const helpOverlayToggleWithContent = () =>
+      helpOverlayToggle(DatasetAccess.helpTitle, DatasetAccess.helpContent);
+
+    const custodiansObj = policies.find((policy) => policy.name === 'custodian');
```
nmalfroy: might be worth adding these to `constants/index.js`

src/components/dataset/query/sidebar/panels/DatasetAccess.jsx
```diff
@@ -0,0 +1,105 @@
...
+    const custodians = (custodiansObj && custodiansObj.members) || [];
+    const snapshotCreatorsObj = policies.find((policy) => policy.name === 'snapshot_creator');
+    const snapshotCreators = (snapshotCreatorsObj && snapshotCreatorsObj.members) || [];
+    const stewardsObj = policies.find((policy) => policy.name === 'steward');
+    const stewards = (stewardsObj && stewardsObj.members) || [];
+
+    return (
+      <div className={classes.root}>
+        <div>
+          <Link href="#" className={classes.blueLink} onClick={helpOverlayToggleWithContent}>
```
nmalfroy: The `#` might mess up our routing.  It's cleaner to to handle this with an action.  So there would be an action that triggers saving the 'is expanded' state in the central store and then the component can just pull that information from that store.

I don't love that we're using component state and not the main store (since now we have state scattered across different areas)

nmalfroy (approved):
Looks good!  Couple things in the comments...generally speaking, I'd love to avoid tracking state on components but now's probably not the time to change that.  Also, do you have this deployed in your env? it'd be nice to be able to click test

tlangs:
I went ahead and fixed the side drawer positioning. It was blocking the help icon.

tlangs:
Pushed latest code to https://jade-tl.datarepo-dev.broadinstitute.org

src/components/dataset/query/sidebar/panels/DatasetAccess.jsx
```diff
@@ -0,0 +1,105 @@
...
+    margin: theme.spacing(1),
+  },
+  paperBody: {
+    padding: theme.spacing(2),
+  },
+  headerText: {
+    fontSize: '14px',
+    lineHeight: '22px',
+    fontWeight: '600',
+    marginTop: theme.spacing(3),
```
myessail: is this being used anywhere? also, we have some preset text styling in theme.js that might be worth looking at

src/components/dataset/query/sidebar/panels/InfoView.jsx
```diff
@@ -3,94 +3,125 @@ import moment from 'moment';
...
+    this.toggleHelpOverlay = this.toggleHelpOverlay.bind(this);
+  }
+
   static propTypes = {
     classes: PropTypes.object,
     dataset: PropTypes.object,
     datasetPolicies: PropTypes.array,
   };
 
+  toggleHelpOverlay(helpTitle, helpContent) {
```
myessail: you can define this as ```toggleHelpOverlay = (helptitle, helpContent) => { }``` so you don't have to use `.bind` in the constructor

myessail (approved):
LGTM! few small things

PR #1064, opened 2022-01-26
DR-2403, DR-2181: Switch 'Export to Workspace' to use export by copy

snf2ye:
Deployed here: https://jade-sh.datarepo-dev.broadinstitute.org/snapshots/details/13380c1c-854a-4f48-a7bc-c0b63c0cf208
Let me know if you don't have access to this snapshot - I can add you.

### 1. Export to Workspace by Copy: Update UX + export 
- Add title above export button
- Tooltip on hover of "Export snapshot"
Screen Shot 2022-01-28 at 4 02 03 PM

- On click, button displays  Loading circle and "Preparing Snapshot"
Screen Shot 2022-01-28 at 4 02 16 PM
- When export is completed, button then displays "Snapshot ready - continue"
Screen Shot 2022-01-28 at 4 02 27 PM
- New window opens for terra workspace
Screen Shot 2022-01-28 at 4 02 38 PM

### 2. Snapshot Create Confirmation: 
* Remove Edit, share and export buttons
* Replace w/ "Go to snapshot details page" 
Screen Shot 2022-01-28 at 4 42 07 PM

### 3. Tooltips across TDR now use the new styling for the tooltip
image


myessail (approved):

nmalfroy:
I tried it out on your instance and it doesn't seem to work for me.  It triggers the export job but then I can't get to the big that does the opening of the new url?

src/components/DetailViewHeader.jsx
```diff
@@ -110,20 +131,39 @@ export class DetailViewHeader extends React.PureComponent {
...
+              <DialogContent>
+                <Typography variant="h5">Your data snapshot is being exported</Typography>
+                {/* TODO: Make this loading state more descriptive */}
+                <div className={clsx(classes.centered, classes.content)}>
+                  <CircularProgress />
+                </div>
+              </DialogContent>
+            </Dialog>
+            <Tooltip
+              title="Exporting a snapshot to a workspace means that all members of your workspace will be able to have read only access to the tables and files in the snapshot"
```
nmalfroy: I think this is different than what Lou had in the mocks...is that ok with him?

src/components/DetailViewHeader.jsx
```diff
@@ -110,20 +131,39 @@ export class DetailViewHeader extends React.PureComponent {
...
         </Grid>
       </Grid>
     );
   }
 }
 
-export default withStyles(styles)(DetailViewHeader);
+function mapStateToProps(state) {
+  return {
+    isOpen: state.snapshots.dialogIsOpen,
```
nmalfroy: You should use a different variable in the state.  This one if for the snapshot creation dialog.  I'd create an exportDialogIsOpen method to track it separately

src/sagas/repository.js
```diff
@@ -100,6 +100,38 @@ function* pollJobWorker(jobId, jobTypeSuccess, jobTypeFailure) {
...
+    yield call(
+      pollJobWorker,
+      jobId,
+      ActionTypes.EXPORT_SNAPSHOT_SUCCESS,
+      ActionTypes.EXPORT_SNAPSHOT_FAILURE,
+    );
+    const snapshots = yield select(getSnapshotState);
+    const manifestJsonPath = snapshots.exportResponse.format.parquet.manifest;
+    window.open(
+      `${terraUrl}#import-data?url=${window.location.origin}&snapshotId=${snapshotId}&format=tdrexport&snapshotName=${snapshotName}&tdrmanifest=${manifestJsonPath}`,
```
nmalfroy: OIC...I think this is a little weird from a UX standpoint.  I think we want the user to do a second click, no?  Perhaps as a link.  My popup blocker blocked this

nmalfroy: window.open in general is kind of a bad idea for that reason

nmalfroy (changes requested):
I think the UX is not quite right. Also, I noticed that if you hit export after the first one has finished it doesn't work...not sure why exactly but I suspect that it's related to state not getting totally reset

src/components/DetailViewHeader.jsx
snf2ye: The tooltip text? It should match the latest mocks: https://broadworkbench.atlassian.net/browse/DR-2403?focusedCommentId=55502

src/sagas/repository.js
snf2ye: Is there a way to do this in straight javascript that would allow it open automatically when the export finishes? Or do we have to do it as a second link, with an anchor component?

snf2ye:
> I tried it out on your instance and it doesn't seem to work for me. It triggers the export job but then I can't get to the big that does the opening of the new url?

It sounds like from your later comments that it did work, but it was blocked by the pop-up blocker. Is this correct?

snf2ye:
> Also, I noticed that if you hit export after the first one has finished it doesn't work...not sure why exactly but I suspect that it's related to state not getting totally reset

@nmalfroy - I'm not able to reproduce this issue - I'm able to export twice in a row

src/components/DetailViewHeader.jsx
nmalfroy: Oh sorry, I meant generally having a dialog

src/sagas/repository.js
nmalfroy: The later (which what I thought we wanted)

nmalfroy:
> > Also, I noticed that if you hit export after the first one has finished it doesn't work...not sure why exactly but I suspect that it's related to state not getting totally reset
> 
> @nmalfroy - I'm not able to reproduce this issue - I'm able to export twice in a row

I wonder if it's because it blocked the popup and that caused a JS failure.

I thought. that the original idea was to have the button change (and gray out) while the export was running and then become clickable on successful run and that would allow the user to link.  Did that change?

src/modules/bigquery.js
```diff
@@ -326,11 +326,13 @@ export default class BigQuery {
...
-        const prevTable = path[i - 1];
-        const relationship = this.findRelationshipData(schema, currTable, prevTable);
-        relationships.push(relationship);
+      if (path != null) {
+        for (let i = 1; i < path.length; i++) {
+          const currTable = path[i];
+          const prevTable = path[i - 1];
+          const relationship = this.findRelationshipData(schema, currTable, prevTable);
+          relationships.push(relationship);
+        }
```
snf2ye, line 335: I ran into an NPE here when trying to load a dataset details page - I don't totally understand why.

nmalfroy: Ah, I know.  I think it's possible to have the `relationships` response from a dataset or asset be undefined...which stinks.  Yeah, I think it's safer to make defend against it (e.g. 
```suggestion
      if (path != null) {
        for (let i = 1; i < path.length; i++) {
          const currTable = path[i];
          const prevTable = path[i - 1];
          const relationship = this.findRelationshipData(schema, currTable, prevTable);
          (relationships || []).push(relationship);
        }
```

src/components/DetailViewHeader.jsx
```diff
@@ -37,26 +49,38 @@ export class DetailViewHeader extends React.PureComponent {
     addSteward: PropTypes.func,
     canReadPolicies: PropTypes.bool,
     classes: PropTypes.object.isRequired,
+    dispatch: PropTypes.func.isRequired,
+    isProcessing: PropTypes.bool,
```
nmalfroy, line 56: Looks like you need a sort here

src/components/DetailViewHeader.jsx
```diff
@@ -110,20 +134,52 @@ export class DetailViewHeader extends React.PureComponent {
...
+                  className={classes.exportButton}
+                  variant="outlined"
+                  color="primary"
+                >
+                  Export snapshot
+                </Button>
+              </TerraTooltip>
+            )}
+            {isProcessing && !isDone && (
+              <Button className={classes.exportButton} variant="outlined" color="primary">
```
nmalfroy, line 161: You should also disable the button here to avoid double-clicking

src/modules/bigquery.js
snf2ye: It was failing on the "path.length" call, so "path" was null.

src/components/DetailViewHeader.jsx
```diff
@@ -37,26 +49,38 @@ export class DetailViewHeader extends React.PureComponent {
...
     readers: PropTypes.arrayOf(PropTypes.string),
     removeReader: PropTypes.func,
     terraUrl: PropTypes.string,
     removeSteward: PropTypes.func,
     stewards: PropTypes.arrayOf(PropTypes.string).isRequired,
   };
 
+  exportToWorkspaceCopy = () => {
+    const { dispatch, of } = this.props;
+    dispatch(exportSnapshot(of.id));
```
nmalfroy, line 68: We should test timing here.  But for clicks that launch jobs, we should consider having something to track _just_ that the user has clicked to ensure that the user doesn't click again while the job is starting (maybe that could get handled inside the `exportSnapshot` action reducer)

src/components/DetailViewHeader.jsx
```diff
@@ -110,20 +134,52 @@ export class DetailViewHeader extends React.PureComponent {
...
+                  variant="outlined"
+                  color="primary"
+                >
+                  Export snapshot
+                </Button>
+              </TerraTooltip>
+            )}
+            {isProcessing && !isDone && (
+              <Button className={classes.exportButton} variant="outlined" color="primary">
+                <CircularProgress />
```
nmalfroy: Can you add a size constraint so that it doesn't make the button jump in size?

src/components/common/TerraTooltip.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
```
nmalfroy, line 1: Smart to add this as an explicit component

src/components/snapshot/SnapshotPopup.jsx
```diff
@@ -190,21 +190,7 @@ export class SnapshotPopup extends React.PureComponent {
...
-            </Button>
-            <Button color="primary" data-cy="exportSnapshot">
-              <a
-                target="_blank"
-                rel="noopener noreferrer"
-                href={`${terraUrl}/#import-data?url=${window.location.origin}&snapshotId=${snapshot.id}&snapshotName=${snapshot.name}&format=snapshot`}
-              >
-                Export
-              </a>
+              <Link to={`/snapshots/details/${snapshot.id}`}>Go to Snapshot Details Page</Link>
```
nmalfroy, line 215: 👍

src/reducers/job.js
```diff
@@ -29,6 +29,14 @@ export default {
         immutable(state, {
           jobStatus: { $set: STATUS.SUCCESS },
         }),
+      [ActionTypes.EXPORT_SNAPSHOT_JOB]: (state, action) =>
+        immutable(state, {
+          jobId: { $set: action.payload.jobId },
+        }),
+      [ActionTypes.EXPORT_SNAPSHOT_SUCCESS]: (state) =>
+        immutable(state, {
+          jobStatus: { $set: STATUS.SUCCESS },
```
nmalfroy, line 42: You may need a handler for failure.

src/reducers/snapshot.js
```diff
@@ -25,6 +25,10 @@ export const snapshotState = {
   dialogIsOpen: false,
   // for snapshot creation
   snapshotRequest: defaultSnapshotRequest,
+  // for snapshot export to workspace
+  exportIsProcessing: false,
```
nmalfroy, line 29: you could track exportIsRequested or submitted which would track when the user has clicked

nmalfroy (changes requested):
Oh wow, I like this UX a lot more.  Thanks for updating!  I have a couple of smallish things that would be great to have addressed

src/modules/bigquery.js
nmalfroy: oh weird...

src/components/DetailViewHeader.jsx
tlangs: It would be pretty easy for the click handler to set the element to `disabled`, and then have the success/failure action un-disable it.

src/components/dataset/query/sidebar/panels/FilterPanel.jsx
```diff
@@ -230,7 +230,7 @@ export class FilterPanel extends React.PureComponent {
             ))}
         </div>
         <div className={clsx(classes.rowTwo, classes.snapshotBtnCntnr)}>
-          <Tooltip title={canLink ? '' : billingErrorMessage}>
+          <TerraTooltip title={canLink ? '' : billingErrorMessage}>
```
tlangs, line 233: Why not conditionally render the `TerraTooltip` based on `canLink`?

src/reducers/job.js
tlangs: What are you talking about? Our API never fails!

src/components/SnapshotDetailView.jsx
```diff
@@ -108,7 +108,15 @@ export class SnapshotDetailView extends React.PureComponent {
...
+      features,
+      snapshot,
+      snapshotPolicies,
+      terraUrl,
+      canReadPolicies,
+      dispatch,
+    } = this.props;
     const { filteredDatasets } = this.state;
     const snapshotReadersObj = snapshotPolicies.find((policy) => policy.name === 'reader');
     const snapshotReaders = (snapshotReadersObj && snapshotReadersObj.members) || [];
```
tlangs, line 122: There's actually convenience methods for this now in `src/libs/utils.js#getRoleMembersFromPolicies`

tlangs (changes requested):

src/components/DetailViewHeader.jsx
snf2ye: Do you mean that these need to be sorted alphabetically? (just want to make sure I'm fixing the thing you're referencing)

nmalfroy: exactly

src/reducers/job.js
snf2ye: ohhh that's probably what's missing on snapshot create failure too, I was following the same action pattern - https://broadworkbench.atlassian.net/browse/DR-2181

src/components/DetailViewHeader.jsx
snf2ye: Edit: See other comment - https://github.com/DataBiosphere/jade-data-repo-ui/pull/1064#discussion_r796246892

src/components/DetailViewHeader.jsx
snf2ye: Edit: Actually - I've added an action that occurs right at the beginning of the export action that will trigger the "export" button to be hidden right away. I think any additional logic for disabling the button would be unnecessary.

src/reducers/job.js
snf2ye: I think I found the actual issue for DR-2181 (b/c I'm encountering the same issue for export) -- We weren't handling the case that the poll job worker had an exception. We were just showing the error message. I've changed it so that we're also handling the exception corresponding to either create snapshot or export snapshot.

src/components/dataset/query/sidebar/panels/FilterPanel.jsx
snf2ye: I think it's a little more complicated than that b/c the tooltip component wraps other components and we always want to render the child components. Conditionally rendering TerraTooltip as the parent component would hide the child component. 
This is also out of scope of this ticket - I'm just making sure the formatting for the tooltip remains consistent across the app.

src/components/SnapshotDetailView.jsx
snf2ye: Nice, sounds like something worth switching to - but, not in this PR.

src/components/DetailViewHeader.jsx
```diff
@@ -110,20 +141,57 @@ export class DetailViewHeader extends React.PureComponent {
...
+                  className={classes.exportButton}
+                  variant="outlined"
+                  color="primary"
+                >
+                  Export snapshot
+                </Button>
+              </TerraTooltip>
+            )}
+            {isProcessing && !isDone && (
+              <Button className={classes.exportButton} variant="outlined" color="primary">
```
nmalfroy, line 161: Could we make this disabled?  The button should never be clickable

src/components/DetailViewHeader.jsx
```diff
@@ -110,20 +141,57 @@ export class DetailViewHeader extends React.PureComponent {
...
-                target="_blank"
-                rel="noopener noreferrer"
-                href={`${terraUrl}/#import-data?url=${window.location.origin}&snapshotId=${of.id}&snapshotName=${of.name}&format=snapshot`}
+            <hr className={classes.separator} />
+            <Typography variant="h6" className={classes.section}>
+              Export a copy of the snapshot metadata to an exisiting or new Terra workspace
+            </Typography>
+            {!isProcessing && !isDone && (
+              <TerraTooltip title="Exporting a snapshot to a workspace means that all members of your workspace will be able to have read only access to the tables and files in the snapshot">
+                <Button
```
nmalfroy, line 150: I noticed that the button takes the whole width.  Is that ok?  Not a huge deal

nmalfroy (approved):
Looks great!  Minor nits that'd be great to fix if you've got a chance

src/components/DetailViewHeader.jsx
snf2ye: I'm checking w/ Lou. I needed to make the button a bit bigger anyway to accommodate the spin/preparing snapshot text, so I thought I would just match the other buttons in the panel.

src/components/DetailViewHeader.jsx
snf2ye: Ack, that messes with the styling..material UI styling. I'll work on fixing that up.

src/components/DetailViewHeader.jsx
```diff
@@ -110,20 +141,57 @@ export class DetailViewHeader extends React.PureComponent {
                 canManageUsers={true}
               />
             )}
-            <Button className={classes.exportButton} variant="contained" color="primary">
-              <a
-                target="_blank"
-                rel="noopener noreferrer"
-                href={`${terraUrl}/#import-data?url=${window.location.origin}&snapshotId=${of.id}&snapshotName=${of.name}&format=snapshot`}
+            <hr className={classes.separator} />
+            <Typography variant="h6" className={classes.section}>
```
tlangs, line 145: Minor quibble: I'm not sure that header styling is the right choice here. I'm thinking more italicized, grey text.

tlangs (approved):
One small thing, but otherwise, looks good!

src/components/DetailViewHeader.jsx
snf2ye: (discussed in standup)

PR #1065, opened 2022-01-27
DR-2405 Ability to manage users on datasets

tlangs:
Implemented the management of users on datasets. The styling probably needs to be updated, pending @lstrano-broad's mocks, and I should probably add an e2e test, but the functionality is there. Tested locally and on my personal deployment.

https://jade-tl.datarepo-dev.broadinstitute.org

image


src/components/dataset/query/sidebar/panels/InfoView.jsx
```diff
@@ -112,10 +111,7 @@ export class InfoView extends React.PureComponent {
                   Summary of roles and memberships
                 </AccordionSummary>
                 <AccordionDetails>
-                  <DatasetAccess
```
nmalfroy, line 115: On line 104 above (`<Accordion>`) would it be possible to add control to the expanded state through the dataset store? I ask because without it, I noticed an annoying behavior that when I click on help then close, it's collapsed the accordion.  It'd be great to persist it as expanded

nmalfroy (changes requested):
This is awesome!  Quick usability thing that might be good to fix that I think is a quick fix.  I'd love to get that in as part of this if that's possible

snf2ye (approved):
Looks awesome!

src/components/dataset/query/sidebar/panels/InfoView.jsx
tlangs: The accordion now stays expanded when a user clicks on the help link!

tlangs: I've deployed it on my personal instance and you can check it out: https://jade-tl.datarepo-dev.broadinstitute.org.

Side not, I just used my feature to add Jade Dev Stewards to my dataset. That's so cool!

src/components/dataset/query/QueryView.jsx
```diff
@@ -78,6 +80,10 @@ export class QueryView extends React.PureComponent {
     if (datasetPolicies == null || dataset.id !== datasetId) {
       dispatch(getDatasetPolicy(datasetId));
     }
+
+    if (userRole == null || dataset.id !== datasetId) {
```
nmalfroy, line 84: I think this if statement and the one above it are race conditions (you'll NPE if dataset is null...eg hasn't loaded yet.  You can do this to protect

```suggestion
    if (userRole == null || dataset?.id !== datasetId) {
```

src/components/dataset/query/sidebar/panels/InfoView.jsx
nmalfroy: Hmm...I'm not seeing that works on the deployed UI but the code looks like it should work so might be a deployment thing

nmalfroy (approved):
Looks great!  Quick thing about maybe having a defense against a race condition.  If you wouldn't mind adding that, I'm cool with merging

src/components/dataset/query/QueryView.jsx
tlangs: We specifically set the `dataset` object in state to default to an empty object. See https://github.com/DataBiosphere/jade-data-repo-ui/pull/1065/files#diff-514b4d730e0e38947ad19c45203aace96bc11ac05259adf1a519c832f3137e85R8

nmalfroy: oooooh ok.  Then not sure why we're checking for null on line 76

PR #1069, opened 2022-02-04
DR-1604 Script to reconstitute UI integration data

tlangs:
Here's a little python script to create the datasets necessary for UI integration tests to run. I've used it to populate integration-4 with the proper data, and tests can now run against it!


tools/ui_integration/setup_ui_integration.py
```diff
@@ -0,0 +1,160 @@
...
+        "snapshot_creators": ["JadeStewards-dev@dev.test.firecloud.org"]
+    }
+]
+
+
+class Clients:
+    def __init__(self, host):
+        config = Configuration()
+        config.host = host
+        token_output = subprocess.run(['gcloud', 'auth', 'print-access-token'], capture_output=True)
```
nmalfroy, line 41: Is there any information about what user (if any) is expected to set up the data?

tools/ui_integration/setup_ui_integration.py
```diff
@@ -0,0 +1,160 @@
...
+        return SnapshotsApi(api_client=self.api_client)
+
+    def datasets_api(self):
+        return DatasetsApi(api_client=self.api_client)
+
+    def profiles_api(self):
+        return ProfilesApi(api_client=self.api_client)
+
+    def jobs_api(self):
+        return JobsApi(api_client=self.api_client)
```
nmalfroy: Is it overkill to recreate a client every call?  For instance, this will create a new client instance every time that you poll for the job status

nmalfroy (approved):
Thanks for automating.  Couple comments but looks good!

tools/ui_integration/setup_ui_integration.py
tlangs: Anyone with access can run it.

tools/ui_integration/setup_ui_integration.py
tlangs: I went for speed of development over efficiency. I can fix that now.

tlangs:
Check it out. The tests ran on `integration-4`! 
image

tools/ui_integration/setup_ui_integration.py
nmalfroy: Oh, I know anyone can run it but then you have to make sure that the test user has access too

tlangs: The test user is a part of the dev jade stewards, which is given the appropriate access

nmalfroy: derp, you're totally right

snf2ye (approved):

PR #1072, opened 2022-02-11
DR-2421 [PoC] Convert the UI to TypeScript

myessail:


package.json
```diff
@@ -23,13 +23,23 @@
   ],
   "dependencies": {
     "@babel/polyfill": "^7.12.1",
+    "@emotion/styled": "^11.6.0",
     "@material-ui/core": "4.11.3",
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@material-ui/styles": "^4.11.3",
+    "@mui/material": "^5.4.1",
+    "@mui/styles": "^5.4.1",
```
snf2ye, line 32: Why did these three libraries need to be added? Is there a difference between "material-ui/styles" and "mui/styles"?

src/components/table/LightTable.jsx
```diff
@@ -122,7 +125,11 @@ function LightTable({
                     })}
                   >
                     {columns.map((col) => (
-                      <TableCell key={col.property} style={{ wordBreak: 'break-word' }}>
+                      <TableCell
+                        className={classes.cell}
```
snf2ye, line 129: What motivated this change?

src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+// / <reference types="react-scripts" />
```
snf2ye: What is this for?

src/actions/app.js
```diff
@@ -6,7 +6,7 @@
 
 import { createActions } from 'redux-actions';
 
-import { ActionTypes } from 'constants/index';
+import { ActionTypes } from '../constants/index';
```
snf2ye: It seems like there are a lot of changes to the import reference paths, but I don't think I see any corresponding file location changes. Why did these paths have to change?

snf2ye (commented):
A few questions - just trying to make sure I understand all of the changes going in. 

Also - Is there any change to how we build and run things with these changes?

src/components/table/SearchTable.tsx
```diff
@@ -0,0 +1,75 @@
...
+      height: '60%',
+      transform: 'translateY(30%)',
+      cursor: 'pointer',
+      ...theme.mixins.jadeLink,
+    },
+  });
+
+interface IProps extends WithStyles<typeof styles> {
+  searchString: string;
+  onSearchStringChange: any;
```
nmalfroy, line 49: Let's avoid using `any` from the get-go if we can.  Shouldn't these at least be functions?  Also, it's weird that one uses `on...` and the other doesn't  Maybe rename `clearSearchString` to `onClearSearchString`

nmalfroy (commented):
Looking good.  I have the same questions as shelby

src/components/table/LightTable.jsx
myessail: This was the styling bug that I was trying to figure out the problem with, I wasn't able to figure out why the color had changed so I just added the styling manually to make it the same as before

package.json
myessail: the "mui" libraries are the v5 versions, which are more compatible with typescript. there were some pretty big changes between v4 and v5 so i figured for now we could keep both to make sure everything else in the ui stays the same

src/react-app-env.d.ts
myessail: i believe this was added by react-scripts, but we don't actually have any types to add here at the moment

src/actions/app.js
myessail: ugh...they don't need to change now, typescript was complaining about it so I originally went around changing a bunch of them before I realized the config change I needed to make -- both forms should work now but I can try and go back/undo some of these to clean up the diff

src/react-app-env.d.ts
snf2ye: Can we just delete it?

src/actions/app.js
snf2ye: Ah, gotcha. Yeah, unfortunately, I think it would be better to undo.

snf2ye (approved):

nmalfroy (approved):
nicely done!

PR #1075, opened 2022-02-18
DR-2433 Add checkbox to UI for enabling pre-resolve DRS urls on export

tlangs:
Checkbox has been added to allow users to pre-resolve DRS URIs to gs paths on snapshot export. The UI now sends a `exportGsPaths` URL parameter in its snapshot export request. This can be changed to be part of a post body if needed, but I think a URL param makes sense.

### DIFF
--- a/src/components/DetailViewHeader.jsx
+++ b/src/components/DetailViewHeader.jsx
@@ -5,7 +5,18 @@ import moment from 'moment';
 import { withStyles } from '@material-ui/core/styles';
 import { exportSnapshot, resetSnapshotExport } from 'actions/index';
 import { connect } from 'react-redux';
-import { Card, Grid, Typography, Button, CircularProgress, Divider } from '@material-ui/core';
+import {
+  Card,
+  Grid,
+  Typography,
+  Button,
+  CircularProgress,
+  Divider,
+  Checkbox,
+  FormGroup,
+  FormControlLabel,
+  FormHelperText,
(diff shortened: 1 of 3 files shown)


src/components/DetailViewHeader.jsx
```diff
@@ -174,8 +206,19 @@ export class DetailViewHeader extends React.PureComponent {
             )}
             <Divider className={classes.separator} />
             <Typography variant="h6" className={classes.section}>
-              Export a copy of the snapshot metadata to an exisiting or new Terra workspace
+              Export a copy of the snapshot metadata to an existing or new Terra workspace
```
snf2ye, line 209: OH no, I think this is my misspelled word - Thanks for fixing

snf2ye (approved):
LGTM

tcjiang9 (approved):

src/components/DetailViewHeader.jsx
```diff
@@ -174,8 +206,19 @@ export class DetailViewHeader extends React.PureComponent {
...
             </Typography>
+            <FormGroup>
+              <FormControlLabel
+                control={gsPathsCheckbox}
+                label="Convert DRS URLs to Google gs-paths"
```
tcjiang9: nit: I think we should call them Google Cloud Storage paths

nmalfroy (approved):

PR #1076, opened 2022-02-22
DR-2419 Specify dataset columns in order rather than selecting all

tlangs:
The UI portion of https://broadworkbench.atlassian.net/browse/DR-2419 making sure that the UI displays dataset columns in the order they were returned by the API.


snf2ye (approved):

fboulnois (approved):
lgtm!

PR #1078, opened 2022-02-25
DR-2363 New dataset overview page

tlangs:
Its here! A new dataset details page! This will now be the default page users see when they click on a dataset from the home view. There's a link to the dataset contents, and a new button that takes them back to the details page. 

I did my best to match the styles in the mockup, but its not an exact match, and there's a long tail of work to make the small changes to make it look exactly the same. The functionality, however, is there! 

You can check it out on my personal instance: https://jade-tl.datarepo-dev.broadinstitute.org

image
image


tlangs:
I still need to fix the cypress tests, which are very much going to break as a result of these changes.

deploy_local_changes.sh
```diff
@@ -0,0 +1,51 @@
...
+trap swap_back EXIT
+
+function stderr {
+    >&2 echo "$@"
+}
+
+function help() {
+  stderr "This little script makes it easy to switch the Dockerfile and Dockerfile.direct files "
+  stderr "and deploy local UI changes to a personal deployment. "
+  stderr "The script requires a 'skaffold.yml' file to be present and set up to point to a personal deployment. "
```
snf2ye, line 12: nice!

snf2ye (approved):
All the code looks reasonable to me and the UI looks great!

Thanks for providing example datasets on your deployment. You probably already thought of these, but a couple of edge cases to test with a bit more data:
- What happens when the schema is long - does the scrolling work as expected?
- Does the new snapshot list page work as expected when there are bunch of snapshots? (paging, sorting, etc.)

tlangs:
If the schema is long enough, then it will simply extend the bottom of the page. The panel itself won't scroll, but the page will extend to fit the contents.

As for snapshots, I've wired in the pagination code. Getting enough snapshots to get the pagination to show up, however, would take a lot of snapshots. It should work the same as every other `JadeTable`, and I was able to re-use the existing `getSnapshots` saga code, so I don't see why it wouldn't work. 

If it does break, I think we could do a fast follow on. The changes in here are already substantial, and I'm anxious to get them in before develop diverges too much.

snf2ye:
@tlangs  - Sounds great - I'm good with proceeding!

src/components/ManageUsersModal.jsx
```diff
@@ -42,6 +43,19 @@ const styles = (theme) => ({
   openButton: {
     width: '100%',
   },
+  iconButton: {
+    border: `1px solid ${theme.palette.primary.main}`,
+    borderRadius: '3px',
+    padding: '1px',
+    backgroundColor: theme.palette.primary.main,
```
nmalfroy, line 50: Is this not standard button styling?

nmalfroy: I think if you give it the primary variant, the normal button styling should work

src/components/dataset/details/DatasetDetailPanel.jsx
```diff
@@ -0,0 +1,143 @@
...
+import { renderCloudPlatforms, renderStorageResources } from '../../../libs/render-utils';
+import DatasetAccess from '../DatasetAccess';
+import DatasetSnapshotsTable from '../../table/DatasetSnapshotsTable';
+
+const styles = (theme) => ({
+  root: {
+    flexGrow: 1,
+  },
+  tabsRoot: {
+    color: '#333F52',
```
nmalfroy: This color should likely be coming coming from the theme

src/components/ManageUsersModal.jsx
nmalfroy: sorry, `contained` variant

src/components/dataset/query/sidebar/panels/InfoViewDatasetAccess.jsx
```diff
@@ -0,0 +1,89 @@
...
+);
+
+const helpContent = (
+  <Fragment>
+    <Typography variant="h6">Stewards</Typography>
+    <Typography>
+      Creating a dataset makes a user the Steward (or owner) of it. The steward of a dataset can:
+    </Typography>
+    <ul>
+      <li>Define the schema for a dataset</li>
```
nmalfroy: Hmm...the implies that the dataset's schema can be updated...

src/components/table/DatasetSnapshotsTable.jsx
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  handleFilterSnapshots = (limit, offset, sort, sortDirection) => {
+    const { dispatch, dataset } = this.props;
+    dispatch(getDatasetSnapshots(limit, offset, sort, sortDirection, [dataset.id]));
+  };
+
+  render() {
+    const { classes, snapshotCount, snapshots, summary } = this.props;
+    // TODO add back modified_date column
```
nmalfroy: Why this TODO?

nmalfroy (approved):
Looks so cool!  A couple of small comments that would be good to fix before merging.  Also, it's probably worth sitting down with Lou to make sure that this lines up with what he's thinking.  I'm noticing some general UX strangeness (like the fact that there's no side bar on details is odd to me

src/components/dataset/query/sidebar/panels/InfoViewDatasetAccess.jsx
tlangs: This is pulled directly from our documentation. I'll remove it now, but we should update this when we do add schema migration

src/components/ManageUsersModal.jsx
tlangs: No, this is a lot of custom styling to force the icon button to be the same size to the left of the header.

tlangs:
@nmalfroy I removed a _bunch_ of unused code. Let me know if you think it should still be there.

nmalfroy:
That's cool man!  As long as we click test I'm aways down with red lines in PRs :)

PR #1081, opened 2022-02-28
DR-2456 - Display Readers list for readers 

snf2ye:


tlangs (approved):
Gah this looks like my mistake when I refactored. Thanks for fixing!

nmalfroy (approved):

PR #1085, opened 2022-03-10
DR-2344 Represent null and empty strings differently

tlangs:
Representing empty strings and null values differently in data tables makes it easier for users to track down what's in their datasets. 

The following screenshot is the result of ingesting:
```json
{"sample_name":"emptyDataType","data_type":""}
{"sample_name":"nullDataType","data_type":null}
{"sample_name":"missingDataType"}
```

image


snf2ye (approved):

src/components/dataset/query/DatasetQueryView.jsx
```diff
@@ -213,7 +213,7 @@ export class DatasetQueryView extends React.PureComponent {
               <QueryViewDropdown options={names} onSelectedItem={this.handleChange} />
             </Grid>
             <Grid item xs={3}>
-              <Link to="details">
+              <Link to="overview">
```
nmalfroy: huh...I thought we fixed this already

src/components/table/JadeTable.jsx
```diff
@@ -126,25 +131,34 @@ export class JadeTable extends React.PureComponent {
   };
 
   handleArrayValues = (value, column) => {
+    const returnValue = [];
+    if (column.mode === COLUMN_MODES.REPEATED) {
+      returnValue.push(<span key="start">[</span>);
+    }
+    returnValue.push(
+      ...value.flatMap((v, i) => [
+        v,
```
nmalfroy: what if the value in the array is null?  I think you want
```suggestion
        _.isNull(v) ? this.handleNullValue(v) : v,
```

nmalfroy (changes requested):
nice!  I do wonder about nulls in arrays?

src/components/dataset/query/DatasetQueryView.jsx
tlangs: We did! Just updated with contents from develop

nmalfroy (approved):
thanks for making the changes!

PR #1087, opened 2022-03-10
DR-2421 Fix styling from MUI changes

myessail:


src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
nmalfroy: why do we need this?

nmalfroy (approved):
thanks!

src/react-app-env.d.ts
myessail: I think this gets generated every time I run `npm start`... not sure what it would be used for but I can delete/add to gitignore

snf2ye (approved):

PR #1090, opened 2022-03-15
DR-2436: Snapshot Query

snf2ye:
Changes deployed here:
https://jade-sh.datarepo-dev.broadinstitute.org/


src/actions/app.js
```diff
@@ -168,19 +168,30 @@ export const { hideAlert } = createActions({
...
-  [ActionTypes.RUN_QUERY]: (projectId, query, maxResults) => ({
+  [ActionTypes.RUN_QUERY]: (projectId, query) => ({
     projectId,
     query,
-    maxResults,
+  }),
+});
+
+export const { previewData } = createActions({
+  [ActionTypes.PREVIEW_DATA_SUCCESS]: (queryResults, columns) => ({
```
nmalfroy, line 185: minor: could we swap the order?  I like `FOO_SUCCESS` being after `FOO`

src/components/table/JadeTable.jsx
```diff
@@ -150,82 +106,91 @@ export class JadeTable extends React.PureComponent {
...
+                  );
+                })}
+              </TableBody>
+            )}
+          </Table>
+        )}
+        {polling && (
+          <LoadingSpinner
+            delay={delay}
+            delayMessage="For large datasets, it can take a few minutes to fetch results from BigQuery. Thank you for your patience."
```
nmalfroy: While we're in here, let's maybe take out `from BigQuery` since we're gonna want this for parquet too

src/reducers/query.js
```diff
@@ -26,13 +32,55 @@ export default {
...
+        const rows = queryResults.map((row) => {
+          const res = {};
+          columns.forEach((col) => {
+            res[col.id] = row[col.id];
+          });
+          res.datarepo_id = row.datarepo_row_id;
+          return res;
+        });
+        const queryParams = {
+          // pageToken: queryResults.pageToken,
```
nmalfroy: commented code

src/actions/app.js
nmalfroy: should we also have a FAILURE handler?

nmalfroy (commented):
Did an initial pass. looking good!  I need to click test too

src/actions/app.js
snf2ye: Good point - added. I added a "Failure" component as a sorta placeholder -- probably should come up with a design for handling the case when the request fails. Right now it's just an error icon + error message.

src/actions/app.js
```diff
@@ -168,19 +168,31 @@ export const { hideAlert } = createActions({
...
-  [ActionTypes.RUN_QUERY]: (projectId, query, maxResults) => ({
+  [ActionTypes.RUN_QUERY]: (projectId, query) => ({
     projectId,
     query,
-    maxResults,
   }),
 });
 
+export const { previewData } = createActions({
+  [ActionTypes.PREVIEW_DATA]: (snapshotId, table, columns, totalRowCount) => ({
```
tlangs: Will we be using the preview API for datasets as well as snapshots? If so, we might want to not call the id `snapshotId` here.

src/actions/app.js
```diff
@@ -200,6 +212,18 @@ export const { applySort } = createActions({
   }),
 });
 
+export const { resetQuery } = createActions({
```
tlangs, line 216: If these new actions are a part of the new preview API, shouldn't they be declared under the `previewData` constant?

src/components/common/query/QueryView.jsx
```diff
@@ -0,0 +1,129 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import { withStyles } from '@material-ui/core/styles';
+import Grid from '@material-ui/core/Grid';
+
+import { Button, Typography } from '@material-ui/core';
+
+import { Link } from 'react-router-dom';
+import SidebarDrawer from 'components/dataset/data/sidebar/SidebarDrawer';
```
tlangs: the `SidebarDrawer` should be moved out of the `dataset` folder if its used for both datasets and snapshots.

src/components/common/query/QueryView.jsx
```diff
@@ -0,0 +1,129 @@
...
+  sidebarWidth,
+  tableNames,
+}) {
+  return (
+    <Fragment>
+      {resourceLoaded && (
+        <div className={classes.pageRoot}>
+          <AppBreadcrumbs
+            context={{
+              type:
```
tlangs: I would love to make typescript types for datasets and snapshots. Then, we could just pass the dataset/snapshot to `AppBreadcrumbs` and it could figure out what to do based on the type instead of needing to pass the resource type in as a parameter.

src/components/common/query/QueryView.jsx
```diff
@@ -0,0 +1,129 @@
+import React, { Fragment } from 'react';
```
tlangs: Since this is now referred to as the "Data" page for datasets/snapshots, can we rename files to match?

src/components/dataset/data/sidebar/QueryViewSidebarItem.jsx
```diff
@@ -42,19 +42,19 @@
~    const { filterMap } = this.state;
~    // enable the button when there are unsaved changes
~    if (!_.isEqual(prevState.filterMap, filterMap)) {
~      this.setState({ disableButton: this.invalidChange(filterMap) });
~    }
```
tlangs: Nice catch!

src/components/dataset/data/DatasetDataView.jsx
```diff
@@ -0,0 +1,220 @@
...
+import {
+  DATAREPO_ROW_ID_COLUMN_NAME,
+  DATASET_INCLUDE_OPTIONS,
+  GOOGLE_CLOUD_RESOURCE,
+  RESOURCE_TYPE,
+} from '../../../constants';
+
+const QUERY_LIMIT = 1000;
+
+function DatasetQueryView({
```
tlangs: Did we decide to make all new components in Typescript? I can't remember. It would be cool to see Typescript proliferate in our codebase, though!

src/components/table/DatasetSnapshotsTable.jsx
```diff
@@ -39,20 +39,23 @@ class DatasetSnapshotsTable extends React.PureComponent {
     const columns = [
       {
         label: 'Snapshot Name',
-        property: 'name',
+        name: 'name',
+        allowSort: true,
         render: (row) => (
-          <Link to={`/snapshots/details/${row.id}`} className={classes.jadeLink}>
+          <Link to={`/snapshots/${row.id}/overview`} className={classes.jadeLink}>
```
tlangs: There's no more `overview` route. The snapshot overview pages now live at `/snapshot/{ID}`

src/components/table/JadeTable.jsx
```diff
@@ -45,201 +36,172 @@ const styles = (theme) => ({
...
+            <JadeTableHead columns={columns} createSortHandler={createSortHandler} />
+            {!polling && (
+              <TableBody data-cy="tableBody">
+                {rows.map((row, i) => {
+                  const drId = row.datarepo_row_id;
+
+                  return (
+                    <TableRow hover tabIndex={-1} key={drId}>
+                      {columns.map((column) => {
+                        const value = handleValues(row[column.name], column);
```
tlangs, line 141: I like this change! Much clearer!

src/components/DetailViewHeader.jsx
```diff
@@ -57,6 +58,10 @@ const styles = (theme) => ({
     marginTop: '20px',
     marginBottom: '10px',
   },
+  viewSnapshotButton: {
+    width: '100%',
+    marginBottom: '12px',
+  },
 });
 
 export class DetailViewHeader extends React.PureComponent {
```
tlangs: Should this now be `OverviewHeader`?

src/reducers/query.js
```diff
@@ -1,20 +1,28 @@
 import { handleActions } from 'redux-actions';
 import immutable from 'immutability-helper';
 import { LOCATION_CHANGE } from 'connected-react-router';
-import { ActionTypes } from 'constants/index';
+import { ActionTypes, TABLE_DEFAULT_ROWS_PER_PAGE } from 'constants/index';
 import BigQuery from 'modules/bigquery';
 
 export const queryState = {
```
tlangs, line 7: I'm wondering if its worth renaming this to `previewState`, since it's using the preview API

tlangs (changes requested):
This was a hefty effort to take on. Thanks so much for chewing through it!

src/components/dataset/data/DatasetDataView.jsx
snf2ye: Definitely - It was just more than I wanted to take on with this PR. I did attempt to rework any models to consistent var names/etc, which should help with that transition.

src/components/table/DatasetSnapshotsTable.jsx
snf2ye: Oh thanks, I thought I had caught all of the route changes.

src/components/DetailViewHeader.jsx
snf2ye: Yeah - I've done another pass now and renamed a bunch more files to corresponding data/overview names.

src/reducers/query.js
snf2ye: There's so much tied into this "query" name - I think we can push off that name migration to when we ditch the BQ query aspect altogether. At the very least, I'd want to punt to another ticket/PR.

src/components/common/query/QueryView.jsx
snf2ye: Hopefully with my last pass, I've changed more of the components to the new names.

src/components/common/query/QueryView.jsx
snf2ye: Agreed - I think this will be covered in https://broadworkbench.atlassian.net/browse/DR-2483 - I'll add this to A.C.

src/actions/app.js
snf2ye: Any of the "query" actions apply to all of the tables, so it wouldn't be appropriate to move this under previewData.

src/components/common/data/DataView.jsx
```diff
@@ -0,0 +1,129 @@
...
+  queryParams: PropTypes.object,
+  resourceId: PropTypes.string,
+  resourceLoaded: PropTypes.bool,
+  resourceName: PropTypes.string,
+  resourceType: PropTypes.string,
+  selected: PropTypes.string,
+  selectedTable: PropTypes.object,
+  sidebarWidth: PropTypes.number,
+  tableNames: PropTypes.array,
+  pageBQQuery: PropTypes.func,
```
nmalfroy: Mind just making this alphabetical (or should we get rid of this linting rule?)

src/components/dataset/data/DatasetDataView.jsx
```diff
@@ -0,0 +1,220 @@
...
+  const [selected, setSelected] = useState('');
+  const [selectedTable, setSelectedTable] = useState(null);
+  const [sidebarWidth, setSidebarWidth] = useState(0);
+  const [canLink, setCanLink] = useState(false);
+  const [datasetLoaded, setDatasetLoaded] = useState(false);
+  const [tableNames, setTableNames] = useState([]);
+  const [panels, setPanels] = useState([]);
+
+  useEffect(() => {
+    const datasetId = match.params.uuid;
```
nmalfroy, line 50: We may want to make the datasetId be a property of the component and do the mapping in a `connect` wrapper.

src/components/snapshot/data/SnapshotDataView.jsx
```diff
@@ -0,0 +1,137 @@
...
+function SnapshotDataView({ dispatch, match, page, queryParams, rowsPerPage, snapshot }) {
+  const [selected, setSelected] = useState('');
+  const [selectedTable, setSelectedTable] = useState(null);
+  const [sidebarWidth, setSidebarWidth] = useState(0);
+  const [snapshotLoaded, setSnapshotLoaded] = useState(false);
+  const [tableNames, setTableNames] = useState([]);
+  const [panels, setPanels] = useState([]);
+
+  useEffect(() => {
+    const snapshotId = match.params.uuid;
```
nmalfroy, line 19: Same comment about grabbing this from state in a connect method

nmalfroy (approved):
Man, this PR must have been a slog, thanks for working through it!  Is your deployed instance up to date with this code?  I'll do a list quick click test.  Let's merge this as soon as this week's deploy is done

tlangs (approved):
Once Muscle's comments are resolved, merge this puppy!

src/components/dataset/data/DatasetDataView.jsx
snf2ye: I'm not sure how to do that. We are already pulling the match value via the connect wrapper.

There is, however,  a hooks implementation we could use instead: https://v5.reactrouter.com/web/api/Hooks/useparams

snf2ye:
>\s your deployed instance up to date with this code? I'll do a list quick click test. Let's merge this as soon as this week's deploy is done

@nmalfroy  - Yes, latest is deployed

src/components/dataset/data/DatasetDataView.jsx
snf2ye: It also seems like keeping this particular info out of the store makes sense for readability - It's clear that we're triggering updates starting from this component when the url uuid updates.

nmalfroy: I guess I was thinking of doing this mapping explicitly in the mapStateToProps method?  So at the bottom of the file you might have:

```function mapStateToProps(state) {
  return {
    datasets: state.datasets.datasets,
    datasetsCount: state.datasets.datasetsCount,
    filteredDatasetsCount: state.datasets.filteredDatasetsCount,
    features: state.user.features,
    datasetId: state.params.uuid,
  };
}``

And then datasetId is a first class prop of the component.

Basically, consolidate the mapping there...does that make sense?

nmalfroy:
> > Is your deployed instance up to date with this code? I'll do a list quick click test. Let's merge this as soon as this week's deploy is done
> 
> @nmalfroy - Yes, latest is deployed

Cool, thanks!  I've been click testing and it looks great!

src/components/dataset/data/DatasetDataView.jsx
snf2ye: Yes, that makes sense. 
I'm not sure how to populate the state.params value, though. As far as I can tell, the router state in the store doesn't currently include that information.

nmalfroy: talked offline and it's totally fine to leave

PR #1092, opened 2022-03-18
DR-2468 Breadcrumbs for datasets and snapshots

tlangs:
Here's the initial implementation of breadcrumbs for the UI. The `AppBreadcrumbs` Component contains almost all the logic, with only child pages of datasets/snapshots needing to provide any breadcrumb information. The breadcrumbs are rendered based on the `context` passed into the component.

I took this as an opportunity to re-work some of the styling in the dataset and snapshot pages. The `pageRoot` is now re-used for all three views, and the styling is stored in `theme.ts` as a mixin. This makes the app behave consistently across pages! 

I've also re-routed the dataset details page to be at `/datasets/{id}/data` and renamed it to the `DatasetDataView`. I think this is much more descriptive of the page contents. Additionally, the Overview is now at `/datasets/{id}`, because the data page feels like a child of the overview page. 

Give it a test drive at https://jade-tl.datarepo-dev.broadinstitute.org!

image
image
image


src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { Breadcrumbs, Link } from '@material-ui/core';
+import { Link as RouterLink } from 'react-router-dom';
+
+function capitalize(str: string): string {
+  return str.charAt(0).toUpperCase() + str.slice(1);
+}
+
+const styles = (theme: any) => ({
+  breadcrumb: {
+    fontStyle: 'italic',
```
nmalfroy: why italics?  I don't think the designs have the italics (also, I like it better without them :) )

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
```diff
@@ -0,0 +1,88 @@
...
+  },
+  terminalBreadcrumb: {
+    fontStyle: 'italic',
+    color: theme.palette.primary.dark,
+    cursor: 'default',
+    '&:hover': {
+      textDecoration: 'none',
+    },
```
nmalfroy: You replace all of this with (also for consistency):
```suggestion
    ...theme.mixins.jadeLink,
```

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const TerminalBreadcrumb = withStyles(styles)((props: any) => {
+  const { classes, ...other } = props;
+  return (
+    <Link
+      component={RouterLink}
+      color="primary"
+      className={classes.terminalBreadcrumb}
+      {...other}
+      onClick={(e) => e.preventDefault()}
```
nmalfroy: I don't think you need this do you?

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
```diff
@@ -0,0 +1,88 @@
...
+      textDecoration: 'none',
+    },
+  },
+});
+
+export type Context = { type: string; id: string; name: string };
+export type BreadcrumbElement = { text: string; to: string };
+
+type AppBreadcrumbsProps = {
+  links: Array<BreadcrumbElement>;
```
nmalfroy: I think it's a more common pattern to make this `children` and pass in the react `BreadcrumbsLink` or `TerminalBreadcrumb` components

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
```diff
@@ -0,0 +1,88 @@
...
+      className={classes.terminalBreadcrumb}
+      {...other}
+      onClick={(e) => e.preventDefault()}
+    />
+  );
+});
+
+const Breadcrumb = withStyles(styles)((props: any) => {
+  if (props.disabled) {
+    return <TerminalBreadcrumb {...props} />;
```
nmalfroy: This doesn't feel like it should be its own component.  It's really just `BreadcrumbsLink` (maybe use BreadcrumbLink - singular) with no link property passed in

src/components/DetailViewHeader.jsx
```diff
@@ -142,7 +143,7 @@ export class DetailViewHeader extends React.PureComponent {
             <CircularProgress />
           ) : (
             <Fragment>
-              <Typography noWrap className={classes.title}>
+              <Typography noWrap variant="h3" className={classes.pageTitle}>
```
nmalfroy, line 146: 👍

src/components/SnapshotDetailView.jsx
```diff
@@ -141,31 +141,40 @@ export class SnapshotDetailView extends React.PureComponent {
...
-          />
-          {snapshot && snapshot.source && (
-            <DatasetTable
-              datasets={filteredDatasets || datasets}
-              datasetsCount={snapshot.source.length}
-              features={user.features}
-              handleFilterDatasets={this.handleFilterDatasets}
-              searchString=""
+      <div className={classes.pageRoot}>
+        <AppBreadcrumbs
```
nmalfroy, line 145: Again, I think this is simplified if you just pass in the breadcrumb elements

nmalfroy (changes requested):
It looks super cool in action!  I think think the way that the breadcrumb component is setup is a little bit of a react anti-pattern.  It may be a little more code but I do think passing in the child elements directly is going to be more maintainable on a page by page level

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
tlangs: You do. If you don't include it, the React Router will still register the location change and mess with history.

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
tlangs: This would actually achieve the opposite of what we want. The terminal breadcrumb should be styled differently than the clickable breadcrumbs before it.

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
tlangs: Whoops. 1998 Trevyn made an appearance and italicized the links!

src/components/SnapshotDetailView.jsx
tlangs: I don't think it would be simpler. By passing in elements directly, we'd be letting each page be able to potentially style the breadcrumbs itself. Additionally, the page doesn't necessarily know where it is in the hierarchy, so containing the logic in `AppBreadcrumbs` means we don't have to continually calculate the url for hrefs. Also, if we refactor again, we have to go through and make sure that every invocation of `AppBreadcrumbs` is updated to reflect a new hierarchy. By containing all the logic in `AppBreadcumbs`, we avoid duplication and make refactoring easier.

.eslintrc
```diff
@@ -197,6 +197,7 @@
     ],
     "jsx-quotes": "warn",
     "react/forbid-prop-types": "off",
+    "react/jsx-filename-extension": [2, { "extensions": [".js", ".jsx", ".ts", ".tsx"] }],
```
snf2ye, line 200: Did this remove the `JSX not allowed in files with extension '.tsx'` errors for you on npm start? I tried adding this, but it didn't seem to help.

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
```diff
@@ -0,0 +1,73 @@
...
+        component={RouterLink}
+        color="primary"
+        className={classes.terminalBreadcrumb}
+        onClick={(e) => e.preventDefault()}
+        {...other}
+      />
+    );
+  }
+  return <Link component={RouterLink} color="primary" {...other} />;
+});
```
snf2ye: I don't know that it makes sense to pull this into a separate constant within the same file. If it's not shared with anything, I think it should live within the AppBreadcrumbs component. If it needs to be a separate, reusable component, then it should live in its own file.

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
```diff
@@ -0,0 +1,73 @@
+import React from 'react';
+import { withStyles } from '@material-ui/core/styles';
+import { Breadcrumbs, Link } from '@material-ui/core';
+import { Link as RouterLink } from 'react-router-dom';
+
+function capitalize(str: string): string {
+  return str.charAt(0).toUpperCase() + str.slice(1);
+}
```
snf2ye: I think this should be a const function:
```suggestion
const capitalize(str: string) = (string) => ({
  return str.charAt(0).toUpperCase() + str.slice(1);
});
```

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
```diff
@@ -0,0 +1,73 @@
...
+        onClick={(e) => e.preventDefault()}
+        {...other}
+      />
+    );
+  }
+  return <Link component={RouterLink} color="primary" {...other} />;
+});
+
+const AppBreadcrumbs: React.FC<AppBreadcrumbsProps> = (props) => {
+  const { context, childBreadcrumbs } = props;
```
snf2ye: UPDATE: I'm wrong. But, I do think we could get rid of React.FC. New comment forthcoming.
It's been forever since I've looked at typescript -- This could still be a functional component, right?
```suggestion
function AppBreadcrumbs: React.FC<AppBreadcrumbsProps> = ({
  context, childBreadcrumbs
  }) => { 
```

src/components/dataset/query/DatasetDataView.jsx
```diff
@@ -34,16 +37,12 @@ const styles = (theme) => ({
...
-    paddingTop: theme.spacing(1),
-    paddingBottom: theme.spacing(1),
-  },
 });
 
 const PAGE_SIZE = 100;
 const QUERY_LIMIT = 1000;
 
-export class DatasetQueryView extends React.PureComponent {
+export class DatasetDataView extends React.PureComponent {
```
snf2ye, line 45: Ok.. this is going to be less fun to merge with my changes 😁

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
```diff
@@ -0,0 +1,73 @@
...
+    color: theme.palette.primary.dark,
+    cursor: 'default',
+    '&:hover': {
+      textDecoration: 'none',
+    },
+  },
+});
+
+export type Context = { type: string; id: string; name: string };
+export type Breadcrumb = { text: string; to: string };
```
snf2ye: Do these need to be exported? If they're shared across components, we might want to consider pulling these into their own files.

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
snf2ye:
```suggestion
const AppBreadcrumbs = ({context, childBreadcrumbs}: AppBreadcrumbsProps) => {
```
Ok, does this work? Following the example[ here](https://github.com/typescript-cheatsheets/react#function-components) - Seems that React.FC is discouraged.

.eslintrc
tlangs: It did!

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
tlangs: Here's where my understanding of scoping in Javascript reaches its limits. My assumption is that, if the function is in a file and not `export`ed, then its scope is confined to the file. For me, it makes sense for this component to be a separate function in the same file for organization and readability reasons.

snf2ye: I'm okay with keeping it as it, but a few thoughts:
- My understanding of React conventions is what I stated before: This - which is definitely more a small component than just a function - should either live in its own file or within the default component of the page. I think there is some merit to sticking to conventions, but they're not always the best
- I do not find this approach more readable - the additional decoration of the function withStyles plus having to track a second list of props... which has the type of any! (the horror!) makes this seem more complex than it needs to be.

snf2ye (approved):
One last comment to consider, but I think it looks great! Thanks for making the typescript changes!

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
tlangs: Good points. Changed so that `BreadcrumbLink` is in its own file!

nmalfroy (approved):
looks great!

PR #1094, opened 2022-03-22
DR-2475 Limit number of repeated elements rendered in tables

tlangs:
If a dataset has an array column with a large number of elements, it breaks the UI because all of those elements are rendered, making each row in the table huuuuuuuge.

Now, only the first 5 elements of an array column are rendered, and a `See full list` link is added as the 6th element. Clicking this link pulls up a dialog box with the full contents of the column. If the full contents of the column are too large for the screen, the dialog box will scroll. 

Check it out at https://jade-tl.datarepo-dev.broadinstitute.org/datasets/2a8af66c-bb35-42ed-918d-aa4b6e64a5e6/query

image
image
image


nmalfroy (approved):
looks good!  let's maybe just wait until @snf2ye 's PR is merged first before we merge this one

snf2ye (approved):

PR #1095, opened 2022-03-23
[DC-220] Update setup script to create snapshots

samanehsan:
The Data Catalog needs to create test datasets and snapshots in the Data Repo with different policy combinations. Since the `setup_ui_integration.py` script is so convenient, I made a couple of changes so we could re-use it:
- The datasets to upload are defined in a separate json file now, which has:
  - A new "schema" field so that the same schema can be used to create multiple datasets in the same environment
  - A new "upload_prefix" field for the path to the metadata ingest file
  - An optional "snapshots" field to define snapshots to create from this dataset.

Example usage for the data catalog:

1. `gcloud auth login datacatalogadmin@test.firecloud.org`
2. `python setup_ui_integration.py --host 'https://jade.datarepo-dev.broadinstitute.org' --datasets './suites/catalog_admin_datasets.json'`

Notes:  
- User credentials are in vault - see https://broadworkbench.atlassian.net/browse/DC-218 for details.
- Both test users and the Data Repo service accounts for dev, alpha, staging and prod are storage object viewers on the `catalog-testdata-useastregion` bucket containing test dataset data in the dev gcloud project.
~- We also need to add the puppeteer service account to the same policy group as the user in each environment for testing the UI~


tlangs:
Glad you found it useful!

tools/ui_integration/README.md
```diff
@@ -0,0 +1,18 @@
+The `setup_ui_integration.py` script can be used to easily create datasets and snapshots in the Data Repo.
```
tlangs, line 1: Yay documentation!

tools/ui_integration/catalog_admin_datasets.json
```diff
@@ -0,0 +1,31 @@
+[
```
tlangs: Splitting these out into separate "suites" is a great idea!

tlangs (approved):

tools/ui_integration/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+###Usage:
```
pshapiro4broad: I think these need spaces to render properly
```suggestion
### Usage:
```

tools/ui_integration/setup_ui_integration.py
```diff
@@ -113,8 +87,9 @@ def create_dataset(clients, dataset_to_upload, profile_id):
     dataset_name = dataset_to_upload['name']
 
     dataset = None
-    with open(os.path.join("files", dataset_name, "dataset_schema.json")) as dataset_schema_json:
+    with open(os.path.join("files", dataset_to_upload['schema'], "dataset_schema.json")) as dataset_schema_json:
```
pshapiro4broad: consistent quotes?
```suggestion
    with open(os.path.join("files", dataset_to_upload["schema"], "dataset_schema.json")) as dataset_schema_json:
```

tools/ui_integration/setup_ui_integration.py
```diff
@@ -139,14 +152,20 @@ def main():
...
+        dataset_name = created_dataset['name']
+        output_ids = {dataset_name: {'dataset_id': created_dataset['id']}}
+        if dataset_to_upload.get('snapshots'):
+            snapshot_ids = create_snapshots(clients, dataset_to_upload['name'], dataset_to_upload['snapshots'],
+                                            profile_id)
+            output_ids[dataset_name]['snapshot_ids'] = snapshot_ids
+            outputs.append(output_ids)
+
+    output_filename = f"{args.datasets.split('.')[0]}_outputs.json"
+    with open(output_filename, 'w') as f:
```
pshapiro4broad, line 169: What is the output file used for? Maybe add it to the documentation as well.

pshapiro4broad (approved):
This looks good to me. Although, I am wondering if the catalog tests will need to create snapshots dynamically instead of using static snapshots. That's a bigger conversation though, and either way it's good to have the ability to create snapshots and policies here as well.

pshapiro4broad:
> That's a bigger conversation though, and either way it's good to have the ability to create snapshots and policies here as well.

We talked about this during a mobbing meeting yesterday (3/30) and decided that the end-to-end UI tests need to have a consistent set of snapshots in all test environments to test different operations. The service end-to-end tests, which need to modify catalog state, will most likely need to create snapshots dynamically to avoid test collisions.

tools/ui_integration/setup_ui_integration.py
samanehsan: The script used to print out all of the dataset ids, but I think it's easier to write the results to a file now that it includes snapshot ids as well.

PR #1098, opened 2022-03-28
DR-2437 Update 404 page

tlangs:
Updating the 404 page with a fun new graphic! This PR simply the old 404 page. There is no change in behavior when a dataset/snapshot is not found. Those still render out a blank page with the error toast. If we want to change that behavior, we should consider a follow-on ticket.

image


src/routes/NotFound.tsx
```diff
@@ -0,0 +1,71 @@
...
+        </Typography>
+        <Typography className={classes.h2} variant="h2">
+          Page Not Found
+        </Typography>
+        <Typography className={classes.text}>
+          We're sorry, the page you requested could not be found{' '}
+        </Typography>
+        <Button
+          className={classes.button}
+          component={RouterLink}
```
nmalfroy, line 55: I think you want to add `disableElevation`

nmalfroy (changes requested):
Nice!  Another tsx file! You mind just creating the followon ticket for 404-ing if the dataset/snapshot isn't found?

tlangs:
I've created https://broadworkbench.atlassian.net/browse/DR-2511 as a follow on.

src/routes/NotFound.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
```
snf2ye, line 1: Since this is a component, I think this file should live in components/common folder. But, probably still worth re-visiting organization as a team.

snf2ye (approved):

fboulnois (approved):
neat! 📸

nmalfroy (approved):

PR #1100, opened 2022-03-29
DR-2474 Loading indicators for Dataset and Snapshot Views

tlangs:
Added loading indicators on the Dataset and Snapshot tables of the Home view. I was able functionalize SnapshotTable, but trying to do the same to SnapshotView caused an infinite loop that I couldn't figure out how to prevent.


snf2ye (approved):

src/reducers/query.js
```diff
@@ -28,6 +28,22 @@ export const queryState = {
 export default {
   query: handleActions(
     {
+      [ActionTypes.GET_SNAPSHOTS]: (state) =>
```
nmalfroy: I think that this is the wrong reducer.  This should be in the snapshots reducer I think.  Also, it's never getting unset on failure

nmalfroy (changes requested):
I'm not sure about the approach.  I was kinda thinking we might introduce a request object to wrap requests so that we can set loading status consistently.  Might be worth holding off on this one since it's right in the middle of a lot of the changes I'm doing around typescript

package.json
```diff
@@ -38,6 +38,7 @@
     "@types/react": "^17.0.39",
     "@types/react-dom": "^17.0.11",
     "@types/react-jss": "^10.0.0",
+    "@typescript-eslint/eslint-plugin": "^5.18.0",
```
nmalfroy: I think this should be a dev dependency

nmalfroy (approved):
Looks great!  I think you just need to chance the dev dependency thing

PR #1107, opened 2022-04-11
Remove e2e cypress recording

tlangs:
Testing out if removing the recording functionality from the e2e tests speeds things up. Our tests currently take 30 minutes on github runners, but much much less time locally. The recording functionality is one of the main differences, and the recordings from github runners are not really of much use anyways.


tlangs:
This didn't actually end up helping anything.

PR #1112, opened 2022-04-20
[DR-2471] Add new snapshot overview page

samanehsan:
Add a new snapshot overview page that is consistent with the existing dataset overview page.  I tried to break up the changes into meaningful commits to make this easier to review. Changes include:
- Generalizing the DatasetOverviewSchemaPanel
- Moving the TabPanel component to its own file
- Add files for new snapshot overview page
- Convert shared components (SchemaPanel and TabPanel) to typescript
- Add tests
- Remove old snapshot overview components

Changes:
Screen Shot 2022-04-20 at 11 20 45 AM

Screen Shot 2022-04-20 at 11 20 55 AM


src/components/snapshot/overview/SnapshotOverview.tsx
```diff
@@ -0,0 +1,118 @@
...
+interface RouteParams {
+  uuid: string;
+}
+
+type AllSnapshotProps = StateProps & WithStyles<typeof styles> & RouteComponentProps<RouteParams>;
+
+function SnapshotOverview(props: AllSnapshotProps) {
+  const { classes, match, snapshot, snapshotPolicies } = props;
+  const snapshotId = match.params.uuid;
+  const dispatch = useDispatch();
```
samanehsan: @nmalfroy I can get `dispatch` directly in a component or define it as a component [property](https://github.com/DataBiosphere/jade-data-repo-ui/blob/2028e65f9f8e3535f7aca588c2f46f7f10c7892f/src/components/snapshot/overview/SnapshotExport.tsx#L42) (I'm not sure where that gets passed in though). Is one method preferred over the other?

samanehsan: As discussed offline, the `connect` function adds `dispatch` to the component properties so we can define it as a component property instead of calling `useDispatch()`

nmalfroy:
Do we need to bump up the timeout for the unit tests?

package.json
```diff
@@ -70,6 +70,7 @@
     "redux-actions": "^2.6.5",
     "redux-immutable-state-invariant": "^2.1.0",
     "redux-logger": "^3.0.6",
+    "redux-mock-store": "^1.5.4",
```
nmalfroy: I think this should be a dev dependency no?

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
```diff
@@ -7,7 +7,7 @@ type Context = { type: BreadcrumbType; id: string; name: string };
 type Breadcrumb = { text: string; to: string };
 
 type AppBreadcrumbsProps = {
-  childBreadcrumbs: Breadcrumb[];
+  childBreadcrumbs?: Breadcrumb[];
```
nmalfroy, line 10: Are we really expecting this to ever be null?

nmalfroy (approved):
This looks great! Any chance you can deploy this on your personal deployment so we can click test?

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
samanehsan: It's undefined on the dataset and snapshot overview pages (e.g. https://github.com/DataBiosphere/jade-data-repo-ui/blob/develop/src/components/dataset/overview/DatasetOverview.jsx#L76) and then the AppBreadCrumbs component sets the default to `[]`. Do you think explicitly passing in an empty list would be clearer?

package.json
samanehsan: You're right! Good catch!

src/components/AppBreadcrumbs/AppBreadcrumbs.tsx
nmalfroy: Yeah, I think probably forcing an empty list (if that's our default anyways) makes sense

samanehsan:
> Any chance you can deploy this on your personal deployment so we can click test?

Yes! I deleted my deployment last quarter to cut down costs so I'm getting that set up again

okotsopoulos (approved):
I interacted with snapshots on your dev deployment, what a huge improvement in user experience!  Thank you!

PR #1115, opened 2022-04-25
DR-2505 Enable B2C (and all other) Oauths

nmalfroy:
This PR should be fairly consumable commit-by-commit.  It:
- Makes it so we don't use gapi for authing to google anymore (or talking to BQ)
- Reworks how we handle auth to be more redux-y
- When we log in, we now fetch user info.  Note: this is currently to make sure that existing users are migrated to B2C but longer term, this allows us to track if users are registered in terra and indicate accordingly if that's the case)
- Adds a fallback to make sure that even is there is no user picture, they get a legit avatar
- Some more typescriptification

Instance deployed at:
https://jade-nm.datarepo-dev.broadinstitute.org/


src/index.jsx
```diff
@@ -53,10 +52,8 @@ function bootstrap() {
   return new Promise((resolve, reject) => {
     // We need to do this (in order) before bothering to render:
     // 1. Check status of the server
-    // 2. grab the configuration + load Google's auth2 library (in parallel)
```
snf2ye, line 56: For the Google Sheets PR, I'm relying on loading the google client as a part of this call ([see spike PR](https://github.com/DataBiosphere/jade-data-repo-ui/pull/1114/files#diff-fd91852876d8943d1f7adbb51e497d186783e17321432d8b886bdf51b31fdf38R12)). Not sure how that will translate.

nmalfroy: Ok cool.  I think we can maybe add some branching logic to only show the sheet thing when we log in with google?  I wonder if maybe there's a way to use the api with the access token (provided we get the right scopes from B2C)

src/containers/App.jsx
```diff
@@ -156,6 +155,22 @@ export function App(props) {
...
+        auth.isAuthenticated &&
+        !auth.isLoading &&
+        !user.isAuthenticated &&
+        !auth.activeNavigator
+      ) {
+        dispatch(logInSuccess(auth.user));
+        dispatch(getUserStatus());
+      }
+    })();
+  }, [auth, dispatch, user]);
```
snf2ye: I think there is an extra layer of parenthesis in here. 
```suggestion
  useEffect(() => {
      if (
        auth.isAuthenticated &&
        !auth.isLoading &&
        !user.isAuthenticated &&
        !auth.activeNavigator
      ) {
        dispatch(logInSuccess(auth.user));
        dispatch(getUserStatus());
      }
  }, [auth, dispatch, user]);
```

snf2ye (approved):

samanehsan:
When I sign in with a google account and open up a new tab, the new tab shows the login page. Is that expected to happen with these changes?

Also, when I log in using the Azure multi-tenant option I get a bunch of errors because the account is not registered in Terra. Does it need to be registered via the Terra API? I can only log in with a google account through the UI 🤔

samanehsan (approved):
Lgtm! I just had a couple of questions around the login behavior but I think it's fine to make a follow-on PR if needed. :octocat:

PR #1119, opened 2022-04-26
[DR-2568] Test fixing java install error

samanehsan:
Installing OpenJDK 11 fails because it is looking for an Ubuntu distribution ("kinetic") that does not exist in https://launchpad.net/~openjdk-r/+archive/ubuntu/ppa.

This changes the base image from one that uses Debian 9 to Debian 10 (which includes OpenJDK 11 as the default Java version) so that installing from the ppa is no longer necessary.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,7 @@ FROM node:14.0-buster
 
 # Install java to run codegen
 RUN apt-get update && \
-    apt-get install -y software-properties-common && \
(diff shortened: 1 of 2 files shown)


nmalfroy (approved):
Woo! Nice job figuring it out!

okotsopoulos (approved):
Thanks for fixing this!

PR #1120, opened 2022-04-27
DR-2570 Fix URI that gets generated

nmalfroy:
This fixes the URL when linking an export to a workspace


okotsopoulos (approved):

tlangs (approved):
That is sooooooo annoying!

PR #1128, opened 2022-05-19
DR-2559 - BQ Connected Google Sheets

snf2ye:
Latest deployed here - https://jade-sh.datarepo-dev.broadinstitute.org/

Both on the snapshot and dataset page:

image
image

Please ignore the very ugly list of commits 😅


src/setupProxy.js
```diff
@@ -40,4 +42,19 @@ module.exports = (app) => {
...
+      pathRewrite: { '^/googlesheets': '' }
+    }),
+  );
+  app.use(
+    '/drive',
+    createProxyMiddleware({
+      target: driveApi,
+      changeOrigin: true,
+    }),
+  );
```
snf2ye, line 59: This is needed in order to do the worksheet delete on failure.

src/modules/notifications.jsx
```diff
@@ -10,7 +10,7 @@ export function showNotification(err) {
     const errDetail = _.get(err.response, 'data.errorDetail');
     status = String(err.response.status);
     if (_.isEmpty(errDetail)) {
-      message = _.get(err.response, 'data.message');
+      message = _.get(err.response, 'data.message') ?? _.get(err.response, 'data.error.message');
```
snf2ye, line 13: This allows us to get a more meaningful error message when a user logs in with a non-enterprise workspace google account (Which is not allowed to use this feature - https://docs.google.com/document/d/1YG7UpBpExxHViwcC-MIYji7mwhlh9vQSfYwxoSxIcpA/edit#heading=h.h4uq7f5zs05h)

src/components/common/overview/GoogleSheetExport.tsx
```diff
@@ -0,0 +1,139 @@
...
+      );
+      // If no BQ sources added to sheets object, then something errored.
+      if (sheets.length > 0) {
+        await cleanupSheet(response.spreadsheetId, sheets, token);
+        setIsSheetProcessing(false);
+        setIsSheetDone(true);
+      } else {
+        await deleteSpreadsheetOnFailure(response.spreadsheetId, token);
+        // reset create sheet button if not successful
+        setIsSheetProcessing(false);
```
samanehsan: You could call `resetCreate` here

src/modules/googlesheets.ts
```diff
@@ -0,0 +1,149 @@
...
+  for (const table of bigQueryAccessInfo.tables) {
+    const requests: object[] = [
+      {
+        addDataSource: {
+          dataSource: {
+            spec: {
+              bigQuery: {
+                projectId: bigQueryAccessInfo.projectId,
+                querySpec: {
+                  rawQuery: `select * from \`${bigQueryAccessInfo.projectId}.${bigQueryAccessInfo.datasetName}.${table.name}\``,
```
samanehsan: Putting my HackEDU hat on, I'm concerned by the raw SQL query here. Even though it's not based on user input, someone could theoretically intercept and modify the API response right?

Could we add some sanitization here? We could also add ANTLR so we’re doing the same query parsing as our API, but that's probably worth splitting out into a separate ticket.

snf2ye: Thanks for putting on your HackEDU Hat 😄 
The main reason why we don't have to be too concerned here is that the scope defined here is only for read-only access to big query. 

If still want to avoid building a query in the UI (which is totally fair), then we could use "sampleQuery" provided by the getDataset/snapshot endpoints. I just have to remove the "LIMIT 1000" clause so the sheet has access to the whole dataset. I just put up a draft PR and would appreciate any thoughts on that approach! https://github.com/DataBiosphere/jade-data-repo/pull/1286

src/components/common/overview/GoogleSheetExport.tsx
snf2ye: Good point! Will update.

src/modules/googlesheets.ts
samanehsan: That's reassuring! If it's not too big of a lift I think using the sampleQuery would be great

src/modules/googlesheets.ts
```diff
@@ -0,0 +1,149 @@
...
+  for (const table of bigQueryAccessInfo.tables) {
+    const requests: object[] = [
+      {
+        addDataSource: {
+          dataSource: {
+            spec: {
+              bigQuery: {
+                projectId: bigQueryAccessInfo.projectId,
+                querySpec: {
+                  rawQuery: table.sampleQuery,
```
nmalfroy, line 60: I think this is fine for this PR but longer term, I think we may want to change either the api or the query here to explicitly use columns so that we have consistent column ordering

nmalfroy (approved):
this looks great!

src/modules/googlesheets.ts
snf2ye: Ah, great point. I would lean towards updating the sample query, but I suppose that could get a bit unmanageable for tables with a lot of columns. I'll put in a ticket.

snf2ye: https://broadworkbench.atlassian.net/browse/DR-2604

nmalfroy: awesome, thanks!

samanehsan (approved):

PR #1130, opened 2022-05-24
DR-2559 Follow-on - Fix Google Sheets Export for B2C

snf2ye:
3 Changes:
- Slightly better error handling
- Point to delegate token for google drive / google sheets api calls
- Point to delegate token for bigquery changes


nmalfroy (approved):
Thanks for fixing the panel too!

samanehsan (approved):

PR #1132, opened 2022-05-31
DR-2610 Fix Oauth Timeout with refresh token

nmalfroy:
This is an example of how refreshing access tokens via a refresh token could work.

I've deployed this on:
https://jade-nm.datarepo-dev.broadinstitute.org


dinvlad:
I think we need to add removal of the token to the existing logic dealing with explicit/15-minute logout also.

nmalfroy:
Ah interesting.  I think it does an explicit log out so it clears out the refresh token but I can confirm

dinvlad:
Oh I see - the entire "user data" config (including refresh token) is removed upon logout right?

src/index.jsx
```diff
@@ -103,27 +103,38 @@ function bootstrap() {
...
     prompt: 'login',
     scope: scopes.join(' '),
+    loadUserInfo: isGoogleAuthority,
     userStore: new WebStorageStateStore({ store: window.localStorage }),
+    extraQueryParams: {'access_type': 'offline'},
   };
 
+  // Log oauth client events
+  Log.setLevel(Log.WARN);
+  Log.setLogger(console);
```
dinvlad, line 138: Please make sure we don't log the tokens accidentally, even if it's just via `console.log` :)

nmalfroy: haha oh totally.  This is still very much a draft

src/index.jsx
```diff
@@ -103,27 +103,38 @@ function bootstrap() {
 function render(Component) {
   const root = document.getElementById('react');
   const { configuration } = store.getState();
-  const googleAuthority = 'https://accounts.google.com';
-  // TODO once we no longer query BQ directly, we'll no longer need the BQ scope
+  const isGoogleAuthority = configuration.configObject.authorityEndpoint.startsWith('https://accounts.google.com');
```
dinvlad: Which format does the new authority endpoint take that we can no longer use strict equality?

nmalfroy: Oh it could have a slash at the end

nmalfroy:
That's

> Oh I see - the entire "user data" config (including refresh token) is removed upon logout right?

That's correct. When we do the 15 minute logout, it goes one step further and actually logs the user out of google

src/index.jsx
snf2ye: Sanity check - Anything leftover from when this was a draft?

snf2ye (approved):

src/index.jsx
samanehsan: ~Was this just for debugging?~ Oops I didn't see Shelby's comment until I reloaded the page!

nmalfroy:
I left that there (it's super minimal logging) that way it's easier for future us to add more logging locally

samanehsan (approved):

PR #1137, opened 2022-06-10
DR-2636 Revoke Oauth2 refresh token on logout

nmalfroy:
...specifically when using google (B2C doesn't support revoking refresh tokens though their refresh tokens expire)

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -118,6 +118,7 @@ function render(Component) {
     scopes.push('https://www.googleapis.com/auth/drive.file');
 
     metadata.userinfo_endpoint = 'https://openidconnect.googleapis.com/v1/userinfo';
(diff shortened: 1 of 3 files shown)


okotsopoulos (approved):

snf2ye (approved):

src/index.jsx
```diff
@@ -130,6 +131,7 @@ function render(Component) {
     loadUserInfo: isGoogleAuthority,
     userStore: new WebStorageStateStore({ store: window.localStorage }),
     extraQueryParams: { access_type: 'offline' },
+    revokeTokensOnSignout: true,
```
rtitle, line 134: Cool - doing this in Terra UI as well :)

nmalfroy: Nice :)

PR #1146, opened 2022-06-27
DR-2485: Combine JadeTable and LightTable

snf2ye:
## Background
We have two table components in the UI that are almost identical: JadeTable and LightTable. This separation means that we've created features and updated styling in slightly different ways between the two tables. In order to avoid inconsistencies in the UI and to remove the need to update code for tables in multiple places, we should combine these components. 

Changes deployed to my [personal dev environment](https://jade-sh.datarepo-dev.broadinstitute.org/datasets/c5248002-05d6-469e-bf31-38f1792198f3/data):
- Potentially breaking: Any table could have been broken in the process in combining
- Biggest change can be seen in the data view pages for datasets and snapshots

## Changes included in PR
- [x]  LightTableHead converted to Functional Component
- [x] LightTableHead, LightTable, DataView components converted to Typescript
- [x] Removed DatasetTablePreview Component - As far as I can tell it's unused
- [x] Point DataView to use LightTable instead of JadeTable - means that dataset and snapshot data pages now use LightTable. 
- [x] New way of handling case where no rows are displayed in table
- [x] Converted LightTable to use page and rowsPerPage from redux rather than component state
- [x] Move ability to open long values in "see more" dialog box from JadeTable to LightTable
- [x] Move ability to handle null values from JadeTable to LightTable
- [x] Styling Adjustments
- [x] Delete JadeTable and JadeTableHead Component

## Changes in action
With Changes:

https://user-images.githubusercontent.com/13254229/176681536-fc304c4d-587b-4b32-94bd-25366e28ec67.mov

Before Changes:

https://user-images.githubusercontent.com/13254229/176682089-1fe3eb2c-a789-47fb-9f30-08f6dfde36c8.mov


src/components/common/data/DataView.tsx
```diff
@@ -1,46 +1,81 @@
...
   selected,
   selectedTable,
   sidebarWidth,
   tableNames,
-}) {
+  totalRows,
+}: DataViewProps) {
+  const rowKey = (row: TableRowType): string => {
+    const drId = row.datarepo_row_id;
+    return drId ?? row.name;
```
nmalfroy, line 89: huh...wasn't familiar with this syntax.  What does that do?

src/components/common/data/DataView.tsx
```diff
@@ -81,11 +116,17 @@ function DataView({
           <Grid container spacing={0}>
             <Grid item xs={11}>
               <div className={classes.scrollTable}>
-                <JadeTable
+                <LightTable
+                  columns={columns}
+                  filteredCount={totalRows}
+                  loading={polling}
+                  noRowsMessage="No rows exist in the table" // TODO - handle case where filtered
```
nmalfroy: Is this outside the scope of this PR?

nmalfroy (approved):
Looks awesome!  Code looks good and I'm not sure if you're adding more based on reviews with Lou.  You may want to add a margin at the bottom of the table to makes sure there is clearance on the page

src/components/common/data/DataView.tsx
snf2ye: > The nullish coalescing operator (??) is a logical operator that returns its right-hand side operand when its left-hand side operand is null or undefined, and otherwise returns its left-hand side operand.

For the non-data view tables, datarepo_row_id will be null, so we can just use the row name.

nmalfroy: oh cool.  it's like a coalesce?

snf2ye: I use it as a short-hand for a ternary statement - the above is equivalent to`drId != null ? drId : row.name`.

lstrano-broad:
Styling looks good - met with Shelby on minor tweaks on pagination and table width

okotsopoulos (approved):
Thanks for taking on this consolidation!  I interacted with your dev environment versus the existing styling, it looks good.  Calling out that with these changes, hovering on a row no longer highlights it, though with the zebra striping it's generally easier to visually connect a row's values across columns.

snf2ye:
@okotsopoulos 
> hovering on a row no longer highlights it

Thank you for catching that!! I added the hover back - it was a property on the JadeTable, but not LightTable.  I didn't realize we had this feature.

src/components/common/data/DataView.tsx
```diff
@@ -0,0 +1,180 @@
...
+              filteredCount={totalRows}
+              loading={polling}
+              noRowsMessage={
+                isDatasetFiltered ? 'No rows match your filter' : 'No rows exist in the table'
+              }
+              pageBQQuery={pageBQQuery}
+              rowKey={rowKey}
+              rows={rows}
+              searchString={filterStatement}
+              totalCount={totalRows} // TODO - DR-2663 - instead should display total rows regardless of filtering
```
snf2ye, line 144: Note: Ticket added to address TODO comment - https://broadworkbench.atlassian.net/browse/DR-2663

src/components/common/data/DataView.tsx
snf2ye: Good point - Now should be updated!

PR #1159, opened 2022-07-29
DR-2681 Sort data columns when previewing snapshots

nmalfroy:
This PR does enables sorting on columns in the snapshot preview (this already works with dataset preview)

I changed how the fetch gets triggered to make it so that data is fetched when ordering changes.

I also did a drive-by change to make sure that booleans render correctly in the UI

You can see this in action at:
https://jade-nm.datarepo-dev.broadinstitute.org/snapshots/9cbb702c-51f8-4396-85e0-9ad2b43aedb7/data

### DIFF
--- a/src/actions/app.js
+++ b/src/actions/app.js
@@ -177,7 +177,15 @@ export const { runQuery } = createActions({
 });
 
 export const { previewData } = createActions({
-  [ActionTypes.PREVIEW_DATA]: (resourceType, resourceId, table, columns, totalRowCount, orderDirection, orderProperty) => ({
+  [ActionTypes.PREVIEW_DATA]: (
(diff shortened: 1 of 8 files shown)


samanehsan (approved):
This looks great! 🎉

okotsopoulos (approved):
Thank you for the opportunity to interact with your changes, and for the sprint review demo!

I know that you are going to fix the page numbering, but otherwise this looks good.

A question / not a blocker to merge: I see the current behavior is to sort descending on the first column click, then ascending.  Is this intuitive or should the order be reversed?

nmalfroy:
Yeah, I noticed that too and it bugs me too. I'll fix that

PR #1161, opened 2022-08-03
[DR-2677] Conditionally disable export snapshot button

samanehsan:
Only enable exporting snapshots to a workspace if the user is a Steward. Otherwise, disable and make the tooltip display an error message:

Screen Shot 2022-08-03 at 12 58 30 PM

For testing, I created two snapshots in my personal environment. The team has read-access to one and steward-access to the other: https://jade-se.datarepo-dev.broadinstitute.org/snapshots


fboulnois (approved):
lgtm

okotsopoulos (approved):
Thank you, looks good!  Even this getting released absent the changes to SAM and the API would improve the UX, IMO.

PR #1171, opened 2022-08-25
DR-2693 Allow exporting to Azure

nmalfroy:
This PR enables exporting Azure snapshots to Terra workspaces.  Specifically:
- Allow skipping Uniqueness validation for Azure snapshots (since it's not implemented)
- Hide the "resolve to gs:// paths" checkbox (since it doesn't make sense for Azure)
- Hides the export to sheets button (since it doesn't make sense for Azure)

Of note: this also uses the dev instance's swagger to generate the TDR client

Looks like this:
image

### DIFF
--- a/src/components/snapshot/overview/SnapshotExport.tsx
+++ b/src/components/snapshot/overview/SnapshotExport.tsx
@@ -57,7 +57,9 @@ const formatExportUrl = (
   snapshot: SnapshotModel,
   manifest: string,
 ) =>
-  `${terraUrl}#import-data?url=${window}&snapshotId=${snapshot.id}&format=tdrexport&snapshotName=${snapshot.name}&tdrmanifest=${manifest}`;
+  `${terraUrl}#import-data?url=${window}&snapshotId=${snapshot.id}&format=tdrexport&snapshotName=${
+    snapshot.name
+  }&tdrmanifest=${encodeURIComponent(manifest)}`;
 
 function SnapshotExport(props: SnapshotExportProps) {
   const [exportGsPaths, setExportGsPaths] = React.useState(false);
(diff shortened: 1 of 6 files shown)


src/components/snapshot/overview/SnapshotOverviewPanel.tsx
```diff
@@ -63,7 +63,7 @@ function SnapshotOverviewPanel(props: SnapshotOverviewPanelProps) {
   const { classes, snapshot } = props;
   // @ts-ignore
   const sourceDataset = snapshot.source[0].dataset;
-  const linkToBq = snapshot.accessInformation?.bigQuery !== undefined;
+  const linkToBq = snapshot.cloudPlatform === 'gcp';
```
fboulnois, line 66: Is there a situation where the `cloudPlatform` is `gcp` but it is not possible to link to Big Query?

src/components/snapshot/overview/SnapshotExport.tsx
```diff
@@ -109,18 +112,20 @@ function SnapshotExport(props: SnapshotExportProps) {
...
+      {of.cloudPlatform === 'gcp' && (
+        <FormGroup>
+          <FormControlLabel
+            data-cy="gs-paths-checkbox"
+            control={gsPathsCheckbox}
+            label="Convert DRS URLs to Google Cloud Storage Paths (gs://...)"
+          />
+          <FormHelperText>
+            <i>
+              <b>Note: </b> gs-paths can change over time
```
fboulnois:
```suggestion
              <b>Note:</b> gs-paths can change over time
```

fboulnois (approved):
Overall lgtm, some minor feedback / suggestions:

src/components/snapshot/overview/SnapshotOverviewPanel.tsx
nmalfroy: not really. this page gets hit if you have access and on GCP you should always have access to the BQ DB at that point

okotsopoulos (approved):
Looks good!  Thanks for demoing this during Design Discussion today.

PR #1175, opened 2022-09-06
DR-2701 Add column resizing to tables and make table consistent with Terra UI

nmalfroy:
Adds the ability to resize table columns (if resizable is specified in columns and all column widths are populated)

This PR also made some changes to the table to make it consistent with the Terra UI table, namely:
- Colors match
- Sticky headers
- Borders match

I also made some drive-by fixes to some react warning (e.g. we had some null keys in arrays that were causing some warnings)

The PR should be good to review commit-by-commit (I isolated pulling in react draggable)

~~I'll deploy what I've got to my local instance once it builds~~
I've deployed my changes at:
https://jade-nm.datarepo-dev.broadinstitute.org if you'd like to click around

image


src/components/table/LightTable.tsx
```diff
@@ -241,38 +261,54 @@ function LightTable({
...
+  const tableWidth = columns.reduce((agg, column) => agg + (column.width || NaN), 0);
+  const effectiveTableWidth = _.isNaN(tableWidth) || !supportsResize ? '100%' : tableWidth;
   return (
     <div>
       {!loading && (
         <Paper className={classes.root}>
-          <div className={classes.tableWrapper}>
-            <Table className={classes.table}>
-              <LightTableHead columns={columns} onRequestSort={handleRequestSort} />
+          <TableContainer className={classes.tableWrapper} sx={{ maxHeight: 300 }}>
```
otchet-broad: Intentional change of the maxHeight?  If the tableWrapper class is being applied to the TableContainer, why not just change the maxHeight there?

nmalfroy: derp...no, my bad.  Left over from trying to get the table to behave.  Will remove

otchet-broad: Screen Shot 2022-09-06 at 10 20 54 AM

nmalfroy: right exactly. The class takes care of it.

src/components/table/LightTableHead.tsx
```diff
@@ -40,44 +83,99 @@ function LightTableHead({
...
+      onResizeColumn(event, column.name, initialWidth + data.x);
+    }
+  };
+
+  const createDragHandle = (col: TableColumnType) =>
+    col.allowResize && (
+      <Draggable
+        axis="x"
+        onStart={createStartHandler(col)}
+        onStop={createStopHandler(col)}
```
otchet-broad, line 147: Would you consider updating the column width while dragging is in progress?  As a user, I tend to like to resize and see the impact at the same time.

otchet-broad (approved):

src/components/table/LightTableHead.tsx
okotsopoulos: +1, not a dealbreaker but it was something I was wondering about in your demo (thanks @otchet-broad for putting it into words 😄 )

nmalfroy: Yeah, I agree.  The only reason I didn't is that the Terra table doesn't do this and I've found in the past that can get kinda chunky as an interaction.  I think if I make the table virtualized, it might work but it's a pretty hefty change and wouldn't match the terra UI.  I'll mention it to Lou

okotsopoulos (approved):
This is going to be a nice improvement, thanks!  Appreciated getting to click around in your dev environment.

Nothing I consider a blocker to merge, but I wanted to draw attention to the default rendering with few columns that don't fill up the table width.  A more natural default behavior may be to increase the starting column width for fewer than `n` columns so that in all cases the columns minimally fill the table.

Screen Shot 2022-09-07 at 11 10 09 AM

nmalfroy:
Yeah, that's a totally fair point.  I was trying to think of a clever way to draw the border _only_ if it doesn't fill up the table width. Let me see if I can figure something out

okotsopoulos:
Clicked around again, my approval still stands and is even more enthusiastic :)

The dynamic resizing is slick!  Thanks for adding that in.

PR #1177, opened 2022-09-09
DR-2739 Force group syncing for now when exporting snapshots to workspaces

nmalfroy:
This will allow the AJ team introduce the new parameter on their end without changing the current behavior.  This should not cause any change in behavior


jordansafer (approved):

okotsopoulos (approved):
Looks good! I understand that this is a short-term measure. Were you able to click test?

nmalfroy:
> Looks good! I understand that this is a short-term measure. Were you able to click test?

I was.  I successfully ignores

PR #1182, opened 2022-09-15
[DR-2712] Adds support for using DescriptionView to update Snapshot description…

otchet-broad:
- Generalizes DescriptionView so it can be used for both Snapshot and Description editing.
- Converts null usage to undefined in DescriptionView to match patterns in the SnapshotOverviewPanel.
- Updates SnapshotOverviewPanel and related components so they can make the "canEdit" decision for the DescriptionView component
- Updates the DatasetOverviewPanel so it can make the canEdit decision for the DescriptionView component.
- Adds code to effect the persisting of the description.
- Updates the test and selector used by the SnapshotOverviewPanel to reflect the use of the DescriptionView component


src/sagas/repository.ts
```diff
@@ -268,6 +268,21 @@ export function* getSnapshots({ payload }: any): any {
   }
 }
 
+export function* patchSnapshotDescription({ payload }: any): any {
```
lu-c, line 271: What does a `function*` do?

otchet-broad: This indicates the function is a Generator
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*

Generators are nice because they support yield
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield

This helps deal with managing callbacks.  There are other approaches, but this seems to be a good one for this codebase and is already the pattern in use.

okotsopoulos (approved):
This looks good for me, glad that our theory on the smaller scope after the description ticket was correct!  Thank you for the PR description.

lu-c:
Confirmed this works locally 👍

lu-c (approved):

PR #1185, opened 2022-09-16
DR-2729 Add utility button so an admin can add themselves to a dataset or snapshot

nmalfroy:
This is to address a long running annoyance when debugging where admins have to go into swagger to add themselves as users to a dataset or snapshot to help debug issues.

This adds a button if a user has *only* the admin role on the dataset or snapshot (which can only happen if a user is in the admin group).  Here's an screenshot:

image

(note this looks the same in the snapshot).

Part of the challenge with this PR is that we needed a way to refresh the datasets/snapshots list without losing filter state (to reload so that we could see the result of adding one's self).  To solve this: I created `REFRESH_DATASETS`, `REFRESH_SNAPSHOTS`, and `REFRESH_QUERY` (for the data pages...not used yet but needed for consistency and we will eventually need it) actions.  These actions increment a counter (`refreshCnt` in the various redux sub-stores).  That `refreshCnt` counter is tracked in the table component so that if it changes, it triggers a refresh of the data.

The other challenge is that I converted more files into Typescript, so that adds a bit of overhead.

I've tried to make the commits meaningful for ease of reviewing

Also worth noting: I made a drive by change to fix an issue with the edit button on description fields overlapping on smaller screens

I also updated the column widths on the snapshots list page and add some ellipsis overflow text handling (I'd forgotten to do that in my last PR)

Edit:
Ok, in my quest to get the tests to pass I ended up:
- Adding logic to make sure that columns are in a consistent order
- Using `scrollTo` in tests to make sure that everything is in view
- Doing a bunch of work to speed up our builds!
  - Now, most tasks run with `DISABLE_ESLINT_PLUGIN=true` which makes everything run faster (and we don't need it since we explicitly run linting)
     - Note: I recommend that users set in their environment variables locally (I updated the README).  That makes local server/cypress startup go from 11 minutes to like 1 minute
  - Use a docker image to run codegen (no more downloading java!)


cypress/integration/queryBuilder.spec.js
```diff
@@ -33,7 +33,7 @@ describe('test query builder', () => {
 
     it('does show array values', () => {
       // Click the header to make sure we don't sort or trigger an error
-      cy.get('[data-cy=columnHeader-consequence_terms]').click();
+      cy.get('[data-cy=columnHeader-consequence_terms]').click({ force: true });
```
otchet-broad:
```suggestion
      cy.get('[data-cy=columnHeader-consequence_terms]').scrollIntoView().click({ force: true });
```

src/components/DescriptionView.tsx
```diff
@@ -11,7 +11,7 @@ const styles = (theme: CustomTheme) => ({
     display: 'flex',
   },
   iconDiv: {
-    width: '2%',
+    width: 24,
```
otchet-broad, line 14: Nice!

otchet-broad (approved):
Looks good after resolving the column ordering or scrolling the column into view.

okotsopoulos (approved):
Thanks for the multiple quality-of-life improvements!

src/components/table/ResourceName.tsx
```diff
@@ -0,0 +1,85 @@
...
+      '& svg': {
+        height: 20,
+        fill: theme.palette.primary.main,
+      },
+    },
+  });
+
+function hasAdminOnlyAccess(id: string, roleMaps: { [key: string]: Array<string> }) {
+  const roles = roleMaps[id];
+  return _.isArray(roles) && roles.length === 1 && roles[0] === 'admin';
```
otchet-broad, line 34: Any reason to use _.isArray instead of Array.isArray?

https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L11286

nmalfroy: Mostly just that I'm already familiar with lodash :) I'm not sure if there's a benefit of one over the other.  I like the idea of lodash since there's the one entry point to do all of the things

src/components/DatasetView.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+function mapStateToProps(state: TdrState) {
+  return {
+    datasets: state.datasets.datasets,
+    datasetRoleMaps: state.datasets.datasetRoleMaps,
+    datasetsCount: state.datasets.datasetsCount,
+    filteredDatasetsCount: state.datasets.filteredDatasetsCount,
+    features: state.user.features,
+    loading: state.datasets.loading,
+    user: state.user.email,
```
otchet-broad: Would you consider using userEmail (or just email) since a complex?  'user' object exists on state?

src/components/table/ResourceName.tsx
otchet-broad: I'll save you the link trip.  _.isArray just calls Array.isArray.

nmalfroy: Haha, I may just leave it (since I like having the facade with lodash and it's not any more complicated)

src/components/DatasetView.tsx
nmalfroy: Oh, I most definitely would.  Let me update

src/components/table/ResourceName.tsx
otchet-broad: After playing with this locally for a bit would you consider a slight tweak to the logic?
return _.isArray(roles) && roles.includes('admin') && !roles.includes('steward');

Handles the case where you remove yourself as a steward, remain a custodian or reader, and want to let yourself back as a steward.

nmalfroy: Yeah, I was debating. I guess I like the simplicity of it now and it has the added benefit of working for both snapshots and datasets (datasets and snapshots have steward and admin but after that they diverge) so if it's ok, I'd like to keep it as is

PR #1188, opened 2022-09-24
[DR-2741][DR-2742] Added tab for roles and memberships

lu-c:
BEFORE
Screen Shot 2022-09-24 at 3 10 15 PM

AFTER
Screen Shot 2022-09-24 at 3 10 52 PM
Screen Shot 2022-09-24 at 3 09 47 PM

Made some changes to the ticket ask, will outline them in the comments section!

1. Added new tab for the roles and memberships access information
2. removed the columns from the dataset summary view


src/components/dataset/DatasetAccess.jsx
```diff
@@ -30,6 +41,14 @@ function DatasetAccess(props) {
 
   return (
     <Grid container spacing={1}>
+      {showHelp ? (
+        <div className={classes.helpContainer}>
+          <a href="#" className={classes.genericLink}>
```
lu-c: Not sure where this link is supposed to go - Please advise! :)

src/style.css
```diff
@@ -39,11 +39,6 @@ body.tdr table {
   border-collapse: collapse;
 }
 
-body.tdr a {
```
lu-c: Removed this block - the comment from the commit it was made on said that the purpose was to enable typescript conversion, but I think that might've been related to something else in that same pr. This `a` style was overriding the style assigned on the component level to links. Is there a better way to do this? Should we try to keep this?

src/components/dataset/DatasetAccess.jsx
```diff
@@ -20,7 +31,7 @@ function DatasetAccess(props) {
       dispatch(removeDatasetPolicyMember(dataset.id, removableEmail, role));
     };
   };
-  const { horizontal, policies, userRoles } = props;
+  const { horizontal, policies, userRoles, showHelp, classes } = props;
   const stewards = getRoleMembersFromPolicies(policies, DatasetRoles.STEWARD);
   const custodians = getRoleMembersFromPolicies(policies, DatasetRoles.CUSTODIAN);
   const snapshotCreators = getRoleMembersFromPolicies(policies, DatasetRoles.SNAPSHOT_CREATOR);
```
lu-c, line 37: The ticket did not ask for snapshot creators to get their block, but they did have a column. I decided to leave these in there instead of removing them completely. Is that okay?

src/components/dataset/overview/DatasetOverviewPanel.jsx
```diff
@@ -89,6 +89,13 @@ function DatasetOverviewPanel(props) {
             {...a11yProps(2)}
           />
         )}
+        <Tab
+          label="Roles & memberships"
```
lu-c: The ticket said to name this tab "access permissions", but the mock up said to name it "roles and memberships" - which one should we follow?

src/components/dataset/DatasetAccess.jsx
nmalfroy: oh, that should make a sidebar pop out.  Check out in the dataset info view:
image

src/style.css
nmalfroy: oh yeah, let's keep this. The way that we get links to work is to render a span inside of the link

src/components/dataset/DatasetAccess.jsx
nmalfroy: yeah, I think that makes sense

src/components/dataset/overview/DatasetOverviewPanel.jsx
nmalfroy: @lstrano-broad, I'm guessing what's in mocks but I'd love to confirm

src/components/dataset/overview/DatasetOverviewPanel.jsx
```diff
@@ -51,6 +51,9 @@ function DatasetOverviewPanel(props) {
   const { classes, dataset, dispatch, userRoles } = props;
   const linkToBq = dataset.accessInformation?.bigQuery !== undefined;
 
+  const canManageUsers =
+    userRoles.includes(DatasetRoles.STEWARD) || userRoles.includes(DatasetRoles.CUSTODIAN);
```
nmalfroy, line 98: I don't think custodians can add stewards so we may need to have more targeted checking

nmalfroy (dismissed):
This looks great! I do think that you need to re-add that top level CSS and have more targeted checks for access.  Also meta point, I've been trying to convert jsx files to tsx when I hit them to gradually make progress to convert the whole codebase.  Maybe not in this PR but maybe as you get more of these PRs under your belt we can start chipping away?

src/components/ManageUsersModal.jsx
```diff
@@ -102,7 +113,8 @@ export class ManageUsersModal extends React.PureComponent {
         variant="outlined"
         onClick={this.handleClickOpen}
       >
-        {modalText}
+        {/* <Edit fontStyle="small" className={classes.openButtonIcon} /> {modalText} */}
+        <i className={`${classes.openButtonIcon} fa-solid fa-pen-circle`} /> {modalText}
```
otchet-broad, line 116: This doesn't look like a blue circle with a white pencil in the attached screenshots.  Are the screenshots old or is this because DR-2735 is still pending?

src/components/UserList.jsx
```diff
@@ -46,11 +66,18 @@ class UserList extends React.PureComponent {
...
     } = this.props;
 
+    const { isAccordionExpanded } = this.state;
+
     return (
-      <div className={classes.root}>
-        <div className={classes.header}>
-          {typeOfUsers}:{' '}
-          {horizontal && canManageUsers && (
+      <Accordion expanded={isAccordionExpanded} onChange={this.toggleAccordion}>
```
otchet-broad: Is the 'isAccordionExpanded' state used by anything else?  I think you might be able to remove expanded, onChange and the state tracking of this Accordion.

The starting state of the current UI allows users to see the user lists without clicking.  Should these be expanded by default?  defaultExpanded should do the trick, but I suggest checking with Lou for the UX decision.

src/components/dataset/DatasetAccess.jsx
```diff
@@ -20,16 +31,25 @@ function DatasetAccess(props) {
...
 
-  const canManageUsers = userRoles.includes(DatasetRoles.STEWARD);
+  const canManageUsers =
+    userRoles.includes(DatasetRoles.STEWARD) || userRoles.includes(DatasetRoles.CUSTODIAN);
```
otchet-broad, line 41: @nmalfroy - the same note about Custodians not being able to create Stewards applies here too, right?

src/components/snapshot/overview/SnapshotOverviewPanel.tsx
```diff
@@ -68,6 +68,7 @@ function SnapshotOverviewPanel(props: SnapshotOverviewPanelProps) {
   const [value, setValue] = React.useState(0);
   const { classes, dispatch, snapshot, userRoles } = props;
   const canEditDescription = userRoles.includes(SnapshotRoles.STEWARD);
+  const canManageUsers = userRoles.includes(SnapshotRoles.STEWARD);
```
otchet-broad, line 71: More of an FYSA, I've got an upcoming change to rename canManageUsers to 'isSteward' because I'm starting to use it in a new component.  If you think it makes sense to introduce that concept in the decisions (especially the ones that need to be made on a more granular basis), run with it in this PR and I'll happily resolve the conflicts in my own branch.  :-)

otchet-broad (dismissed):

src/components/dataset/overview/DatasetOverviewPanel.jsx
lu-c: Added a variable for `canManageStewards`!

src/components/ManageUsersModal.jsx
lu-c: It's because that PR wasnt merged when I made this pr! :) I can update it now!

lu-c: Or actually wait, sorry, I already updated it. The screenshots are old, but the code is the pencil circle!

src/components/ManageUsersModal.jsx
```diff
@@ -102,7 +113,8 @@ export class ManageUsersModal extends React.PureComponent {
         variant="outlined"
         onClick={this.handleClickOpen}
       >
-        {modalText}
+        {/* <Edit fontStyle="small" className={classes.openButtonIcon} /> {modalText} */}
```
lu-c:
```suggestion
        
```

src/components/UserList.jsx
lu-c: I have asked Lou for what to do with the accordions initial state!

src/components/dataset/DatasetAccess.jsx
lu-c: I think I've addressed it here already for datasets. for the other one, the only people who can manage users for snapshots are stewards - the Custodian role doesnt exist at all, so I didnt touch that logic!

src/components/dataset/overview/DatasetOverviewPanel.jsx
lu-c: This section will continue to just use `canManageUsers` because it is the logic to show/hide the tab on the overview page, and we do a separate check again later inside datasetAccess to see which ones they have permission to manage

src/components/snapshot/overview/SnapshotOverviewPanel.tsx
lu-c: If I were to have a preference, I think I would rather keep the variable called `canManageUsers`, because maybe the conditions for that will change, but the abilities this bool unlocks should not. For example, custodians can manage some users, and not just "isSteward"

Would you be okay with keeping it as is? :)

otchet-broad: I'm thinking more and more about how to leverage ABAC to help clarify access decisions and in this example, the way I'm thinking about the security control is as follows:
Because I'm a data steward of this snapshot, I'm entitled to:
- update descriptions
- change who is authorized to steward, read and discover this snapshot's data (there are no "custodians" in snapshots)
- (and coming soon) who is able to see workspaces using this snapshot

When I introduced the canEditDescription in a previous commit, I didn't appreciate all of the items that were going to be influenced by the notion of being a steward.
 
As it stands now, if we continue on this path, we'll soon have three constants in this template all deriving the same value from the same attribute: the existing canEditDescription, this new canManageUsers, and my upcoming change canSeeWorkspaces.

I'm fine with you keeping it as-is in this PR.  More of just a heads up on my thinking with the code going forward.

src/components/dataset/DatasetAccess.jsx
lu-c: replying to say that i completely forgot to do this, and am working on adding it now 🙏 sorry!

nmalfroy (approved):
Nice job.  Before merging, I'd love to click test.  Do you think you could deploy this branch into your dev instance?  Happy to walk you through that if you'd like

src/components/UserList.jsx
lu-c: Lou responded with:

First accordion starts open, to show the user the behavior possible. The other two remain closed until user interacts.

I've updated the PR to match!

otchet-broad: Are you trying to cover a case that "defaultExpanded" doesn't cover? [See Props](https://mui.com/material-ui/api/accordion/#:~:text=for%20more%20details.-,defaultExpanded,-bool).  If you use that, you can eliminate the need to track the expanded state yourself and just leverage the property passed in.  eg. defaultExpanded={prop.defaultOpen}.

lu-c: wow ha ha, yup, that totally works in the props - pushed up a  fix to clean this up!

otchet-broad (approved):

lu-c:
Yayyyyyyyy

PR #1189, opened 2022-09-26
DR-2729 Fix the main dockerfile

nmalfroy:
I accidentally broke the main docker build.  Update here should get those passing again


okotsopoulos (approved):
Thanks for fixing! Were all of these changes required to fix the build, or were there some drive-bys?

otchet-broad (approved):
I like how codegen is decoupled from the npm build.

nmalfroy:
> Thanks for fixing! Were all of these changes required to fix the build, or were there some drive-bys?

np!  Those were actually the linter in VSCode that replaces `"`'s with `'`'s

PR #1195, opened 2022-10-06
[DR-2744] Jobs Page

lu-c:
https://jade-cl.datarepo-dev.broadinstitute.org/activity

Check out the Activity tab. :)


src/components/HomeView.jsx
```diff
@@ -57,24 +58,32 @@ function HomeView({ classes, location }) {
   const [searchString, setSearchString] = useState('');
   const tabValue = prefixMatcher.exec(location.pathname)[0];
   let tableValue;
+  let pageTitle = 'Terra Data Respository';
```
nmalfroy, line 61: Can we make the page title be Datasets and Snapshots on the respective pages for now?

src/components/table/JobTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+      succeeded: { icon: `fa fa-circle-check ${classes.statusIconSuccess}`, label: 'Completed' },
+      running: { icon: `fa fa-rotate fa-spin ${classes.statusIconSuccess}`, label: 'In Progress' },
+      failed: { icon: `fa fa-circle-xmark ${classes.statusIconFailed}`, label: 'Failed' },
+    };
+
+    const columns: Array<TableColumnType> = [
+      {
+        label: 'Job ID',
+        name: 'id',
+        allowSort: true,
```
nmalfroy: We can't sort on any column other than submitted time (and I'm not sure we even want to allow sorting on that).  Let's set allowSort to false on all of these

src/components/table/JobTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+        width: '25%',
```
nmalfroy: I don't think that the widths add up to 100%.  Also, the ID should probably be 15%

src/components/table/LightTable.tsx
```diff
@@ -289,6 +293,8 @@ function LightTable({
     0,
   );
   const effectiveTableWidth = _.isNaN(tableWidth) || !supportsResize ? '100%' : tableWidth;
+  const showPagination = (rows && rows.length > 0) || infinitePaging;
```
nmalfroy, line 296: Do we want to show it if `rows.length === 0` and `infinitePaging`?  I would think that we'd want to hide in that case

src/reducers/job.ts
```diff
@@ -32,6 +39,19 @@ interface JobAction {
 export default {
   jobs: handleActions(
     {
+      [ActionTypes.GET_JOBS]: (state) =>
+        immutable(state, {
+          loading: { $set: true },
+        }),
+      [ActionTypes.GET_JOBS_FAILURE]: (state) =>
+        immutable(state, {
+          loading: { $set: false },
```
nmalfroy, line 49: should we clear jobs in that case?

nmalfroy (changes requested):
looks great, thanks for this!  I have a couple of (small) things that we probably want to address before merging

src/sagas/repository.ts
```diff
@@ -545,6 +545,29 @@ export function* getDatasetTablePreview({ payload }: any): any {
...
+    const response = yield call(
+      authGet,
+      `/api/repository/v1/jobs?offset=${offset}&limit=${limit}&sort=${sort}&direction=${direction}&filter=${searchString}`,
+    );
+    yield put({
+      type: ActionTypes.GET_JOBS_SUCCESS,
+      jobs: { data: response },
+    });
+  } catch (err) {
+    showNotification(err);
```
otchet-broad: Nice work on adding a generator!  This is nice and clean.

One small tweak to consider: instead of popping a toast with technical details, I've started to include messages to inform the user what to do.  For example, showNotification could pop: 
`An error occurred loading jobs.  Reload the page to try again.`
In cases where these generators are used by multiple components, it might be useful to pass in the error message with the action parameters.
If we want the browser to capture technical details of the error, we could log those to the console.

otchet-broad (changes requested):
Nice work!  Left one note about error handling and something I've recently added.

Would you be able to add some unit tests around the JobView and the JobTable?

src/components/table/LightTable.tsx
lu-c: Yes! I had it originally hiding the pagination if there were no results, and we were infinite paging, but the problem was that if you paged to no results, you couldnt page back. You would be forced to hard refresh the page to continue seeing data

src/components/table/JobTable.tsx
lu-c: Updated the widths to add up to 100%, but left ID larger - at 15%, the ID column on my screen is being truncated, which I'd liek to avoid!

src/reducers/job.ts
lu-c: Clearing!

src/sagas/repository.ts
lu-c: Mm, I like that, the users often dont need to see the detailed error message. I think this sort of error handling should be done by the component that is calling the method in that case, since the solution could change based on how this method is being called.

For example, instead of refreshing the page, maybe the method is being called via button click. This repository should not know how they are being called or what the error handling would need to be - that is the responsibility of the caller.

What do you think of that?

otchet-broad: Exactly where I was going.  This juice is worth the squeeze.  :-)

lu-c: Updated this so you can pass in an error message - I kind of .... went crazy at first, and was modifying the repository to allow for an onError method to be passed in initially. I think it might be the better long term solution, but I feel a little unconfident about it, like, maybe it would be something better to add if we eventually find the need for it? But I was designing something far too complex, like "what if we wanted x to happen on this type of error, but y to happen on this other type of error"

I think ultimately, if getJobs failed, it's because *we did something wrong, not the user, and in this one specific instance, they probably dont need something too robust!

src/components/JobView.tsx
```diff
@@ -0,0 +1,90 @@
...
+      search: string,
+    ) => {
+      dispatch(
+        getJobs({
+          limit,
+          offset,
+          sort,
+          direction: sortDirection,
+          search,
+          errMessage: 'An error occured loading jobs. Please reload the page to try again.',
```
otchet-broad, line 54: Nice!

src/sagas/repository.ts
```diff
@@ -568,6 +568,30 @@ export function* getDatasetTablePreview({ payload }: any): any {
...
+      authGet,
+      `/api/repository/v1/jobs?offset=${offset}&limit=${limit}&sort=${sort}&direction=${direction}&filter=${searchString}`,
+    );
+    yield put({
+      type: ActionTypes.GET_JOBS_SUCCESS,
+      jobs: { data: response },
+    });
+  } catch (err) {
+    const { errMessage } = payload;
+    showNotification(errMessage || err);
```
otchet-broad, line 588: Huzzah!  I like it!  Thanks for considering the approach.

nmalfroy (approved):
This looks great!

otchet-broad (approved):
Nice work!

PR #1197, opened 2022-10-11
[DR-2750] Add support for markdown display and editing in the UI.

otchet-broad:
BEFORE:
Screen Shot 2022-10-11 at 12 15 30 PM

AFTER:
Screen Shot 2022-10-11 at 12 19 29 PM

Leverages the same editor and markdown features used by Terra-UI, but renders using a rendering engine that will not render embedded HTML without additional configuration so that this application fails closed.  Should a user attempt to embed HTML in markdown, the contents will be rendered as plain text.

Note: The editor renderer uses a custom React component (WithoutStyles....) to display the content in preview mode because the mechanism to deliver the content does not support using withStyles.  When markdown is rendered outside of the editor, style can be applied.  In either case, links are set to open in new tabs.

Editor preview (ipsum link is unstyled):
Screen Shot 2022-10-11 at 12 26 58 PM

Markdown renderer (ipsum link is styled):
Screen Shot 2022-10-11 at 12 26 32 PM

Click around here: [https://jade-eo.datarepo-dev.broadinstitute.org/datasets](https://jade-eo.datarepo-dev.broadinstitute.org/datasets)


nmalfroy (approved):
Wow, this looks great!  Thanks for all of the extra tests too!

src/modules/theme.ts
```diff
@@ -201,6 +201,9 @@ const theme = createTheme({
             backgroundColor: LINK_HOVER,
           },
         },
+        outlined: {
+          backgroundColor: WHITE,
```
otchet-broad, line 205: Adjusted after a discussion with Lou regarding contrast.

lu-c (approved):
the demos of this show it works great! :)

PR #1198, opened 2022-10-11
[DR-2777] Second pr adding the status modal on errored rows

lu-c:
Added a modal on the Activity page - Clickin gon the job ID will now open this modal that will display extra content

ss


src/components/table/JobTable.tsx
```diff
@@ -0,0 +1,112 @@
...
+        name: 'job_status',
+        allowSort: true,
+        width: '10%',
+        render: (row: any) => (
+          <div>
+            <span className={classes.statusContainer}>
+              <i className={clsx(classes.statusIcon, statusMap[row.job_status].icon)} />
+              {statusMap[row.job_status].label}
+            </span>
+            {row.job_status === 'failed' && <JobStatusModal id={row.id} />}
```
nmalfroy: Could you make sure that the button stays on the same line so that the table row heights stay consistent?  This is an interesting pattern for embedding dialog. I like it!

src/reducers/job.ts
```diff
@@ -32,13 +53,38 @@ interface JobAction {
...
         }),
       [ActionTypes.GET_JOB_RESULT_SUCCESS]: (state, action: JobAction) =>
         immutable(state, {
+          jobResult: { $set: action.payload.data.jobResult },
           jobResultObjectId: { $set: action.payload.data.id },
+          jobResultLoading: { $set: false },
+        }),
+      [ActionTypes.GET_JOB_RESULT_FAILURE]: (state, action: JobAction) =>
+        immutable(state, {
+          jobResult: { $set: action.payload.data.jobResult },
```
otchet-broad: Do you want to clear jobResult here?

src/sagas/repository.ts
```diff
@@ -545,6 +545,89 @@ export function* getDatasetTablePreview({ payload }: any): any {
...
+                resultType: 'success',
+                result: resultResponse.data,
+              },
+            },
+          },
+        });
+      } catch (err: any) {
+        if (err.response) {
+          yield put({
+            type: ActionTypes.GET_JOB_RESULT_FAILURE,
```
otchet-broad: In this case, I think you might want to consider mapping this to GET_JOB_RESULT_SUCCESS.  If I understand correctly, you're dealing with HTTP response codes that would have typically tripped our pattern.  The request was successfully accepted by the server and the job status was successfully returned.  It just happened to be returned in a non-2XX response.  If the response payload doesn't contain any of the expected "error" details, I might then emit an ActionTypes.GET_JOB_RESULT_FAILURE.

src/components/table/JobTable.tsx
lu-c: I got rid of this, sorry!

src/reducers/job.ts
lu-c: ok!

src/sagas/repository.ts
lu-c: oh gotcha, updatd this, i think

src/reducers/job.ts
```diff
@@ -12,6 +12,17 @@ export interface JobState {
   jobs: Array<JobModel>;
   refreshCnt: number;
   loading: boolean;
+  jobResult?: JobResult;
+  jobResultLoading: boolean;
+}
+
+export interface JobResult {
+  resultType?: 'success' | 'error';
```
snf2ye: Could we pull the values for result type into an enum in index.ts? Then we can also use the enum when returning the job result in [repository.ts](https://github.com/DataBiosphere/jade-data-repo-ui/pull/1198/files#diff-b0cc26de535914428b254ecf0a5ed0e23a1c0e9753bbcb59a2af7167f6cd37ceR613)

src/components/job/JobResultModal.tsx
```diff
@@ -0,0 +1,197 @@
...
+        >
+          <DialogTitle id="see-more-dialog-title">
+            <div className={classes.dialogTitle} style={{ float: 'left' }}>
+              Job Details
+            </div>
+            <IconButton size="small" style={{ float: 'right' }} onClick={handleSeeMoreClose}>
+              <Close />
+            </IconButton>
+          </DialogTitle>
+          <DialogContent>
```
snf2ye: It looks like we have a similar dialog component in [LightTable](https://github.com/DataBiosphere/jade-data-repo-ui/blob/develop/src/components/table/LightTable.tsx#L395). Would it make sense to pull the dialog container plus close control into a reusable component? I'm thinking that could help keep styling/functionality uniform across the UI, but it might be more work than worth it.

snf2ye (commented):
I'm so excited about this jobs page!! Great job!

snf2ye (approved):

src/sagas/repository.ts
```diff
@@ -592,6 +592,66 @@ export function* getJobs({ payload }: any): any {
...
+              id,
+              jobResult: {
+                resultType: 'success',
+                result: resultResponse.data,
+              },
+            },
+          },
+        });
+      } catch (err: any) {
+        if (err.response) {
```
otchet-broad, line 621: @lu-c - Do you think we should make a ticket to clean this up in the service?  It seems like we've conflated job status and http status of a request for job status - and this is the byproduct.

otchet-broad (approved):
Nice work!  This will be very helpful.

src/components/job/JobResultModal.tsx
lu-c: I dont know! The advantage to styling / functionality would be great, but if we wanted to customize a title to have a warning icon / message, or a "settings" icon, it could faster to edit component.

I feel like the dialog component already contains a lot of the similar styling, just coming from material ui itself already - The only classes that I think would be shared are `root` and `dialogTitle`, and I can drop them to make it just use the default instead?

src/reducers/job.ts
lu-c: Updated! :)

src/sagas/repository.ts
lu-c: Sure, let me add a ticket! 

Edit - added here: https://broadworkbench.atlassian.net/browse/DR-2790

src/components/job/JobResultModal.tsx
snf2ye: True, keeping it separate sounds like it makes sense then. Thanks for entertaining the idea!

PR #1201, opened 2022-10-14
DR-2577 Improve UX for adding/removing members of Policy Groups.

otchet-broad:
BEFORE:
Screen Shot 2022-10-17 at 9 21 06 AM

AFTER:
Screen Shot 2022-10-17 at 9 20 34 AM

Also:
Driveby removal of dead styling on loading mask in Manage Snapshot Worskapce;
ManageUserModal, UserList, DatasetAccess conversion to typescript
Removed dead Custodian specific code

Deployed here for your consideration: [https://jade-eo.datarepo-dev.broadinstitute.org/datasets](https://jade-eo.datarepo-dev.broadinstitute.org/datasets)


src/actions/app.js
```diff
@@ -140,20 +140,6 @@ export const { getDatasetTablePreview } = createActions({
   [ActionTypes.GET_DATASET_TABLE_PREVIEW_SUCCESS]: (_) => _,
 });
 
-export const { addCustodianToDataset } = createActions({
```
otchet-broad, line 143: Dead code.

src/components/ManageUsersModal.tsx
```diff
@@ -0,0 +1,181 @@
+import React from 'react';
```
otchet-broad, line 1: Converted to TypeScript

src/components/snapshot/overview/ManageWorkspacesModal.tsx
```diff
@@ -73,20 +73,18 @@ const styles = (theme: CustomTheme) =>
       bottom: 0,
       left: 0,
       width: 'initial',
-      height: 'fit-contnet',
```
otchet-broad, line 76: Glad this had a typo!  height isn't needed, so it's removed.

src/components/snapshot/overview/ManageWorkspacesModal.tsx
```diff
@@ -73,20 +73,18 @@ const styles = (theme: CustomTheme) =>
...
       overflow: 'clip',
       backgroundColor: theme.palette.common.white,
       zIndex: 100,
     },
   });
 
-type ManageWorkspaceModalProps = {
-  classes: ClassNameMap;
+interface ManageWorkspaceModalProps extends WithStyles<typeof styles> {
```
otchet-broad, line 82: Made this consistent with other interface props.

src/components/dataset/DatasetAccess.tsx
```diff
@@ -1,37 +1,49 @@
 import React from 'react';
```
otchet-broad, line 1: Converted to TypeScript.

src/components/UserList.tsx
```diff
@@ -1,47 +1,47 @@
 import React from 'react';
```
otchet-broad, line 1: Converted to TypeScript.

src/sagas/repository.ts
```diff
@@ -514,40 +523,9 @@ export function* removeDatasetPolicyMember({ payload }: any): any {
     });
   } catch (err) {
     showNotification(err);
-  }
-}
-
-export function* addCustodianToDataset({ payload }: any): any {
```
otchet-broad, line 520: This add/remove custodian work seems to be dead.  In user management, it was replaced by the ADD_SNAPSHOT/DATASET_POLICY_MEMBER work.

nmalfroy: Yeah, I like having the more general purpose method

nmalfroy (approved):
Great work on this!  Looks good to me

lstrano-broad:
Wondering if we should have a 3 second delay to avoid a super quick transition

okotsopoulos (approved):
Thanks for adding this improvement!  Not a blocker to merge, but summarizing a piece of feedback we discussed in standup:

When our TDR API returns its response quickly, the effect of the overlay / spinner / text is a bit strobe-like, especially the text.  We agreed that only showing the overlay after <x> time without a response could suffer the same issue.  Further improvements could be undertaken at a later point.

otchet-broad:
@lstrano-broad - I'm not sure I would want to hold up the user for that length of time if it weren't necessary.  I think a further improvement might be to localize work indicators on actions (delete, add) instead of masking the dialog.  To do this is a bit more complex.  Most importantly, we'll need to track the individual transactions and update the discrete UI components as work on them begins and completes.  

For an analogy, consider the icon that appears in a Chrome tab when that tab is navigating.  It doesn't block your ability to do work in a different tab.  Here I'd suggest doing something similar where the (X) to delete becomes a spinner instead of the entire dialog being covered by a mask.

This would allow the user to click as fast as they wanted with minimal pixel change and masking.

Tracking state at this level is where the real work is at.

Should I make a change before merging?

otchet-broad:
@okotsopoulos - If the concern is mostly with the text that appears, do you want me to try a build that turns the text off for that specific spinner?  That's simple to do and try.

otchet-broad:
@lstrano-broad agreed with the change to turn off the text below the spinner.  He's also following up with tickets for:

- more granular spinners and accounting for multiple actions on the same modals;
- box around the description text;
- close button color on modals;

PR #1203, opened 2022-10-18
DR-2786: Address Dependabot PR Test run failures

snf2ye:
Since at least June 2022, PRs created by dependabot have had E2E test failures due to missing access to github secrets. This is the result of a change by github to restrict access to secrets for workflows triggered by PRs created from an external fork. More details [here](https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852541544) and [here](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/).  

My understanding is that the security concern addressed with Github's change is to prevent unknown third parties taking advantage of automatic processing of PRs.  In our case for E2E tests, "automatic processing" == checkout and build the branch presented by the external fork and run our tests against them.  The suggested "two workflow" approach suggested in the above resource mitigates this risk by specifically allowing only dependabot PRs to be automatically processed. 

How did Github make this restriction change? They made it so that if a workflow is triggered via a `pull_request` from an external fork, the workflow has read-only access to repo and does not have access to secrets. To go hand and hand with this change, they created a new trigger for workflow runs:
- `pull_request_target`: This has access to repo secrets BUT it only runs against the develop branch, NOT the PR's branch/current changes. 

Given that we want to run tests against the branch code, this does not work for us. The resource linked above suggests using the following trigger:
`workflow_run`: Triggered by another workflow run and can be filtered down to only successful/completed runs. 

The "Two Workflow" approach
-  Workflow 1 - Dependabot PR Check - check that the PR is created by dependabot
- Workflow 2 - E2E tests - Run E2E test via `workflow_run` trigger rather than `pull_request` trigger so that the run has access to the necessary secrets. 

Updates:
10/19: Struggling a bit - going in circles trying to understand github documentation + actions code not working as expected. I've decided my next step will be to outline my attempted and potential approaches in [this doc.](https://docs.google.com/document/d/1ppeltuizhh9ZVl8z1FU4a3FAoHMJYeTYZlqaIaKhing/edit#heading=h.kpn4t4invgm7)


snf2ye:
Closing this PR because I've landed on the original solution -- I just need to add duplicate secrets in the "dependabot secrets" list for jade-data-repo-ui. I came across [this quote from the github documentation](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#responding-to-events):

> If you have a workflow that will be triggered by Dependabot and also by other actors, the simplest solution is to store the token with the permissions required in an action and in a Dependabot secret with identical names.

This exactly matches our use case. It seems to be the most secure and simplest solution.

PR #1221, opened 2022-10-27
DR-2803: Fix filtering on null values during snapshot create

snf2ye:
The bug: The query builder was attempting to check for the string "null", so no results are found. 
```
# BROKEN - return 0 rows
SELECT * FROM `<table>` WHERE <column> IN ("null")
 LIMIT 100

#WORKS - returns 100 rows
SELECT * FROM `<table>` WHERE  <column> IS NULL
 LIMIT 100
```

Before the change --

https://user-images.githubusercontent.com/13254229/198398751-4a2b33ea-96d8-47fc-bd51-05bc64fc77cd.mov

With code changes -- 

https://user-images.githubusercontent.com/13254229/198399413-819f6312-d0d8-4a32-b26f-a8e515b821db.mov

### DIFF
--- a/cypress/integration/queryBuilder.spec.js
+++ b/cypress/integration/queryBuilder.spec.js
@@ -132,6 +132,44 @@ describe('test query builder', () => {
     });
   });
 
+  describe('filtering on null checkbox value', () => {
+    it('filters on null', () => {
+      // Switch to variant table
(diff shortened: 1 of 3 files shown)


nmalfroy (approved):
looks good!  thanks for deploying too :)

otchet-broad (approved):
Nice work!

PR #1222, opened 2022-10-28
[DR-2774] Update to accesses in requests

lu-c:
When you create a snapshot from a dataset, you can now add users as all of the policy types. Previously, only reader was available. The endpoint was recently updated, so this PR is intended to make changes to allow for it all now.

1. Updated view to allow for permissions of all three types.
2. Updated membership access management to a single component that is shared on dataset/snapshot creation and for roles/management tab (when updating an existing snapshot/dataset)
3. Invision Mock Up: https://dspuxteam.invisionapp.com/console/share/N5TVJ642HMY

recording_2022-10-31_at_13 05 59


snf2ye (approved):
Thanks for answering a few questions via slack! 
Looks great - thank you!

nmalfroy (approved):
looks great!  I think you have a conflict with your package file but this looks great!

lstrano-broad:
Looks great -- question on the button placement of "Create Snapshot" looks like it's floating above Discoverers card? It should be anchored like the other buttons.

lu-c:
> Looks great -- question on the button placement of "Create Snapshot" looks like it's floating above Discoverers card? It should be anchored like the other buttons.

Yes, it's floating! That was how that panel was before editing it, so I didnt think to change it. Can I it real quick before merging this!

PR #1225, opened 2022-11-01
[DR-2479] Added ability to filter datasets with multiple keys, separated by newline characters

lu-c:
BEFORE:
image

AFTER:
Screen Shot 2022-11-01 at 10 41 45 AM


nmalfroy (approved):
I just deployed this on my instance and it's a heck of an improvement!

snf2ye (approved):

PR #1227, opened 2022-11-02
DR-2816: Bump up checkbox threshold

snf2ye:
**Note:** Changes deployed to my dev env: https://jade-sh.datarepo-dev.broadinstitute.org/datasets/c5248002-05d6-469e-bf31-38f1792198f3/data

### Background
There are two modes for filtering string values on snapshot create in the UI.  Currently, when there are more than 10 distinct values for a column, we switch from a "checkbox" over to a "dropdown list". 

1. Checkboxes
image

2. Dropdown list
image

### The Change/User Request
The user experience is not the best for the dropdown list. So, they want to see the checkbox option more often. We can do this by increasing the threshold for which we switch over the dropdown list. Bumping this up to 30 seems like a reasonable change.

### DIFF
--- a/src/components/dataset/data/sidebar/filter/CategoryWrapper.jsx
+++ b/src/components/dataset/data/sidebar/filter/CategoryWrapper.jsx
@@ -5,7 +5,7 @@ import PropTypes from 'prop-types';
 import FreetextFilter from './FreetextFilter';
 import CategoryFilterGroup from './CategoryFilterGroup';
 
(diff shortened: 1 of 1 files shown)


okotsopoulos (approved):

okotsopoulos:
Thanks for the thorough PR description despite it being essentially a one-line config change :)

nmalfroy (approved):
thanks for fixing!

PR #1230, opened 2022-11-07
DR-2828 - Fix Total Row count for data tables

snf2ye:
Changes deployed here: https://jade-sh.datarepo-dev.broadinstitute.org/

### The Bug
Total rows reported on the dataset data view was incorrectly capped at 1000. 

Example - total rows says 1000:
image

However, we know that there are more than 1000 rows in this table.

**Why was this happening?**
Combo of two things:
- Total rows set by the count of rows returned for a given query.
- Incorrectly having a "limit 1000" clause where we should instead remove the limit

**This is fixed by a combo of two things:**
- Instead of setting total rows as a count of the rows returned by the query, get a count of rows that meet the filtering requirements
- Remove the "limit 1000" clause 
image

### DIFF
--- a/src/components/common/data/DataView.tsx
+++ b/src/components/common/data/DataView.tsx
@@ -198,7 +198,7 @@ function mapStateToProps(state: TdrState) {
     polling: state.query.polling,
     rows: state.query.rows,
     rowsPerPage: state.query.rowsPerPage,
-    totalRows: state.query.queryParams.totalRows,
+    totalRows: state.query.resultsCount,
     refreshCnt: state.query.refreshCnt,
   };
 }
--- a/src/components/dataset/data/DatasetDataView.tsx
+++ b/src/components/dataset/data/DatasetDataView.tsx
@@ -30,8 +30,6 @@ import {
   ResourceType,
 } from '../../../constants';
 
-const QUERY_LIMIT = 1000;
-
 type IProps = {
   dataset: DatasetModel;
   dispatch: Dispatch<Action>;
@@ -39,6 +37,7 @@ type IProps = {
   joinStatement: string;
   orderDirection: OrderDirectionOptions;
   orderProperty: string;
+  rowsPerPage: number;
   profile: BillingProfileModel;
   queryParams: QueryParams;
   snapshotRequest: SnapshotRequest;
@@ -52,6 +51,7 @@ function DatasetDataView({
   match,
   orderDirection,
   orderProperty,
+  rowsPerPage,
   profile,
   queryParams,
   snapshotRequest,
@@ -145,7 +145,7 @@ function DatasetDataView({
           SELECT ${DbColumns.ROW_ID},
             ${selectedTable?.columns?.map((column) => column.name).join(', ')} ${fromClause}
             ${orderProperty ? `ORDER BY ${orderProperty} ${orderDirection}` : ''}
-          LIMIT ${QUERY_LIMIT}`,
+          LIMIT ${rowsPerPage}`,
         ),
       );
       dispatch(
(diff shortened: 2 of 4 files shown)


src/components/common/data/DataView.tsx
```diff
@@ -198,7 +198,7 @@ function mapStateToProps(state: TdrState) {
     polling: state.query.polling,
     rows: state.query.rows,
     rowsPerPage: state.query.rowsPerPage,
-    totalRows: state.query.queryParams.totalRows,
+    totalRows: state.query.resultsCount,
```
snf2ye, line 201: Change 1:  This fixes the issue reported by the user.
We use this shared component for both the dataset and snapshot data views. These differ in that dataset still pull the data directly from BQ, while the snapshot pulls data from our preview endpoint. 
- For Dataset/Direct from BQ - We run a [query for datasets](https://github.com/DataBiosphere/jade-data-repo-ui/blob/develop/src/components/dataset/data/DatasetDataView.tsx#L155) to get the count of rows that meet filtering specifications. We should use that value here. 
- For Snapshots - We can map the row count returned from the preview endpoint to this state variable. (see change in query.ts)

src/components/dataset/data/DatasetDataView.tsx
```diff
@@ -145,7 +145,7 @@ function DatasetDataView({
           SELECT ${DbColumns.ROW_ID},
             ${selectedTable?.columns?.map((column) => column.name).join(', ')} ${fromClause}
             ${orderProperty ? `ORDER BY ${orderProperty} ${orderDirection}` : ''}
-          LIMIT ${QUERY_LIMIT}`,
+          LIMIT ${rowsPerPage}`,
```
snf2ye: Change 2 - We were always querying for 1000 rows, even though we only show up to 250 rows on a given page. [This was why the user was seeing 1000 rows as the total number of rows when there was greater than 1000 rows]

src/sagas/repository.ts
```diff
@@ -540,23 +540,6 @@ export function* removeDatasetPolicyMember({ payload }: any): any {
...
-    yield put({
-      type: ActionTypes.GET_DATASET_TABLE_PREVIEW_SUCCESS,
-      preview: response,
-      tableName,
-    });
-  } catch (err) {
-    showNotification(err);
-  }
-}
-
```
snf2ye, line 558: Change 3 - Drive by change - as far as I can tell, we're not using this anywhere.

snf2ye (commented):
See comments for explanation of each change

src/components/dataset/data/DatasetDataView.tsx
```diff
@@ -144,8 +142,7 @@ function DatasetDataView({
           `#standardSQL
           SELECT ${DbColumns.ROW_ID},
             ${selectedTable?.columns?.map((column) => column.name).join(', ')} ${fromClause}
-            ${orderProperty ? `ORDER BY ${orderProperty} ${orderDirection}` : ''}
-          LIMIT ${QUERY_LIMIT}`,
+            ${orderProperty ? `ORDER BY ${orderProperty} ${orderDirection}` : ''}`,
```
snf2ye, line 145: Change 2 - Remove query limit altogether -- With the limit in place, we were not able to page past that row # in the table.

src/components/dataset/data/DatasetDataView.tsx
snf2ye: EDIT: This change impacted paging - we actually need to remove the limit altogether.

lu-c (approved):
This looks good! Thanks for the explanations in the comments :)

nmalfroy (approved):
Looks good to me!

PR #1237, opened 2022-11-09
[DR-2809][DR-2810] Added dataset activity tab

lu-c:
- Added dataset activity tab, should only be visible to stewards and custodians
- Noticed that failures don't get added to the journal entries, so the status column only reports successes

Screen Shot 2022-11-09 at 11 50 23 AM


src/components/JournalEntriesView.tsx
```diff
@@ -0,0 +1,56 @@
...
+    dispatch(getJournalEntries({ id, resourceType, limit, offset }));
+  };
+
+  return (
+    <JournalEntryTable
+      journalEntries={journalEntries}
+      handleFilterJournalEntries={handleFilterJournalEntries}
+      loading={loading}
+      refreshCnt={refreshCnt}
+    />
```
snf2ye, line 36: Why is JournalEntryTable in a separate component if it is all that is returned by JournalEntriesView?

src/components/JournalEntriesView.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { JournalEntryModel, IamResourceTypeEnum } from 'generated/tdr';
+import { getJournalEntries } from 'actions/index';
+import JournalEntryTable from './table/JournalEntryTable';
+
+const styles = () =>
+  createStyles({
+    className: {
+      color: 'red',
+    },
+  } as const);
```
snf2ye: It's my understanding that this pattern is now [considered legacy](https://mui.com/system/styles/basics/) and we're going to need to migrate away from it in order to update material UI. I think we'll need to use the [mui system components](https://mui.com/system/getting-started/overview/) and styled-components moving forward.

package-lock.json
```diff
@@ -1,12 +1,12 @@
...
   "name": "jade-data-repo-ui",
-  "version": "0.176.0",
+  "version": "0.181.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "jade-data-repo-ui",
-      "version": "0.176.0",
+      "version": "0.181.0",
```
snf2ye: If you get the latest code from develop, we shouldn't need this change, right? 
I'm just worried there might be some conflicts when merging this change since we automatically bump the version.

src/components/JournalEntriesView.tsx
lu-c: That's a good question! I was following the design that we did for DatasetView -> DatasetTable, to try to keep it the same. The wrapper (JournalEntriesView) extracts the state variables and passes them in to the JournalEntryTable. Do you think we should merge them?

src/components/JournalEntriesView.tsx
lu-c: Ok, I'll take a look at these documents! In the meanwhile, this can be deleted, since they arent being used

package-lock.json
lu-c: Ok, let me revert this back!

src/components/snapshot/overview/SnapshotOverviewPanel.tsx
```diff
@@ -204,7 +215,16 @@ function SnapshotOverviewPanel(props: SnapshotOverviewPanelProps) {
           </Grid>
         )}
       </TabPanel>
-      <TabPanel value={value} index={2}>
+      {canViewJournalEntries && (
+        <TabPanel value={value} index={2}>
+          <Grid container spacing={2}>
+            <Grid item xs={12}>
+              <JournalEntriesView id={snapshot.id} resourceType="DATASNAPSHOT" />
```
nmalfroy: We should probably have a constant for DATASNAPSHOT (and DATASET) in constants/index.ts

nmalfroy (commented):
Looks good!  Jut would love to get that field into a constant and any chance that you could put this up on a deployment?

src/components/JournalEntriesView.tsx
snf2ye: perfect!

src/components/snapshot/overview/SnapshotOverviewPanel.tsx
lu-c: Oh, I found one for this in the generated tdr values! :) Updated this code with it!

lu-c:
> Looks good! Jut would love to get that field into a constant and any chance that you could put this up on a deployment?

Got it! I ran a quick deploy, but it's now 404'ing on the journal method not existing

nmalfroy:
Oh! you probably need to update your API pod to the latest value.  I forget the quickest way to do that...maybe a `helmfile apply` and _then_ a skaffold deploy

lu-c:
Oh! Good call, updated with helmfile!

This link now works!
https://jade-cl.datarepo-dev.broadinstitute.org/datasets/04059c65-eab0-4c3b-a6ac-b6ee1583157e

src/components/table/JournalEntryTable.tsx
```diff
@@ -0,0 +1,121 @@
...
+        columns={columns}
+        handleEnumeration={handleFilterJournalEntries}
+        noRowsMessage="No journal entries found"
+        infinitePaging={true}
+        filteredCount={Number.MAX_SAFE_INTEGER}
+        rows={journalEntries}
+        searchString=""
+        loading={loading}
+        refreshCnt={refreshCnt}
+        rowKey="id"
```
snf2ye: I'm seeing some wonkiness with the paging of the journal table. I think we usually set the[ totalCount field](https://github.com/DataBiosphere/jade-data-repo-ui/blob/develop/src/components/table/DatasetTable.tsx#L112).
https://user-images.githubusercontent.com/13254229/201801625-5c81be25-b1a9-4cbd-91be-ff322f2ae45a.mov

snf2ye: Oh, it looks like that is happening with the infinitePaging option. Is this behaving as expected?

src/components/table/JournalEntryTable.tsx
```diff
@@ -0,0 +1,121 @@
...
+    ];
+    return (
+      <LightTable
+        columns={columns}
+        handleEnumeration={handleFilterJournalEntries}
+        noRowsMessage="No journal entries found"
+        infinitePaging={true}
+        filteredCount={Number.MAX_SAFE_INTEGER}
+        rows={journalEntries}
+        searchString=""
```
snf2ye: It doesn't look like there is search set up on this table, so I don't think we need to set searchString or filtered count, right?

src/components/table/JournalEntryTable.tsx
```diff
@@ -0,0 +1,121 @@
...
+        loading={loading}
+        refreshCnt={refreshCnt}
```
snf2ye, line 142: As far as I can see in the code, refreshCnt isn't used.

src/components/JournalEntriesView.tsx
snf2ye: Gotcha! That's fine to leave it as is. Thanks for explaining!

src/components/table/JournalEntryTable.tsx
nmalfroy: that gets used to reload the page (so for instance, if we want to add a refresh button)

nmalfroy:
Couple of things I noticed:
- The API should return the values in descending order
- I wonder if we could parse the mutation to at least show what what field was changed...maybe replace the `UPDATE` value with what was changed

src/components/table/JournalEntryTable.tsx
lu-c: It's required on the LightTable component, but I can update it so it's optional and remove it here

src/components/table/JournalEntryTable.tsx
lu-c: Unfortunately, that is expected behavior - It's because these requests are not returning any paging information whatsoever, so it's impossible to determine from the ui when to disable paging.

I think this weird behavior was less visible on the jobs page because there are a ton of jobs! One change we could make is to disable the next button if there are no results in the current view (or < pageSize results)... Let me try to push that into this PR

lu-c:
> * The API should return the values in descending order

This should be a change made in the data-repo proj!

> * I wonder if we could parse the mutation to at least show what what field was changed...maybe replace the `UPDATE` value with what was changed

OOooohh, let me try this!

src/components/table/JournalEntryTable.tsx
snf2ye: Oh, ok, so it's in place for future use?

src/components/table/JournalEntryTable.tsx
snf2ye: Ah, true. Yeah, if you could add that, I think that would be great.

src/components/table/LightTable.tsx
```diff
@@ -379,11 +382,14 @@ function LightTable({
                 disableTouchRipple: true,
                 disableFocusRipple: true,
                 disableRipple: true,
+                disabled: infinitePaging
+                  ? rows.length < rowsPerPage
+                  : page * rowsPerPage + rows.length >= filteredCount,
```
snf2ye, line 387: Since the paging works for the case we have infinitePaging set to false, what would you think about just setting this for the case that infinitePaging is enabled? 
```suggestion
                disabled: infinitePaging && rows.length < rowsPerPage
```
Otherwise, I think there is something off with this argument: `page * rowsPerPage + rows.length >= filteredCount`. Shouldn't the inequality sign be flipped?

lu-c: this is for "disabled", so if there are more results than there are results, then it should be disabled. I think i could change it to == instead of >=?

lu-c: I cant set the disabled for just infinite paging, i think? And if the attribute is set, I need to have it update for !infinitepaging situations

src/components/table/JournalEntryTable.tsx
lu-c: Yup! :D

src/components/table/LightTable.tsx
snf2ye: > I cant set the disabled for just infinite paging, i think? And if the attribute is set, I need to have it update for !infinitepaging situations

Ah, you're totally right. 

> this is for "disabled", so if there are more results than there are results, then it should be disabled. I think i could change it to == instead of >=?

If there are more results than we can display on the page then I think we want disabled=false so that it will continue to page, right? Double negatives are fun haha

lu-c: Yeah, haha, Im sorry about that! XD

it's saying `disable if (page * rowsPerPage + rows.length >= filteredCount)`, which should mean there are no more results! I think it's right, from clicking around in the app!

snf2ye: Thanks for chatting with me about this offline!

Here's my takeaway from this conversation:
I was hung up on the two different conditions (especially since the equality was flipped), but I think I understand now--
-> When infinitePaging=true, we don't know the total count, so we work with what we have, which works for every case EXCEPT number of rows == rows per page
-> When infinitePagine=false, we have the totalRow count and therefore can handle the edge case where the number of rows == rows per page, but we've hit the end of the results.

snf2ye (approved):

nmalfroy (approved):

PR #1238, opened 2022-11-10
Making a second develop to test the build errors

lu-c:


lu-c:
I think I found the problem?!

https://github.com/OpenAPITools/openapi-generator/pull/13964/files

Openapi merged a PR the same time our develop branch broke! It modifies the mustache files for the exact files that are broken, but uh, looking at the code, it should've helped make it WORK, not break it. I feel kind of stuck here, where/how are we using openapi? (I see it running on npm start, but I don't see it as a dependency in the package.json file)

PR #1239, opened 2022-11-10
[DR-2809] second pr just to test if the merge from develop is what caused the break in the other pr

lu-c:
If this passes, then the breaking code change needs to be identified and fixed in develop.
If this fails, then the breaking code change is outside of those code changes, and might be an external dependency.


lu-c:
Ooops, tested the wrong commit! Opening a new one!

PR #1242, opened 2022-11-14
DR-2834 Pin the version of the code generation library 

nmalfroy:
...to the last known working version.

Typescript doesn't give us any control over this (and others have already complained about this).  We can revisit eventually but for now, we should be good with a hard coded version


okotsopoulos (approved):
Re: "others have already complained about this", do you have any links to share?  Could help us track whether we can ultimately remove this pin. Thanks!

nmalfroy:
It's kinda all over typescript github issues.  They (the typescript team keep pushing back saying: just get the code generators to not break stuff...so that's fun). In any case, I put up a PR against the code generator to fix this:
https://github.com/OpenAPITools/openapi-generator/pull/14011

snf2ye (approved):

PR #1246, opened 2022-11-16
[DR-2836] Added the first step to the dataset creation page

lu-c:
- Added root route for the schema builder page.
- Added component with styling for the schema builder
- Updated the MUI tab theme so it is universally updated, and we dont need to inject the styles every time
- Updated components that share the tab theme
- Added feature flag ability: this flag is stored in localstorage, so the feature will be completely hidden from users unless they mess around in the developer tools (unlikely). But we will have to manually turn it on! To use, open the developer tools, and use this `localStorage.setItem('flags', '{"datasetSchemaCreation":{"active":true}}')`. Then refresh the page

The button is now in the home view
Screen Shot 2022-11-16 at 4 04 50 PM

And when you click on it, the basic page is up. There are follow on tickets to cover adding the content on these panels
Screen Shot 2022-11-16 at 4 04 56 PM


snf2ye (approved):

src/components/HomeView.jsx
```diff
@@ -60,10 +72,29 @@ function HomeView({ classes, location }) {
...
+  let pageHeader = <div className={classes.title}>{pageTitle}</div>;
 
   if (tabValue === '/datasets') {
     pageTitle = 'Datasets';
     tableValue = <DatasetView searchString={searchString} />;
+    pageHeader = (
+      <div className={classes.title}>
+        <div>{pageTitle}</div>
+        <Link to="datasets/new" data-cy="create-dataset-link">
+          {Features.isEnabled('datasetSchemaCreation') && (
```
okotsopoulos, line 84: I *love* the idea of using a feature flag to hide/show new UI components as they're being developed, especially for longer-running work like dataset creation.  In this case (and probably most cases in the UI), the fact that the user could muscle their way into viewing this by probing our code and Developer Tools doesn't really pose a risk, as our hiding it while under construction is more of a courtesy.  I wonder how we'd approach feature flags if more strictness is required.

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
```diff
@@ -0,0 +1,175 @@
...
+            <Typography variant="h4" className={classes.pageTitle}>
+              Have questions?
+            </Typography>
+            Learn more about ingesting data into the Terra Data Repo:
+            <ul className={classes.helpList}>
+              <li>
+                <a
+                  className={clsx(classes.jadeLink, classes.helpListLink)}
+                  href="#"
+                  target="_blank"
```
okotsopoulos, line 142: At what point will we specify these links?

okotsopoulos (approved):
Looks good to me, especially with this being a checkpoint PR on the way to content being added to our steps.

I do want to make sure those _blank links don't fall through the cracks, though!

src/components/HomeView.jsx
lu-c: Yeah, in this case, it doesnt actually do too much if they find their way here, but if the feature flag was protecting something that would write to the database, it could lead to some weird situations if they were pushing up malformed / incomplete data?

Possibly, we could add an `isAdmin` check too? At least the POST methods and all would still be protected by the user login information.

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
lu-c: Whenever we figure out what the links are! I was hoping someone would just know if they already exist - i will ask Lou if he knows!

PR #1249, opened 2022-11-23
[DR-2837] Added form components and validation for step 1 of dataset schema creation

lu-c:
Before
Screen Shot 2022-11-16 at 4 04 56 PM

After
Screen Shot 2022-11-23 at 11 36 53 AM

Branch is currently deployed:
https://jade-cl.datarepo-dev.broadinstitute.org/datasets

To use, open the developer tools, and use this `localStorage.setItem('flags', '{"datasetSchemaCreation":{"active":true}}')`. Then refresh the page. The "create dataset" button should now be available on the dataset page


src/components/dataset/overview/DatasetOverviewPanel.jsx
```diff
@@ -146,16 +146,6 @@ function DatasetOverviewPanel(props) {
             {...a11yProps(3)}
           />
         )}
-        {canViewJournalEntries && (
```
lu-c: Not sure if this was a bad delete, but I think this double was the result of a bad merge?

okotsopoulos: Nice catch! I had noticed this in dev and meant to file a ticket, thanks for fixing.  A bad merge seems plausible.

snf2ye: This was fixed in a [recent PR](https://github.com/DataBiosphere/jade-data-repo-ui/pull/1257/files), right?

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
```diff
@@ -75,93 +90,139 @@ interface TabConfig {
...
-      content: <div>Step 1!</div>,
+      content: <DatasetSchemaInformationView />,
     },
     {
       description: 'Build a schema and create dataset',
       content: <div>Step 2!</div>,
     },
   ];
 
+  const onSubmit = (data: any) => {
```
s-rubenstein, line 126: Is there a reason to have this function before it does anything? It looks like it is feature flagged, but my general rule tends to be waiting to add the method until implementation time to avoid todo code ending in production.

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
```diff
@@ -75,93 +90,139 @@ interface TabConfig {
...
-          Before you can ingest data files, you need to define the structure of the data you'll be
-          ingesting by specifying the schema of the data. The schema is a template for the data
-          you'll ingest later. You'll specify the number and names of the data categories - the
-          tables and columns within the tables - and any associations between columns in separate
-          tables, if multiple tables contain the same data category (for instance, if you have a
-          "subject" table and a "sample" table, and both tables contain a column of the same subject
-          IDs).
-          <Tabs classes={{ root: classes.tabsRoot }} value={value} onChange={handleChange}>
+      <FormProvider {...formMethods}>
+        <form className={classes.contentContainer} onSubmit={handleSubmit(onSubmit)}>
```
s-rubenstein, line 134: Same as the comment above, I would say just leave off the submit for now until implementation time.

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
```diff
@@ -75,93 +90,139 @@ interface TabConfig {
...
-              />
-            ))}
-          </Tabs>
-          {tabs.map((tabConfig: TabConfig, i: number) => (
-            <TabPanel value={value} index={i} key={`dataset-schema-creation-tabpanel-${i}`}>
-              {tabConfig.content}
-            </TabPanel>
-          ))}
-        </div>
+              <div key={`dataset-schema-creation-tabpanel-${i}`} hidden={currentTab !== i}>
```
s-rubenstein: I would recommend pulling this out into a component. There's enough display complexity here that it might help with understandability to pull it out. I also think that this is very reusable if you have other tab based forms.

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
```diff
@@ -75,93 +90,139 @@ interface TabConfig {
...
+                    disableElevation
+                    type="button"
+                    className={classes.tabButton}
+                    onClick={() => setCurrentTab(i + 1)}
+                  >
+                    Go to Step {i + 2}
+                  </Button>
+                ) : (
+                  <Button
+                    color="primary"
```
s-rubenstein, line 180: Nit: For the sake of understandability, would you mind changing the order of the props to match above? Just to make it clear the only difference is `type` and `onClick`

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -0,0 +1,334 @@
...
+  Autocomplete,
+} from '@mui/material';
+import { clsx } from 'clsx';
+import { TdrState } from 'reducers';
+import { Controller, useFormContext } from 'react-hook-form';
+import SimpleMDE from 'easymde';
+import { SimpleMdeReact } from 'react-simplemde-editor';
+import { GCP_REGIONS, AZURE_REGIONS } from 'constants/index';
+import isEmail from 'validator/lib/isEmail';
+import WithoutStylesMarkdownContent from '../../common/WithoutStylesMarkdownContent';
```
s-rubenstein, line 22: Optional Nit: It looks like generally y'all use absolute paths rather than relative? Should this be an absolute path?

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -0,0 +1,334 @@
...
+            <WithoutStylesMarkdownContent markdownText={markdownText} />,
+          );
+        },
+        status: false,
+      } as SimpleMDE.Options),
+    [],
+  );
+
+  return (
+    <Grid container rowSpacing={2} columnSpacing={5} className={classes.contentContainer}>
```
s-rubenstein, line 85: Just to confirm, is this workable with a screen reader? And how does it behave with keyboard only navigation? (It looks okay, just wanna check to be sure.)

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -0,0 +1,334 @@
...
+        <label
+          htmlFor="dataset-description"
+          className={clsx(classes.formLabel, { [classes.formLabelError]: errors.description })}
+        >
+          Dataset description*
+        </label>
+        <Controller
+          name="description"
+          control={control}
+          rules={{ required: 'description is required' }}
```
s-rubenstein: Should there be a max length for this?

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -0,0 +1,334 @@
...
+              id="dataset-cloudPlatform"
+              className={classes.formInput}
+              {...field}
+              onChange={(event: any, change: any) => {
+                const cloudPlatform = event.target.value;
+                setRegionOptions(cloudPlatform === 'gcp' ? GCP_REGIONS : AZURE_REGIONS);
+                field.onChange(event, change);
+              }}
+            >
+              <MenuItem value="gcp">Google Cloud Platform (GCP)</MenuItem>
```
s-rubenstein: Could "gcp" be abstracted into a constant?

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -0,0 +1,334 @@
...
+          )}
+        />
+      </Grid>
+
+      <Grid item xs={12} className={classes.formFieldContainer}>
+        <label
+          htmlFor="dataset-custodians"
+          className={clsx(classes.formLabel, { [classes.formLabelError]: errors.custodians })}
+        >
+          Custodians*
```
s-rubenstein, line 292: Nit: What is the `*` afterwards for?

s-rubenstein:
Please ping me when you are ready for a next round!

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
```diff
@@ -75,93 +90,139 @@ interface TabConfig {
...
+              Learn more about ingesting data into the Terra Data Repo:
+              <ul className={classes.helpList}>
+                <li>
+                  <a
+                    className={clsx(classes.jadeLink, classes.helpListLink)}
+                    href="#"
+                    target="_blank"
+                  >
+                    Dataset schema overview <OpenInNew className={classes.jadeLinkIcon} />
+                  </a>
```
snf2ye, line 206: Is there a ticket for making this a non-dead link? just don't want to forget!

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -0,0 +1,334 @@
...
+        },
+        status: false,
+      } as SimpleMDE.Options),
+    [],
+  );
+
+  return (
+    <Grid container rowSpacing={2} columnSpacing={5} className={classes.contentContainer}>
+      <Grid item xs={12}>
+        <Typography variant="h3">Dataset information</Typography>
```
snf2ye, line 87: Thoughts on the Capitalization of these headers? I see they match the mocks, but this seems a bit informal. I think we should consider capitalizing the following:
- Dataset Information
- Description (Can we drop "dataset" from this?)
- Terra Project
- Secure Monitoring

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -0,0 +1,334 @@
...
+          )}
+        />
+      </Grid>
+
+      <Grid item xs={6} className={classes.formFieldContainer}>
+        <label
+          htmlFor="dataset-region"
+          className={clsx(classes.formLabel, { [classes.formLabelError]: errors.region })}
+        >
+          Region*
```
snf2ye, line 216: Since we're selecting from a distinct list of possible regions, I don't think you should be able to edit the region once you've selected one. It's strange to me that once you select one, the cursor is still present and you can backspace letters from the region. This is different than how the Cloud Platform option is working. 

https://user-images.githubusercontent.com/13254229/205720985-12e85ed2-1f5b-402d-9716-39cfe046073a.mov

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -0,0 +1,334 @@
...
+        <label
+          htmlFor="dataset-name"
+          className={clsx(classes.formLabel, { [classes.formLabelError]: errors.name })}
+        >
+          Name your dataset*
+        </label>
+        <TextField
+          id="dataset-name"
+          className={classes.formInput}
+          placeholder="Dataset name"
```
snf2ye, line 100: Can we also add placeholders for the other fields?
- description
- Terra Project
- Region
- Steward/custodian list

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -0,0 +1,334 @@
...
+          rules={{ required: true }}
+          render={({ field }) => (
+            <Select
+              id="dataset-cloudPlatform"
+              className={classes.formInput}
+              {...field}
+              onChange={(event: any, change: any) => {
+                const cloudPlatform = event.target.value;
+                setRegionOptions(cloudPlatform === 'gcp' ? GCP_REGIONS : AZURE_REGIONS);
+                field.onChange(event, change);
```
snf2ye: Edge case: If the user changes the region, we should also clear any value already set in the region field.

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -0,0 +1,334 @@
...
+              renderInput={(params: any) => (
+                <TextField
+                  {...params}
+                  error={!!errors.stewards}
+                  helperText={errors.stewards ? errors.stewards.message : ''}
+                />
+              )}
+              {...field}
+              onChange={(_event: any, change: any) => {
+                field.onChange(change);
```
snf2ye, line 280: I don't understand this pattern - why are we calling onChange from within the onChange function?

src/constants/index.ts
```diff
@@ -206,3 +208,122 @@ export const TABLE_DEFAULT_ROWS_PER_PAGE = 100;
...
+  { label: 'Norway West', name: 'norwaywest' },
+  { label: 'Switzerland West', name: 'switzerlandwest' },
+  { label: 'UK West', name: 'ukwest' },
+  { label: 'UAE Central', name: 'uaecentral' },
+  { label: 'Brazil Southeast', name: 'brazilsoutheast' },
+];
+
+const generateGCPRegions: any = () => {
+  const servers = [
+    { name: 'asia-east', count: [1, 2], letters: ['a', 'b', 'c'] },
```
snf2ye, line 217: What do the "count" and "letters" options do?

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -0,0 +1,334 @@
...
+          variant="outlined"
+          error={!!errors.name}
+          helperText={errors.name ? errors.name.message : ''}
+          {...register('name', {
+            required: 'name is required',
+            minLength: { value: 1, message: 'Name must be 1+ characters long' },
+            maxLength: { value: 511, message: 'Name must be less than 511 characters long' },
+            pattern: {
+              value: /^[a-zA-Z0-9][_a-zA-Z0-9]*$/,
+              message: 'Name should fit pattern: ^[a-zA-Z0-9][_a-zA-Z0-9]*$',
```
snf2ye: nit: Capitalization Consistency - Can we either capitalize "name" for the required error message or make the rest of "Name" options lower case?

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
lu-c: No *good* reason 😓 I had it in there initially so I could have the form behavior (with the react-hook-form library) set up, just to confirm the form behavior was working as intended, and was passing the right values to the submission obj.

I hope its okay if we overlook this for now just because of the feature flag, and the implementation has been added already in the follow up branch!

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
lu-c: (see above)

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
lu-c: Would you mind clarifying which part should be a new component? I have it set up rn so the tab contents are separate components... do you mean the tabs should be its own one too?

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
lu-c: Oh shoot, I didnt notice there was a maxlen for this in the mock up!!! Ok, fixing!

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
lu-c: Updated! There was already a constant for this! :D

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
lu-c: the * means the field is required to have a value

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
lu-c: No, let me create one now 👍

lu-c: https://broadworkbench.atlassian.net/browse/DR-2859

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
lu-c: I prefer to capitalize all first letters in any titles or labels too, but I think this is the route that other apps have done? Like terra. So maybe it's better for consistency?

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
lu-c: Yes! Let me capitalize those!

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
lu-c: ok! I wanted this to be different actually, because the cloud platform only has 2 options, so a select is easy. But when the possible regions are > 100, i didn't want the user to have to scroll through the list, and wanted to give them the ability to type the answer in.

I removed the flag `freeSolo` that allowed that component to allow any value (so now it is constrained to the known options list), but still allows the user to type.

Does this help? Let me know if i should readjust!

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
lu-c: Added placeholders for everything but the terra project field - Going to hold off on that one until the follow up PR, if that's okay! The 2nd PR will fix the terra project field so it auto populates the first one, but it has to make a backend call to get them.

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
lu-c: Fixed! :)

src/constants/index.ts
lu-c: I just got these values from the GCP region list in google's documetnation for what the server options are. There are a lot of "asia-east1-a" "asia-east2-a", etc. Instead of copying them all in to a giant variable, I compacted it a little so we generate them instead.
https://cloud.google.com/compute/docs/regions-zones

Unfortunately, this list also changes in the follow up PR too! I didn't realize jade-repo is not expecting letters at all...

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
lu-c: (@snf2ye What do you think?)

I haven't paid much attention to the imports, but I assumed this was okay?

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
lu-c: There are nested components here, so react-hook-forms is wrapping form elements. It knows how to handle updating the form value for default components, but then struggles with more complex methods. They let you have access to the underlying methods when you wrap it in the Controller object, but then you have 2 onChange methods.

The first onChange is from mui/autocomplete. The second onchange being called belongs to the react-hook-form/controller

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
lu-c: Tabbing through, it looks okay to me!

There is one area that sucks for keyboard navigation though, the description markdown editor treats the tabs as characters. Not sure how to resolve that right now, but I will add this as a bug ticket!

Otherwise, it seems okay!

src/components/dataset/overview/DatasetOverviewPanel.jsx
lu-c: Ah yes! Pulled from develop ✨

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
s-rubenstein: Sure, if that is team agreement, that's probably okay.

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
s-rubenstein: I think that the ternary content seems fairly standard (with the stepping through tabs). Because of that, I would pull the div out that starts on line 161.

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
s-rubenstein: I think the Terra UX has first letter only. I also prefer first letter of each word. Might be worth checking with UX to see if it is intentional?

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
s-rubenstein: TOL: I wonder if that would be doable with a subcomponent that takes a prop named `required` and automatically adds that `*`. It feels a little strange to put that in the text, because I think it could easily lead to inconsistencies in the way required fields are handled across the UI.

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -0,0 +1,355 @@
...
+        <div className={classes.formFieldDescription}>
+          A Steward, or Data Owner, is the person who created the dataset. While they are ultimately
+          liable for the data, they can assign the hands-on data management to another person by
+          assigning the Custodian role.
+        </div>
+        <Controller
+          name="stewards"
+          control={control}
+          rules={{
+            validate: {
```
s-rubenstein: Do you think that these validation rules could be reusable in other forms? Did you think about pulling them out into a helper file so that other forms could make use of them? (Especially ones like this one that have some meat, but are going to be identical for many email validation checks)

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -0,0 +1,355 @@
...
+          The Custodian role is defined on a dataset. Someone may be the Custodian for one or more
+          studies. A Custodian is responsible for creating data snapshots over datasets and
+          controlling access to those snapshots.
+        </div>
+        <Controller
+          name="custodians"
+          control={control}
+          rules={{
+            validate: {
+              isEmail: (values: string[]) => {
```
s-rubenstein: (See above comment, this is actually an example where that reusability would be very powerful.)

src/libs/render-utils.tsx
```diff
@@ -5,7 +5,7 @@ import CopyTextButton from 'components/common/CopyTextButton';
 import InfoHoverButton from 'components/common/InfoHoverButton';
 import { DatasetSummaryModel, SnapshotSummaryModel } from '../generated/tdr';
 
-const cloudPlatforms = { gcp: 'Google Cloud Platform', azure: 'Microsoft Azure' };
+export const cloudPlatforms = { gcp: 'Google Cloud Platform', azure: 'Microsoft Azure' };
```
s-rubenstein: For a potential improvement for the future, I wonder if you could put the regions in this constant? Then you could do some filtering and mapping to get regions for a particular cloudPlatform, rather than needing to do conditionals. I think that might be more powerful if you end up needing to add another cloud platform to simplify the number of places you would need to add. (I think you might be able to get it to the point where the only place you would need to modify is here, and every bit of the UI would automatically update to use that new cloudPlatform)

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
lu-c: Lou confirms they want to only capitalize the first letter of the first word

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
lu-c: Ahm, yeah, i can pull it out into a helper file

src/libs/render-utils.tsx
lu-c: The future is now! updated this by moving the constant out of the utils and into the CONSTANTS file where it belongs!

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
lu-c: Turns out it already exists as a component! 🎉

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
```diff
@@ -75,93 +91,139 @@ interface TabConfig {
...
+
+  const formMethods = useForm({
+    mode: 'onTouched',
+    reValidateMode: 'onChange',
+    defaultValues: {
+      name: '',
+      description: '',
+      terraProject: '',
+      enableSecureMonitoring: 'true',
+      cloudPlatform: 'gcp',
```
s-rubenstein: This could also use the constant.

src/constants/index.ts
s-rubenstein: Could you add that as a comment? I think that is valuable information (showing what the compound version is)

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -0,0 +1,342 @@
...
+              />
+              {errors.description && (
+                <span className={classes.formInputError}>{errors.description.message}</span>
+              )}
+            </>
+          )}
+        />
+      </Grid>
+
+      <Grid item xs={6}>
```
s-rubenstein, line 145: It feels like a lot of the Grid item code is repeated. I wonder if that could be an abstracted component? Especially because this is a very meaty component right now, and I think it could easily end up with high complexity as things need to change.

s-rubenstein (commented):
I think trying to break up the `DatasetSchemaInformationView` component is the last thing that I would try to change! After that it should be good to go.

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
lu-c: updated!

src/constants/index.ts
lu-c: done!

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
lu-c: I am going to push back on that a little? Grid is already a component, I don't really see too much value in abstracting it out further, we would be trading lines of code for lines of config imo. I prefer not to make components unless they are really reuseable as well, and I don't think a Grid surrounding a single component will be

s-rubenstein: My biggest concern is the repeated code of the children. For example:
```
<label
          htmlFor="{FIELD}"
          className={clsx(classes.formLabel, { [classes.formLabelError]: {ERROR CLASS} })}
        >
          {LABEL DISPLAY}
        </label>
```
Is almost verbatim repeated across every grid component. However, especially since its not my codebase, if you think that the value gain is negligible, I am totally okay being overruled here. I just always find it a bit of a yellow flag when I see a lot of repeated code like this. (Another option would be to pull form fields out into a list of objects, and then map over that list to create the grid elements, but I think I also defer to the folks working in the codebase for style preferences)

s-rubenstein (approved):

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
lu-c: In this case, I also want to defer to the tech leads in the project too 😂 I'm also interested in hearing other opinions too! Maybe I'm off base here! :)

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
snf2ye: Gotcha! Yes, I like that you can't actually submit an invalid region now.

snf2ye (approved):

PR #1250, opened 2022-11-23
[DR-2851] Removed feature flag on route

lu-c:
I think this will fix our feature flag hiccup!

### DIFF
--- a/src/routes/Private.jsx
+++ b/src/routes/Private.jsx
@@ -72,9 +72,10 @@ const tabsConfig = [
   { label: 'Ingest Data', path: '/ingestdata' },
 ];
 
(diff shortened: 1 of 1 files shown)


nmalfroy (approved):

snf2ye (approved):

okotsopoulos (commented):
Speedy merge before I could approve :) thank you so much for fixing!  I was happy with the behavior I saw on your dev instance with this change deployed.

For all combos of (feature flag = true, feature flag = false, feature flag unset) I was able to successfully navigate through all expected routes - clicking into snapshots, datasets, and the activity page were previously not working consistently for me.

PR #1252, opened 2022-11-29
[DR-2838] Dataset schema creation - Step 2 and beyond!

lu-c:
[Link to Dataset Schema Builder on dev](https://jade-cl.datarepo-dev.broadinstitute.org/datasets/new)

This PR:
- Re-orders dataset info drop downs
- If changing cloud provider, reset billing profiles that are available
  and reset and selected billing profiles and regions
- Uses nulls and not empty strings for empty values
- Makes description and curators not required
- Makes sure that deleting columns or tables in the JSON viewer properly
  clears selection
- Makes sure that the input of the datatype dropdown shows the formatted
  value
- Scrolls to the top when changing tabs
- When clicking create, opens a job spinner instead just redirecting to the
  datasets page (and when the job finishes, allow the user to go to the
  new dataset)
- Updates tests to take modal into account
- Modifies the SchemaPanel component to reuse between the live edit and the dataset/snapshot schema views


src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
```diff
@@ -80,158 +85,220 @@ const styles = (theme: CustomTheme) => ({
...
-  ];
-
-  const onSubmit = (data: any) => {
-    // eslint-disable-next-line
-    console.log(data);
-  };
+const DatasetSchemaCreationView = withStyles(styles)(
+  ({ classes, dispatch, profiles, history }: IProps) => {
+    const [currentTab, setCurrentTab] = React.useState(0);
+    const changeTab = (_event: any, newCurrentTab: any) => setCurrentTab(newCurrentTab);
```
snf2ye: Could newCurrentTab be typed?

src/components/dataset/schemaCreation/DatasetSchemaCreationView.test.tsx
```diff
@@ -0,0 +1,1021 @@
...
+
+const initialState = {
+  searchString: '',
+  profiles: {
+    profiles: [{ profileName: 'default profile' }, { profileName: 'second profile' }],
+  },
+  user: _.cloneDeep(initialUserState),
+};
+
+// const fillValidInfoFields = () => {
```
s-rubenstein: This file seems to be largely commented out code. Could we remove that? Do you still need it for some reason?

src/components/dataset/schemaCreation/DatasetSchemaCreationView.test.tsx
```diff
@@ -0,0 +1,1021 @@
...
+                },
+              },
+            ],
+          };
+          expect(elem.attr('data-cy')).to.equal(JSON.stringify(comparison));
+        });
+      });
+
+      it('updating column name should update relationship value', () => {
+        cy.get('#schemabuilder-createTable').click();
```
s-rubenstein: The setup seems to be repeated between these tests, could that be abstracted to a helper?

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
```diff
@@ -0,0 +1,803 @@
...
+interface IProps extends WithStyles<typeof styles> {
+  userEmail: string;
+}
+
+const DatasetSchemaBuilderView = withStyles(styles)(({ classes }: IProps) => {
+  const { register, getValues, setValue } = useFormContext();
+  const [datasetSchema, setDatasetSchema] = useState(
+    getValues().schema as DatasetSpecificationModel,
+  );
+  const [selectedTable, setSelectedTable] = useState(-1);
```
s-rubenstein, line 136: Why does this need to default to `-1`? Is that a dummy value to replicate having not started yet? Could undefined be used as a more semantic representation?

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
```diff
@@ -0,0 +1,803 @@
...
+  const [selectedColumn, setSelectedColumn] = useState(-1);
+  const [expandedTables, setExpandedTables] = useState({} as any);
+  const [relationshipModalOpen, setRelationshipModalOpen] = useState(false);
+  const [relationshipModalDefaultValues, setRelationshipModalDefaultValues] = useState({
+    from: '',
+    to: '',
+    expandedTables: {},
+  });
+
+  const [anchorElDetailsMenu, setAnchorElDetailsMenu] = React.useState<null | HTMLElement>(null);
```
s-rubenstein: Does this work if you delete the null from the function call? I would expect it to default to undefined in that case.

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
```diff
@@ -0,0 +1,803 @@
...
+
+  const handleClickDetailsMenu = (event: React.MouseEvent<HTMLButtonElement>) => {
+    setAnchorElDetailsMenu(event.currentTarget);
+  };
+  const handleCloseDetailsMenu = () => setAnchorElDetailsMenu(null);
+
+  const onJsonViewerChange = React.useCallback(
+    (value: string) => {
+      try {
+        // using these regexes to add / remove quotation marks just because the viewer will then color code
```
s-rubenstein: Which regexes? Is that value? potentialSchema?

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
```diff
@@ -0,0 +1,803 @@
...
+        // the keys and values differently. It's easier to read for modifying, but the downside is that
+        // it isnt copy-pastable as ready json.
```
s-rubenstein: TOL: Could you override the copy behavior to make it ready json, although that could be a bit magic/behind the scenes for users? Or add a copyToClipboard button that does some work for them?

s-rubenstein:
I added a few initial comments, but I have only gotten a bit into it. Is it possible to break this PR up some? I don't know if I can give a solid review with this many changes at once, especially missing context?

Also, would you be able to drop a few screenshots into the PR description? That might help me understand what is going on.

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
lu-c: When I was setting the state to `undefined`, and then changing it, i was getting an error for setting an unset variable - I think this has to stay

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
lu-c: Oh shoot, I deleted those regexes

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
lu-c: i will add a ticket for a copyToClipboard function, to try to keep this pr down (better late than never?)

src/components/dataset/schemaCreation/DatasetSchemaCreationView.test.tsx
lu-c: These are valid unit tests! I forgot to uncomment them to make them while working locally

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
s-rubenstein: Interesting! I wonder if that's a typescript thing. Well, sounds fine to me then.

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
s-rubenstein: Probably the same issue as above, this sounds good.

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
```diff
@@ -0,0 +1,800 @@
...
+                            rel.to.column === origName
+                          ) {
+                            rel.to.column = newName;
+                          }
+                          return rel;
+                        },
+                      );
+                    }
+                    setDatasetSchema(schemaCopy);
+                  }}
```
snf2ye: I think we generally put this sort of function at the top of the file

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -144,19 +118,33 @@ const DatasetSchemaInformationView = withStyles(styles)(({ classes }: IProps) =>
 
       <Grid item xs={6}>
         <label
-          htmlFor="dataset-terraProject"
-          className={clsx(classes.formLabel, { [classes.formLabelError]: errors.terraProject })}
+          htmlFor="dataset-defaultProfile"
+          className={clsx(classes.formLabel, { [classes.formLabelError]: errors.defaultProfile })}
         >
-          Terra project*
+          Billing Project*
```
lu-c: Update this name to Billing Profile

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -144,19 +118,33 @@ const DatasetSchemaInformationView = withStyles(styles)(({ classes }: IProps) =>
...
-          name="terraProject"
+          name="defaultProfile"
           control={control}
+          rules={{ required: 'default project is required' }}
           render={({ field }) => (
-            <Select id="dataset-terraProject" className={classes.formInput} {...field}>
-              <MenuItem value="yes">Yes</MenuItem>
-              <MenuItem value="no">No</MenuItem>
-            </Select>
+            <Autocomplete
```
lu-c: Update this to a select

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -144,19 +118,33 @@ const DatasetSchemaInformationView = withStyles(styles)(({ classes }: IProps) =>
...
+                />
+              )}
+              {...field}
+              onChange={(_event: any, change: any) => {
+                field.onChange(change);
+              }}
+            />
           )}
         />
       </Grid>
```
lu-c: Make default secure monitoring value "no"

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
```diff
@@ -0,0 +1,800 @@
...
+      width: '100%',
+      padding: 10,
+      marginLeft: 10,
+      marginTop: 50,
+    },
+    formInputDatatype: {
+      maxWidth: 200,
+    },
+    jsonWrapper: {
+      backgroundColor: '#272822',
```
lu-c, line 110: add comment describing what this color is

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
```diff
@@ -0,0 +1,800 @@
...
+      borderRadius: 10,
+      padding: 20,
+    },
+  } as any);
+
+interface IProps extends WithStyles<typeof styles> {
+  userEmail: string;
```
lu-c: deletethis

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
```diff
@@ -0,0 +1,800 @@
...
+  const [selectedColumn, setSelectedColumn] = useState(-1);
+  const [expandedTables, setExpandedTables] = useState({} as any);
+  const [relationshipModalOpen, setRelationshipModalOpen] = useState(false);
+  const [relationshipModalDefaultValues, setRelationshipModalDefaultValues] = useState({
+    from: '',
+    to: '',
+    expandedTables: {},
+  });
+
+  const [anchorElDetailsMenu, setAnchorElDetailsMenu] = React.useState<null | HTMLElement>(null);
```
lu-c: Remove `React.`

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
```diff
@@ -0,0 +1,800 @@
...
+                  }}
+                />
+              </div>
+              <div>
+                <label htmlFor="column-datatype" className={classes.formLabel}>
+                  Data type
+                </label>
+                <Autocomplete
+                  id="column-datatype"
+                  options={columnDatatypeOptions}
```
lu-c: Remove this variable and inline `_.keys()`

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
```diff
@@ -0,0 +1,800 @@
...
+          {selectedTable !== -1 && selectedColumn === -1 && (
+            <div className={classes.schemaBuilderDetailView} data-cy="schemaBuilder-detailView">
+              <div className={classes.schemaBuilderStructureViewControls}>
+                <Typography variant="h4">Table attributes</Typography>
+                <IconButton
+                  id="details-menu-button"
+                  size="small"
+                  color="primary"
+                  className={classes.iconButton}
+                  onClick={handleClickDetailsMenu}
```
lu-c: let's pull out the icon stuff to a thing. :P

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
```diff
@@ -0,0 +1,800 @@
...
+  const openDetailsMenu = Boolean(anchorElDetailsMenu);
+
+  const columnDatatypeOptions: string[] = _.keys(TableDataType);
+
+  const handleClickDetailsMenu = (event: React.MouseEvent<HTMLButtonElement>) => {
+    setAnchorElDetailsMenu(event.currentTarget);
+  };
+  const handleCloseDetailsMenu = () => setAnchorElDetailsMenu(null);
+
+  const onJsonViewerChange = React.useCallback(
```
lu-c: remove `React.`?

lu-c:
disable column checkbox when one is selected

src/components/dataset/schemaCreation/DatasetSchemaCreationView.test.tsx
lu-c: Added a helper method to call these methods!

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
```diff
@@ -0,0 +1,780 @@
...
+                >
+                  <i className="fa fa-angle-up" />
+                </IconButton>
+
+                <IconButton
+                  id="datasetSchema-down"
+                  size="small"
+                  color="primary"
+                  className={classes.iconButton}
+                  disabled={selectedTable === -1}
```
snf2ye: There are other cases when we could disable this, like when the column is the last in the last table, etc. But I wonder if it's worth the more complicated logic.

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
```diff
@@ -0,0 +1,780 @@
...
+                  color="primary"
+                  className={classes.iconButton}
+                  style={{ marginLeft: 50 }}
+                  disabled={!datasetSchema.tables || datasetSchema.tables.length < 2}
+                  onClick={() => {
+                    setRelationshipModalDefaultValues({
+                      from: '',
+                      to: '',
+                      expandedTables: {},
+                    });
```
snf2ye: This matches the original default value - can we put this into a constant and share it between the two?

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
```diff
@@ -0,0 +1,780 @@
...
+    });
+    setSelectedTable(newIndex);
+    setSelectedColumn(-1);
+    setExpandedTables({
+      ...expandedTables,
+      [newIndex]: true,
+    });
+  };
+
+  const deleteTable = () => {
```
snf2ye, line 196: I mentioned the idea of having a confirmation dialog to confirm before deleting. I don't think I see that yet. 
(Also totally fine to push that off to another PR/ticket)

snf2ye (commented):
Thanks for walking us through some of the code today!
Just a few odds & ends comments.

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
lu-c: moved!

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
lu-c: I have a unit test that confirms that clicking the button when the col is already the last in the list doesnt do anything - We could disable the button instead too!

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
snf2ye: Whoops, I'm sorry, I did not mean to submit this comment. It looks like you were following a paradigm where only shared functions are at the top of the file and the rest are in line. I think that's fine if you want to revert. I'd rather stay consistent with the rest of the components.

lu-c:
Updated the PR with the comments - First, there is definitely a bug, but I could use help looking into it!

Right now, when you click on the detailed menu button, it now shows up like this:
Screen Shot 2022-12-13 at 7 31 12 PM
and it is totally detached from the popping element. I'm not sure exacty what changed to cause this and how to fix - the anchor element is still being declared the same way as before. The only difference is that the IconButton is now being handled in a helper component

Any ideas? :( Otherwise, i will revert this back to duplicating code

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
lu-c: I totally forgot about this! Let me add a ticket for it, and then if I can get it in today, I'll do it

src/components/common/overview/SchemaPanel.tsx
```diff
@@ -172,48 +212,120 @@ const renderColumnName = (
...
+  // Expanded nodes by id where the id is the {table index}
+  expanded?: Array<string>;
+  // Highlighted nodes by id where the id is the {table index}-{column index}
+  highlighted?: Array<string>;
+  // If passed in, callback method to execute when any node gets toggled
+  onNodeToggle?: (event: React.SyntheticEvent, nodeIds: string[]) => void;
+  // Array of react elements to add before a column label
+  beforeLabelIcons?: React.ReactElement[];
+  // Array of react elements to add after a column label
+  afterLabelIcons?: (table: TableModel, column: ColumnModel) => LabelIcon[];
```
snf2ye, line 258: Thanks for the helpful comments!

snf2ye (approved):
This is awesome! Thank you!

okotsopoulos (approved):
Thank you for the work on this!

I'm approving because my chief criteria for merging was that users who do *not* enable the feature flag see no degradation of behavior, I was able to interact with dataset and snapshot schemas in your dev environment with no issues in rendering (I know that was shared code that was changed).

I'm going through to offer comments on the new feature now, but didn't want to hold up your merge on them as I know we'll continue to work in this space going forward.

src/components/dataset/schemaCreation/DatasetCreationModal.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+interface IProps extends WithStyles<typeof styles> {
+  isOpen: boolean;
+  dataset?: DatasetModel;
+  creationIsProcessing: boolean;
+}
+
+const DatasetCreationModal = withStyles(styles)(
+  ({ classes, isOpen, dataset, creationIsProcessing }: IProps) => {
+    if (creationIsProcessing) {
```
okotsopoulos, line 50: If the dataset creation job fails, it's hard for the user to debug.

As an example, I tried to create a dataset with a billing profile whose account I was missing permissions for.  I got this ephemeral generic error message:
Screen Shot 2023-01-20 at 1 04 38 PM

But the underlying job has a more descriptive error cause:
Screen Shot 2023-01-20 at 1 04 00 PM

This is also an issue I've noticed in failed snapshot creation within the UI.  Probably would benefit from some separate design and planning, but it would be nice to provide a link to the job result within the UI.  We currently display job results on click within the Activity tab in a pop-up, but don't link directly (and the Activity tab is also unusable to non-admins given slow performance of Stairway queries).

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
```diff
@@ -0,0 +1,953 @@
...
+          onClose={() => {
+            setRelationshipModalOpen(false);
+          }}
+          onDelete={deleteRelationship}
+        />
+      )}
+    </div>
+  );
+});
+
```
okotsopoulos, line 952: I noticed that we can set relationships in this schema wizard, but we can't set assets.  AFAIK, a dataset must have assets in order to be used in TDR UI's snapshot creation wizard.  I'd recommend that we prioritize supporting their specification here.

src/components/dataset/schemaCreation/DatasetSchemaInformationView.tsx
```diff
@@ -208,7 +157,42 @@ const DatasetSchemaInformationView = withStyles(styles)(({ classes }: IProps) =>
...
+              isOptionEqualToValue={(option, value) => value.profileName === option.profileName}
+              renderInput={(params: any) => (
+                <TextField
+                  {...params}
+                  error={!!errors.defaultProfileId}
+                  helperText={errors.defaultProfileId ? errors.defaultProfileId.message : ''}
+                  placeholder="Default billing profile"
+                />
+              )}
+              getOptionLabel={(option: BillingProfileModel) => option.profileName || ''}
```
okotsopoulos, line 185: I agree with displaying a billing profile by its human-readable name, but what do you think about displaying the UUID secondarily?  Maybe something like
```
name (aaaa-bbbb-cccc-dddd)
```
where the UUID is styled to grey.  I'm thinking in case the user wants to interact with our billing profile endpoints, they'll need the UUID.

Also not sure if it's worth calling out in some way at this stage that even if a billing profile is accessible to the user, they will also need access to its underlying billing account… maybe the dataset creation error message in that case covers our bases well enough.

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
```diff
@@ -80,58 +86,100 @@ const styles = (theme: CustomTheme) => ({
...
+      name: null,
+      description: null,
+      terraProject: null,
+      enableSecureMonitoring: false,
       cloudPlatform: CLOUD_PLATFORMS.gcp.key,
-      region: '',
+      defaultProfileId: null,
+      region: null,
       stewards: [],
       custodians: [],
```
okotsopoulos, line 131: Is there a reason why we don't allow snapshot creators to be set at this stage too?

When we allow the policy members to be set here, is there an opportunity to reuse code from snapshot creation? The difference in UX threw me off a bit. 
Screen Shot 2023-01-20 at 3 15 40 PM

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
```diff
@@ -185,7 +233,25 @@ const DatasetSchemaCreationView = withStyles(styles)(({ classes }: IProps) => {
                     Submit
                   </Button>
                 )}
-              </TabPanel>
+                {_.keys(errors).length > 0 && (
```
okotsopoulos, line 236: Sorry, I don't think this is the best place for this comment!

I noticed that if I introduce an error to the schema definition but fix it, the error message remains:
Screen Shot 2023-01-20 at 12 21 52 PM

I also noticed that you can still submit a dataset creation request even if your form has errors.  Is that intentional?

src/components/dataset/schemaCreation/DatasetSchemaBuilderView.tsx
```diff
@@ -0,0 +1,953 @@
...
+      setSelectedColumn(selectedColumn + 1);
+    } else if (selectedTable !== -1 && selectedTable < schemaCopy.tables.length - 1) {
+      swapArrayLocs(schemaCopy.tables, selectedTable, selectedTable + 1);
+      setSelectedTable(selectedTable + 1);
+    }
+    setDatasetSchema(schemaCopy);
+  };
+
+  // ----------------------------------------
+  // CodeMirror (JSON viewer)
```
okotsopoulos, line 672: The CodeMirror is super cool.  I wonder if we want to "lock" it on default, though?  I could see a user wanting to create the schema via the GUI, inadvertently modifying the mirrored JSON, and not being clear on how to fix.  (For example, you could change the JSON for a column definition to illegally stack several types: it would then reflect in the GUI but you'd be unable to fix it there.)

Screen Shot 2023-01-20 at 12 22 22 PM

okotsopoulos (approved):
As promised, my follow-on comments from interacting with this feature :-)

PR #1257, opened 2022-12-01
[DR-2854] Deleted duplicate activity tab

lu-c:
Deleted duplicate tab, probably leftover from a bad merge?


otchet-broad (approved):
Looks great!

okotsopoulos (approved):
Thanks for fixing!  I clicked through a dataset's tabs and they all look good.

PR #1264, opened 2022-12-14
[DR-2832] added duos id field if it exists

lu-c:
Screen Shot 2022-12-14 at 4 42 26 PM

### DIFF
--- a/src/components/snapshot/overview/SnapshotOverviewPanel.tsx
+++ b/src/components/snapshot/overview/SnapshotOverviewPanel.tsx
@@ -105,7 +105,7 @@ function SnapshotOverviewPanel(props: SnapshotOverviewPanelProps) {
           <Grid item xs={12}>
             <Grid item xs={4}>
               {renderTextFieldValue(
-                'DUOS Id',
+                'DUOS ID',
                 snapshot.duosFirecloudGroup?.duosId,
                 'Link with a DUOS dataset ID to automatically add DAC approved users as snapshot readers',
               )}
--- a/src/components/snapshot/overview/SnapshotOverviewPanel.tsx
+++ b/src/components/snapshot/overview/SnapshotOverviewPanel.tsx
@@ -102,6 +102,11 @@ function SnapshotOverviewPanel(props: SnapshotOverviewPanelProps) {
               useMarkdown={true}
             />
           </Grid>
+          <Grid item xs={12}>
+            <Grid item xs={4}>
+              {renderTextFieldValue('DUOS Id', snapshot.duosFirecloudGroup?.duosId)}
+            </Grid>
+          </Grid>
           <Grid item xs={4}>
             <Typography variant="h6">Root dataset:</Typography>
             <Typography data-cy="snapshot-source-dataset" className={classes.datasetText}>
--- a/src/components/snapshot/overview/SnapshotOverviewPanel.tsx
+++ b/src/components/snapshot/overview/SnapshotOverviewPanel.tsx
@@ -104,7 +104,11 @@ function SnapshotOverviewPanel(props: SnapshotOverviewPanelProps) {
           </Grid>
           <Grid item xs={12}>
             <Grid item xs={4}>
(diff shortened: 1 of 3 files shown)


s-rubenstein (approved):
LGTM

src/components/snapshot/overview/SnapshotOverviewPanel.tsx
```diff
@@ -102,6 +102,11 @@ function SnapshotOverviewPanel(props: SnapshotOverviewPanelProps) {
               useMarkdown={true}
             />
           </Grid>
+          <Grid item xs={12}>
+            <Grid item xs={4}>
+              {renderTextFieldValue('DUOS Id', snapshot.duosFirecloudGroup?.duosId)}
```
okotsopoulos: Could we please display this as DUOS ID rather than DUOS Id?

Could we also include a tooltip on an information icon with this text on hover? "Link with a DUOS dataset ID to automatically add DAC approved users as snapshot readers"

src/components/snapshot/overview/SnapshotOverviewPanel.tsx
```diff
@@ -102,6 +102,11 @@ function SnapshotOverviewPanel(props: SnapshotOverviewPanelProps) {
               useMarkdown={true}
             />
           </Grid>
+          <Grid item xs={12}>
```
okotsopoulos, line 105: Hmm, logically I don't think I would put the DUOS ID at the top of our metadata grid, above even the root dataset and snapshot ID. I know that's where it was positioned in the mock though.

We can check with Lou about ordering, but I'd vote for putting this after PHS ID and Consent Code, since these are all fields which impact snapshot access.

okotsopoulos (changes requested):
I put in some requests inline, but the biggest blocker is that I don't think this works for snapshots that have a DUOS ID!  From the ticket description:

> To obtain this, the existing call to [retrieveSnapshot](https://data.terra.bio/swagger-ui.html#/snapshots/retrieveSnapshot) will need to have “DUOS” in its includes list.

Were you able to test this with a snapshot that had a DUOS link?  Can we get this UI on your developer environment?  Happy to pair with you on testing if it helps.

src/components/snapshot/overview/SnapshotOverviewPanel.tsx
lu-c: Yes, can make these changes!

lu-c:
> To obtain this, the existing call to [retrieveSnapshot](https://data.terra.bio/swagger-ui.html#/snapshots/retrieveSnapshot) will need to have “DUOS” in its includes list.

Shoot, I think I had this in the other branch, and forgot to also move it here! Let me make this change :)

src/components/snapshot/overview/SnapshotOverviewPanel.tsx
lu-c: Root dataset > snapshot id > duos id

Is an order that makes sense to me too! If duos id is only used for modifying user access, i think we could move more of the data around in the summary view - Maybe we could start adding section headers, so we can have

Snapshot information: id, date created, billing profile id
Snapshot data metadata: root dataset, storage, google data project
Snapshot access: PHS ID, consent code, duos ID

?

okotsopoulos: I like where you're going with this!  We've exposed quite a bit of metadata over the last few months and I agree that there's an opportunity for more logical grouping.  But I'm going to approve this PR as it stands, it would be better to get to check with Lou and the rest of the team about the subsections and handle separately.

okotsopoulos (approved):
Thanks for deploying to your dev environment!  I was able to verify expected behavior with this snapshot: https://jade-cl.datarepo-dev.broadinstitute.org/snapshots/fd8db08b-52d6-4c35-89b5-302f85b15907

Screen Shot 2022-12-16 at 1 18 27 PM

src/components/snapshot/overview/SnapshotOverviewPanel.tsx
lu-c: Sounds good! :)

PR #1268, opened 2022-12-19
[DR-2838] added null ptr protection

lu-c:
Fixed null ptr exception on dev:

https://jade.datarepo-dev.broadinstitute.org/snapshots/e2151834-13cd-4156-9ea2-168a1b7abf60

On this page, the Activity tab will break. To render the description, we would pull out the `request.json` key-value in the JournalEntryTableModel's "mutation" field, then parse out this json string, and grab values. Some of these values did not exist.

However, after this PR, it wont!

### DIFF
--- a/src/components/table/JournalEntryTable.tsx
+++ b/src/components/table/JournalEntryTable.tsx
@@ -91,8 +91,8 @@ const JournalEntryTable = withStyles(styles)(
           const details = [];
           if (row.entryType === JournalEntryModelEntryTypeEnum.Create) {
             note = description;
-          } else if (_.get(parsedReq, 0) === 'bio.terra.model.IngestRequestModel') {
-            const strategy = _.get(parsedReq, [1, 'updateStrategy'], '');
+          } else if (parsedReq?.[0] === 'bio.terra.model.IngestRequestModel') {
+            const strategy = parsedReq?.[1]?.updateStrategy || '';
             const casedStrategy = `${strategy[0].toUpperCase()}${strategy.substring(1)}`;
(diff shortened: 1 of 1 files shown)


src/components/table/JournalEntryTable.tsx
```diff
@@ -91,8 +92,8 @@ const JournalEntryTable = withStyles(styles)(
           const details = [];
           if (row.entryType === JournalEntryModelEntryTypeEnum.Create) {
             note = description;
-          } else if (parsedReq[0] === 'bio.terra.model.IngestRequestModel') {
-            const strategy = parsedReq[1].updateStrategy;
+          } else if (_.get(parsedReq, 0) === 'bio.terra.model.IngestRequestModel') {
```
pshapiro4broad: Can you use optional chaining here?
```suggestion
          } else if (parsedReq?.[0] === 'bio.terra.model.IngestRequestModel') {
```

pshapiro4broad (approved):
Looks OK to me. It might be good to converge on either optional chaining or lodash `get()` in the code base.

okotsopoulos (approved):
Thanks for fixing this!

For posterity's sake, would you mind expanding the PR's description with more information on why we hit on this NPE?    Even better would be if we can demonstrate that this is fixed in a dataset/snapshot in your personal deployment.

FWIW, I find the optional chaining more readable, and when you consider the added type safety and performance it has my vote for null protection within our repo.  Thanks @pshapiro4broad for bringing it up, and for spurring good convo in #terra-ui Slack: https://broadinstitute.slack.com/archives/C01EHNUM73R/p1671561517695269

PR #1271, opened 2022-12-21
[DR-2875] Fixed datatype npe in snapshot view

lu-c:
https://jade.datarepo-dev.broadinstitute.org/snapshots/f90f5d7f-c507-4e56-abfc-b965a66023fb

This snapshot page was failing on page load because there was an empty column datatype, and then we were substring'ing it without checking for null.

### DIFF
--- a/src/components/common/overview/SchemaPanel.tsx
+++ b/src/components/common/overview/SchemaPanel.tsx
@@ -143,7 +143,7 @@ const renderColumnName = (
   const isPk = _.includes(table.primaryKey || [], column.name);
   retVal.push(
     <span key="dt" className={classes.columnBox} title={column.datatype}>
-      {(column.datatype?.substring(0, 1) || 'N').toUpperCase()}
+      {column.datatype ? column.datatype.substring(0, 1).toUpperCase() : '?'}
       {column.array_of && <span className={classes.columnSubscript}>[ ]</span>}
     </span>,
   );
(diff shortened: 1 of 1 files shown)


src/components/common/overview/SchemaPanel.tsx
```diff
@@ -143,7 +143,7 @@ const renderColumnName = (
   const isPk = _.includes(table.primaryKey || [], column.name);
   retVal.push(
     <span key="dt" className={classes.columnBox} title={column.datatype}>
-      {column.datatype.substring(0, 1).toUpperCase()}
+      {(column.datatype?.substring(0, 1) || 'N').toUpperCase()}
```
okotsopoulos: Is this `N` the default short display name for a column with no datatype?  This seems counterintuitive to me, I would think this would be a number.

Could we just make it the empty string?  I think that's a more appropriate decoration!

okotsopoulos (approved):
This generally looks good!

The link you included in your PR description is to a snapshot from your localhost.  Can you share a publicly accessible link or maybe a screenshot of what you're seeing locally?  It would help me to visualize what the display behavior is when a column has no datatype.

src/components/common/overview/SchemaPanel.tsx
lu-c: Screen Shot 2022-12-21 at 3 49 43 PM
This is what it looks like blank - I know the 'n' is counterintuitive too, but the blank just looks wrong to me!

What would you think of using this then?
https://fontawesome.com/search?q=empty&o=r

An icon for empty?

lu-c: Resolution: We updated this so "unknown datatype" now shows '?'

pshapiro4broad (approved):

PR #1274, opened 2022-12-29
[DR-2842] Snapshot stewards should be able to update DUOS link

lu-c:
https://broadworkbench.atlassian.net/browse/DR-2842

**Snapshot stewards can now set or unset the snapshot's linked DUOS ID from the UI.**

My [developer environment](https://jade-ok.datarepo-dev.broadinstitute.org/snapshots) reflects these changes.

Because the underlying flights can take around 10 seconds to complete, I've added protections to the `EditableFieldView` component (also used for updating snapshot descriptions, consent codes, etc).  While a save is pending:
- The save and cancel buttons are disabled
- The field cannot be edited further
- The save button shows a loading spinner

In order to reset the state on a failure response, `EditableFieldView.isPendingSave` is now an externally-managed state.  All uses of `EditableFieldView` are now responsible for setting it to `true` when kicking off an operation and `false` when reaching a terminal state, be it a success or failure.

After a successful DUOS update, we also refresh the snapshot's policy members since its readers have likely changed.

https://user-images.githubusercontent.com/79769153/229539672-4d172517-bd2d-4b60-ba84-da362437b5a4.mov

My [developer environment](https://jade-ok.datarepo-dev.broadinstitute.org/) is up to date with these changes. I've made `DataRepoAdmins@dev.test.firecloud.org` stewards of the following resources if you'd like to play around:
- https://jade-ok.datarepo-dev.broadinstitute.org/snapshots/a822d3c5-8002-481c-8a13-50bf913ee614
- https://jade-ok.datarepo-dev.broadinstitute.org/datasets/e9899643-4ba7-4c73-9101-13c50ad79788


src/components/EditableFieldView.tsx
```diff
@@ -187,27 +188,29 @@ function EditableFieldView({
                     id="outlined-basic"
                     className={classes.textInput}
                     data-cy={`${cypressFieldNameFormatted}-text-field`}
-                    variant="outlined"
                     defaultValue={updatedFieldValue}
+                    disabled={isPendingSave}
```
okotsopoulos, line 192: I was able to disable the text input for non-markdown `EditableFieldView`s while a save is pending, but for markdown `EditableFieldViews` (snapshot and dataset descriptions) we use a `SimpleMdeReact` component which does not allow for its textarea to be disabled.

I'm open to other ideas on how to handle that case but seeing as descriptions are modified as part of non-flight patch operations which return quickly, I elected to leave this alone… we will still get save and cancel buttons disabled while a save is pending.

src/components/dataset/overview/DatasetOverviewPanel.jsx
```diff
@@ -173,7 +172,7 @@ function DatasetOverviewPanel(props) {
             {renderTextFieldValue('Dataset ID', dataset.id)}
           </Grid>
           <Grid item xs={4}>
-            {renderTextFieldValue('Default Billing Profile ID', dataset.defaultBillingProfile)}
+            {renderTextFieldValue('Default Billing Profile ID', dataset.defaultProfileId)}
```
okotsopoulos, line 175: Drive-by bug fix: this was the wrong name for this field, so a dataset's default billing profile ID was always blank.

I think if this were Typescriptified we would have caught this. I tried to convert this file but it was taking longer than I wanted to, so I walked it back. I'd be interested in talking to the team about how y'all approach Typescript conversions.

src/components/dataset/overview/DatasetOverviewPanel.jsx
```diff
@@ -284,37 +284,35 @@ function DatasetOverviewPanel(props) {
           </Grid>
         </Grid>
       </TabPanel>
-      {true && (
```
okotsopoulos, line 287: Drive-by consolidation.

src/components/EditableFieldView.test.tsx
```diff
@@ -144,4 +179,22 @@ describe('Plain text field in EditableFieldView', () => {
         );
       });
   });
+  it('pending save disables save and cancel buttons', () => {
```
okotsopoulos: I also wanted to verify that the text input was disabled, but I couldn't figure out how to target Cypress appropriately. Suggestions welcome!

src/components/snapshot/overview/SnapshotOverviewPanel.tsx
```diff
@@ -104,11 +107,17 @@ function SnapshotOverviewPanel(props: SnapshotOverviewPanelProps) {
...
+              <EditableFieldView
+                fieldValue={snapshot.duosFirecloudGroup?.duosId}
+                fieldName="DUOS ID"
+                canEdit={isSteward}
+                isPendingSave={pendingSave.duosDataset}
+                updateFieldValueFn={(text: string | undefined) =>
+                  dispatch(updateDuosDataset(snapshot.id, text))
+                }
+                useMarkdown={false}
+                infoButtonText="Link with a DUOS dataset ID to automatically add DAC approved users as snapshot readers"
```
snf2ye: As you mentioned in your PR, the DUOS id can take up to 10 seconds to update. As a user, I tend to assume that if something takes a while that it's actually broken. We could add a note to this info button text that it's expected for the update to take a little bit of time. Thoughts?

src/components/EditableFieldView.test.tsx
snf2ye: I'm not sure if this is the best way, but looks like adding this check works:
`cy.get('[data-cy=phs-id-text-field]').children().children('input').should('be.disabled');`

snf2ye (approved):
Thanks for also handling the loading state for the other dataset and snapshot fields with this PR!

src/components/snapshot/overview/SnapshotOverviewPanel.tsx
okotsopoulos: That's a good idea, I can make that update!

src/components/EditableFieldView.test.tsx
okotsopoulos: Thank you for playing with this, I'll give that a shot!

src/components/EditableFieldView.tsx
samanehsan: I agree, the save occurs pretty quickly so I don't think users would be able to edit the description through the UI while the save is pending

samanehsan (approved):
This looks good to me! Thanks for the demo earlier! 

Also, this is out-of-scope for this PR, but I think it makes sense to put optional fields such as the DUOS ID, PHS ID and consent code at the bottom since they could be empty

src/components/EditableFieldView.test.tsx
okotsopoulos: I was able to get this a little tighter, but this suggestion got me on the right track :)

okotsopoulos:
> Also, this is out-of-scope for this PR, but I think it makes sense to put optional fields such as the DUOS ID, PHS ID and consent code at the bottom since they could be empty

@samanehsan I agree that the ordering is a bit haphazard, I think Lou is working on some new mocks for the snapshot overview page, I couldn't immediately find the ticket for this though!

PR #1275, opened 2022-12-29
[DR-2878] Fixed react dom validation error for snapshot overview

lu-c:
MUI's typography element defaults the object type to `<p>`. On the snapshot overview page, this will trigger an error for reactDomValidation, because <div>s should not nest inside `<p>`. The built in fix is to set the object type to a `<span>` type.

Easy fix <3


okotsopoulos (approved):
I completely missed this PR when it was raised!  But I encountered this same issue when working with the snapshot overview page, and can confirm that this change fixes the error.

nmalfroy (approved):
nice!

PR #1288, opened 2023-02-16
DR-2893 Make the Job Modal URL Addressable

nmalfroy:
This PR does a few things:
- It makes it so that a query parameter can be used to make a job description modal open (via `?expandedJob=<jobId>`)
- This is nice because now you can refresh the page and the job info is still there
- Note: the job doesn't have to be on the page to load
- It also allows linking from a failed job.  This PR also adds the ability to do that.

Here's the interaction.  Note the url

jobmodal

Note: newer designs for the error toast:
**Plain**
image

**With a job link**
image

**Handle large messages**
image

You can see this in action at: 
https://jade-nm.datarepo-dev.broadinstitute.org


src/components/JobView.test.tsx
```diff
@@ -44,19 +45,22 @@ const initialState = {
   },
   user: _.cloneDeep(initialUserState),
   query: _.cloneDeep(initialQueryState),
+  router: { location: {} },
 };
 
 beforeEach(() => {
-  const mockStore = createMockStore([]);
+  // sinon.replace(history, 'push', (path: any) => console.log('!!!!!', path));
```
okotsopoulos: I think these were left in from debugging!

src/components/table/JobTable.tsx
```diff
@@ -45,17 +47,23 @@ interface IProps extends WithStyles<typeof styles> {
...
 const JobTable = withStyles(styles)(
-  ({ classes, jobs, handleFilterJobs, loading, searchString, refreshCnt }: IProps) => {
+  ({ classes, jobs, handleFilterJobs, loading, searchString, expandedJob, refreshCnt }: IProps) => {
     const statusMap: any = {
       succeeded: { icon: `fa fa-circle-check ${classes.statusIconSuccess}`, label: 'Completed' },
       running: { icon: `fa fa-rotate fa-spin ${classes.statusIconSuccess}`, label: 'In Progress' },
       failed: { icon: `fa fa-circle-xmark ${classes.statusIconFailed}`, label: 'Failed' },
     };
 
+    // If the expanded job is not on the page, add the modal explicitly at the top level
```
okotsopoulos: I appreciate the comment here!

src/modules/notifications.tsx
```diff
@@ -17,14 +30,14 @@ export function showNotification(err) {
...
-    content: <Toast errorMsg={message} status={status} />,
+    content: <Toast errorMsg={message} status={status} jobId={jobId} />,
     insert: 'top',
     container: 'top-right',
     animationIn: ['animate__animated', 'animate__slideInRight'],
     animationOut: ['animate__animated', 'animate__slideOutRight'],
     dismiss: {
       duration: 5000,
-      touch: false,
+      touch: true,
```
okotsopoulos: Want to confirm the expected behavior here: I'm seeing that all toast notifications now must be clicked on to close, including those without associated jobIds.

I think I like this better, we've heard from users that their toast disappeared before they could save the information.  But the behavior is not 100% clear.  Would it be possible to give toasts an "x" icon that you click to close?  That may be OOS.

src/components/Toast.tsx
```diff
@@ -0,0 +1,79 @@
...
+    <Card className={classes.card}>
+      <CardContent className={classes.content}>
+        <div className={classes.cardBody}>
+          <Error className={classes.icon} />
+          <div className={classes.text}>{errString}</div>
+        </div>
+      </CardContent>
+      {jobId && (
+        <CardContent className={classes.jobLinkContainer}>
+          <Link to={`/activity?expandedJob=${jobId}`}>Click to view details</Link>
```
okotsopoulos: What do you think about having this open out in a new tab?  For both snapshot creation and dataset creation, I could see it being frustrating to the user to lose their state on clicking.  Even better would be to not need to leave the page at all, if we could have the same job details modal from the activity page pop up here without going to the activity page itself.

I also wonder about the text of this link, I wonder if it would be helpful to expose the job ID here instead.

okotsopoulos (approved):
I have a feeling that direct links to TDR jobs in the UI will make our users very happy!

I left a few suggestions in the comments, but nothing major enough to hold my approval.

src/components/JobView.test.tsx
nmalfroy: derp.  good catch!

src/modules/notifications.tsx
nmalfroy: Yeah, I was trying to style it and it was driving me nuts so I just left it on.  I think it's better this way.  I can certainly style...I want to sync with Lou about the job link anyways

src/components/Toast.tsx
nmalfroy: Oh interesting...yeah, I think the modal should work.  I'll check in with Lou.  My only fear is getting into modal overload

src/components/job/JobResultModal.tsx
```diff
@@ -75,22 +81,26 @@ function JobResultModal({
...
-  const [seeMore, setSeeMore] = useState({ open: false });
-
+  const expandedJob = location.query?.expandedJob;
   useEffect(() => {
-    if (seeMore.open) {
+    if (expandedJob === id) {
       dispatch(getJobResult({ id }));
     }
-  }, [dispatch, seeMore, id]);
+  }, [dispatch, expandedJob, id]);
```
snf2ye: (Just a thought, I'm good with it as is!)
If we set the seeMore variable here, we could avoid the copied logic `expandedJob === id` and instead use the seeMore Variable on[ L125](https://github.com/DataBiosphere/jade-data-repo-ui/pull/1288/commits/85ea5b63c945fe15483eba7b9943988140d96269#diff-d36e518aa80e39d550b639fd716ed3e555ab7425c330083a7aefd35952f74da4R125). 
```
const [seeMore, setSeeMore] = useState(false);
  useEffect(() => {
    setSeeMore(expandedJob === id);
    if (seeMore) {
      dispatch(getJobResult({ id }));
    }
  }, [dispatch, expandedJob, id, seeMore]);
```

src/modules/notifications.tsx
snf2ye: This is an interesting change - I could imagine a user trying to copy/paste the error message, but that would actually dismiss the message, right? I think an "x" icon would be a good idea.

snf2ye (approved):

okotsopoulos (approved):
Thanks so much for the additional work here on error modal behavior.  I love it!

I tried out a snapshot creation job that I knew would fail to induce an error modal with job details.  Keeping everything on the same page in this case made it easy to correct my request error and resubmit without having to reenter the request :)

snf2ye (approved):
These changes look great, thank you!

PR #1291, opened 2023-02-22
[DR-2928] Collapse all resource role accordions by default

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2928

Previously, dataset and snapshot steward accordions were expanded by default when clicking into the resource summary's roles and memberships tab.  Now, we close them all by default, since the stewards list can sometimes be quite long.

My [dev environment](https://jade-ok.datarepo-dev.broadinstitute.org/) has these changes deployed.  Here's a screen recording of the new behavior:

https://user-images.githubusercontent.com/79769153/220751417-be7f29ba-5021-4800-83d8-150d1897707d.mov

We considered also including a member count in the collapsed accordion headers -- i.e. 'Stewards (34)' -- but decided against it because such a count would only reflect direct members and could confuse our users.  Specifically, if a Firecloud group is a direct member of a resource's policy group, we won't be able to tell how many actual members it has.  We could revisit this in the future, though.

### DIFF
--- a/src/components/dataset/DatasetAccess.tsx
+++ b/src/components/dataset/DatasetAccess.tsx
@@ -70,7 +70,6 @@ function DatasetAccess(props: DatasetAccessProps) {
           typeOfUsers="Stewards"
           canManageUsers={canManageStewards}
           removeUser={removeUser(DatasetRoles.STEWARD)}
-          defaultOpen={true}
         />
       </Grid>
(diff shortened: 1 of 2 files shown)


fboulnois (approved):

nmalfroy (approved):
love it!  nice and simple :)

PR #1294, opened 2023-02-27
DR-2951 Bring dependabot-tagged dependencies up to date

nmalfroy:
Note, we'll eventually want to find a way to remove the pins (e.g. MUI, prettier) but for now, this should quiet dependabot.

The best thing from a review standpoint is to look over the packages.json but other than that, I built with this locally and was able to run and everything looks ok (in addition to automated tests)

A few notes on pinned versions:
- newer versions of `prettier` seem to format the code in a way that is incompatible with `eslint` ([ticket to unpin](https://broadworkbench.atlassian.net/browse/DR-2953))
- newer versions of material ui (MUI) cause our styling to not respect theming or straight up break the UI ([ticket to unpin](https://broadworkbench.atlassian.net/browse/DR-2952))
- newer version of @cypress/react changed their package structure and now our unit tests won't run because they require `@cypress/react/plugins/react-scripts` and the entire `@cypress/react/plugins` directory was removed with no explanation that I could find in newer versions.  It's possible that this is part of a larger cypress upgrade that is needed ([ticket here](https://broadworkbench.atlassian.net/browse/DR-2954))


snf2ye (approved):

okotsopoulos (approved):
Thanks for consolidating these updates and explaining the reasoning behind pinning a few of them!

PR #1310, opened 2023-03-07
[DR-2736] Snapshot readers can export snapshots to Terra

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2736

Previously in the TDR UI:
- All snapshot exports also synced destination workspace policy groups as snapshot readers.
- Thus, only snapshot stewards were allowed to initiate snapshot exports from the UI.  This is because snapshot readers do not hold the needed snapshot action to modify its readers, so could not complete the full export workflow.

Behavior now:
- Snapshot stewards can choose whether or not to sync workspace policy groups as snapshot readers during an export to Terra.  By default, this is enabled to remain consistent with existing behaviors.
- Snapshot readers can export snapshots to Terra, but can't sync the workspace policy groups as snapshot readers.
- If a snapshot export is performed without policy group syncing, the workspace members will not have full access to the snapshot's files as they would if they were full snapshot readers.

Export as a snapshot steward:

https://user-images.githubusercontent.com/79769153/223541054-f1a978b0-e7b0-4cd5-b1d3-c2210a42b8b7.mov

Export as a snapshot reader:

https://user-images.githubusercontent.com/79769153/223541156-489c3b4e-771a-4281-9c47-91b88ca3e827.mov

My [dev environment](https://jade-ok.datarepo-dev.broadinstitute.org/snapshots) is up to date with these changes.

### DIFF
--- a/src/components/snapshot/overview/SnapshotExport.test.tsx
+++ b/src/components/snapshot/overview/SnapshotExport.test.tsx
@@ -26,7 +26,7 @@ describe('SnapshotExport', () => {
   beforeEach(() => {
     mockStore = createMockStore([]);
   });
-  it('Test has export button', () => {
+  it('Test steward can export with permission sync', () => {
     const initialState = {
       snapshots: {
         snapshot,
@@ -48,8 +48,11 @@ describe('SnapshotExport', () => {
       </Router>,
     );
     cy.get('[data-cy="export-snapshot-button"]').should('contain.text', 'Export snapshot');
+    cy.get('[data-cy="tdr-sync-permissions-checkbox"] input')
+      .should('not.be.disabled')
+      .and('be.checked');
   });
-  it('Test has export button disabled', () => {
+  it('Test reader can export without permission sync', () => {
     const initialState = {
       snapshots: {
         snapshot,
@@ -70,7 +73,10 @@ describe('SnapshotExport', () => {
         </Provider>
       </Router>,
     );
-    cy.get('[data-cy="export-snapshot-button"]').should('be.disabled');
+    cy.get('[data-cy="export-snapshot-button"]').should('contain.text', 'Export snapshot');
+    cy.get('[data-cy="tdr-sync-permissions-checkbox"] input')
+      .should('be.disabled')
+      .and('not.be.checked');
   });
   it('Test preparing snapshot', () => {
     const preparingSnapshotState = {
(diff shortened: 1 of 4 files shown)


src/components/snapshot/overview/SnapshotExport.test.tsx
```diff
@@ -48,8 +48,11 @@ describe('SnapshotExport', () => {
       </Router>,
     );
     cy.get('[data-cy="export-snapshot-button"]').should('contain.text', 'Export snapshot');
+    cy.get('[data-cy="tdr-sync-permissions-checkbox"] input')
```
okotsopoulos, line 51: Note that we're using a MUI Checkbox and need to drill down to the underlying `input` before performing our assertions.

src/components/snapshot/overview/SnapshotExport.tsx
```diff
@@ -73,16 +72,19 @@ function SnapshotExport(props: SnapshotExportProps) {
     terraUrl,
     userRoles,
   } = props;
-  const exportResponseManifest =
-    exportResponse &&
-    exportResponse.format &&
-    exportResponse.format.parquet &&
-    exportResponse.format.parquet.manifest;
+  const exportResponseManifest = exportResponse?.format?.parquet?.manifest;
```
okotsopoulos, line 75: This looked like a place that we could leverage optional chaining instead.

nmalfroy: good call!

nmalfroy (approved):
looks great!

src/components/snapshot/overview/SnapshotExport.tsx
```diff
@@ -92,31 +94,25 @@ function SnapshotExport(props: SnapshotExportProps) {
...
   };
 
-  const gsPathsCheckbox = !isProcessing ? (
-    <Checkbox checked={exportGsPaths} onChange={handleExportGsPathsChanged} />
-  ) : (
-    <Checkbox checked={exportGsPaths} disabled />
-  );
-
-  const canExport = userRoles.includes(SnapshotRoles.STEWARD);
-  const tooltipMessage =
```
okotsopoulos, line 102: I did get rid of the tooltip, since it was specific to permission syncing.  This info is covered in helper text for the permission syncing checkbox.

(If we leave the tooltip out, we can also close this: https://broadworkbench.atlassian.net/browse/DR-2876)

snf2ye (approved):

PR #1312, opened 2023-03-09
DR-2956 - Pin Dependabot versions & always run status checks

snf2ye:
# Pin Versions
[Versions to pin from previous PR](https://github.com/DataBiosphere/jade-data-repo-ui/pull/1294) 
- newer versions of prettier seem to format the code in a way that is incompatible with eslint ([ticket to unpin](https://broadworkbench.atlassian.net/browse/DR-2953))
- newer versions of material ui (MUI) cause our styling to not respect theming or straight up break the UI ([ticket to unpin](https://broadworkbench.atlassian.net/browse/DR-2952))
- newer version of @cypress/react changed their package structure and now our unit tests won't run because they require @cypress/react/plugins/react-scripts and the entire @cypress/react/plugins directory was removed with no explanation that I could find in newer versions. It's possible that this is part of a larger cypress upgrade that is needed ([ticket here](https://broadworkbench.atlassian.net/browse/DR-2954))

# Always run status checks
With the following change, we need status checks to run in order to merge the PR (without manually bypassing requirements), so let's just always run the tests.
image


.github/dependabot.yml
```diff
@@ -5,3 +5,11 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+  ignore:
+  # newer versions of prettier seem to format the code in a way that is incompatible with eslint (DR-2953)
```
okotsopoulos, line 9: Thank you so much for adding these comments and their tickets!

okotsopoulos (approved):

nmalfroy (approved):
looks good!

.github/dependabot.yml
snf2ye: Thanks @nmalfroy for writing up these explanations!

PR #1323, opened 2023-03-15
DR-2956 - Turn off auto rebase for dependabot PRs

snf2ye:
Right now, when one dependabot PR changes, they all kick off tests. When 10 are kicked off at once, tests never complete.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
   directory: "/"
   schedule:
     interval: daily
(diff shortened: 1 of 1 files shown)


okotsopoulos (approved):
Thanks for the discussion in [Slack](https://broadinstitute.slack.com/archives/C01VBGH431S/p1678899992604489), especially the reminder that our current settings require a rebase before merging.  I was concerned originally that we might forget to rebase if we turned auto-rebase off.

nmalfroy (approved):

PR #1376, opened 2023-04-06
DR-2940: Switch data view to use new view dataset data endpoint

snf2ye:
Changes deployed to my dev env: https://jade-sh.datarepo-dev.broadinstitute.org/

### Background
Previously, we only supported viewing row data in the UI for GCP datasets by directly querying BigQuery. This PR is the first step towards removing code that directly queries Big Query. Instead, we will handle all requests for row data through the TDR API.

This PR does not include retrieving column data to facilitate snapshot create. This will be replaced in a future ticket. 

### The Changes
- total and filtered row count now populated in data view paging
- Azure dataset data view now works (but no filtering enabled)
- GCP dataset data view should not look any different, but powered by TDR API instead of BQ

https://user-images.githubusercontent.com/13254229/231267892-a1227798-95e4-4b97-bac8-8d465c703ddd.mov


samanehsan (approved):
This looks awesome!

src/components/table/LightTable.tsx
```diff
@@ -191,18 +187,10 @@ function LightTable({
   const handleChangeRowsPerPage = async (event: any) => {
     const newRowsPerPage = parseInt(event.target.value, 10);
     await dispatch(changeRowsPerPage(newRowsPerPage));
-    if (pageBQQuery) {
-      pageBQQuery();
-    }
   };
 
-  // Once we no longer need to support BQ Querying,
-  // we can remove the async/await call and pageBQQuery()
```
okotsopoulos, line 200: Nice to see all the cleanup you were able to do here!

src/components/table/LightTable.tsx
```diff
@@ -419,19 +401,13 @@ function LightTable({
           </Dialog>
         </Paper>
       )}
-      {loading && !rows?.length && (
-        <LoadingSpinner
-          delay={delay}
-          delayMessage="For large datasets, it can take a few minutes to fetch results. Thank you for your patience."
```
okotsopoulos, line 425: Curious why you removed the delay messages? Did it make for clunky state transitions?

okotsopoulos (approved):
This looks great! One small question but not a blocker to merge.

src/components/table/LightTable.tsx
snf2ye: The delay variable is populated by a BigQuery-specific function that polled the status of the BQ query. So, since we're no longer getting this data directly from BQ and we don't yet have a way to check in on the status of a TDR request, I removed this.

src/components/table/LightTable.tsx
snf2ye: It was so fun 😄

PR #1382, opened 2023-04-10
[DR-2978] Expose dataset builder

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2859 filled out dataset builder's links to Terra documentation…
https://broadworkbench.atlassian.net/browse/DR-2978 and expose it in the UI.

Now that we have the updated Terra support documents needed to launch our new dataset builder, I've added the links and removed the feature flag functionality surrounding it so that all users will see the "Create Dataset" button when in the "Datasets" tab.

https://user-images.githubusercontent.com/79769153/230997047-01d0c314-7bdc-41b2-bac0-a03328ebb11d.mov

While here:
- Converted affected files to functional Typescript components
- Fixed two bugs in `HomeView` - a typo and an issue where the header title was not changing with the selected tab.
- Fully removed the feature flag functionality: if we want to leverage something like this in the future we’d like to add something similar to how Terra UI displays work-in-progress features to opt into.  The existing functionality was undocumented and could only be accessed via the browser's Javascript console.

My [developer environment](https://jade-ok.datarepo-dev.broadinstitute.org/datasets) is up to date with this PR if you'd like to play around!


src/components/HomeView.jsx
```diff
@@ -1,137 +0,0 @@
...
-    fontSize: '1.5rem',
-  },
-});
-
-function HomeView({ classes, location }) {
-  const prefixMatcher = /\/[^/]*/;
-  const [searchString, setSearchString] = useState('');
-  const tabValue = prefixMatcher.exec(location.pathname)[0];
-  let tableValue;
-  let pageTitle = 'Terra Data Respository';
```
okotsopoulos, line 73: Fixed typo in migration.

src/components/HomeView.tsx
```diff
@@ -0,0 +1,138 @@
...
+  } else if (tabValue === '/snapshots') {
+    pageTitle = 'Snapshots';
+    tableValue = <SnapshotView searchString={searchString} />;
+  } else if (tabValue === '/activity') {
+    pageTitle = 'Activity';
+    searchable = false;
+    tableValue = <JobView searchString={searchString} />;
+  }
+
+  const pageHeader =
```
okotsopoulos, line 95: Fixed bug in migration: previously we declared `pageHeader` too soon, so every tab but Datasets didn't see their title change on tab select.

nmalfroy: oh awesome!

src/components/dataset/schemaCreation/DatasetSchemaCreationView.tsx
```diff
@@ -265,17 +265,19 @@ const DatasetSchemaCreationView = withStyles(styles)(({ classes, dispatch, profi
                 <li>
                   <a
                     className={clsx(classes.jadeLink, classes.helpListLink)}
-                    href="#"
+                    href="https://support.terra.bio/hc/en-us/articles/13817865732123-Overview-Your-TDR-dataset-schema"
```
nmalfroy, line 268: 🎉

nmalfroy (approved):
Excellent!  And thanks for Typescriptifying!

src/components/HomeView.tsx
```diff
@@ -0,0 +1,138 @@
...
+    pageTitle = 'Datasets';
+    tableValue = <DatasetView searchString={searchString} />;
+  } else if (tabValue === '/snapshots') {
+    pageTitle = 'Snapshots';
+    tableValue = <SnapshotView searchString={searchString} />;
+  } else if (tabValue === '/activity') {
+    pageTitle = 'Activity';
+    searchable = false;
+    tableValue = <JobView searchString={searchString} />;
+  }
```
snf2ye, line 93: I think the best practice here would be to use react state and hooks to manage these values. I would think we would need this in the case that the location changed, but clearly, it works without it, so I think I'm just rusty on what happens without react hooks 😅 . At any rate, I still think using react state/hooks is a pattern that I'd like to see us use. 
```suggestion
  const [searchString, setSearchString] = useState('');
  const prefixMatcher = /\/[^/]*/;
  const [tabValue, setTabValue] = useState(prefixMatcher.exec(location.pathname)?.[0]);
  const [pageTitle, setPageTitle] = useState('Terra Data Repository');
  const [searchable, setSearchable] = useState(true);
  const [tableValue, setTableValue] = useState(<div />);
  
 useEffect(() => {
     setTabValue(prefixMatcher.exec(location.pathname)?.[0]);
   }, [location.pathname]);
 
 useEffect(() => {
     if (tabValue === '/datasets') {
         setPageTitle('Datasets');
         setTableValue(<DatasetView searchString={searchString} />);
         setSearchable(true);
      } else if (tabValue === '/snapshots') {
          setPageTitle('Snapshots');
          setTableValue(<SnapshotView searchString={searchString} />);
          setSearchable(true);
       } else if (tabValue === '/activity') {
          setPageTitle(' 'Activity');
          setSearchable(false);
          setTableValue(<JobView searchString={searchString} />);
        }
  }, [tabValue, pageTitle, tableValue, searchable]);
```

okotsopoulos: Thanks for this suggestion! I was porting over what previously existed but it seemed a little funny to me, I'll give this a shot.

snf2ye: Discussed this offline - this is getting out of scope, so disregard this comment.

snf2ye (approved):

src/components/HomeView.tsx
okotsopoulos: For future reference, when I tried to make this change I was seeing some clunky state transitions:

https://user-images.githubusercontent.com/79769153/231244879-1c023355-6ce7-4b91-a152-e1f260f91838.mov

PR #1386, opened 2023-04-11
DR-2996 More UI library updates

nmalfroy:
Of note:
- Major upgrade of Axios (small test change needed...it's a little strange I needed to do it but I'm not losing sleep over it since the code is going to go away very soon)
- A bunch of minor version tagged by dependabot (and then some)


cypress/integration/errors.spec.js
```diff
@@ -67,23 +67,17 @@ describe('test error handling', () => {
...
 
-  it('displays loading message', () => {
+  it.only('displays loading message', () => {
     cy.route({
       method: 'POST',
       url: 'bigquery/v2/projects/**/queries',
       status: 200,
       response: { jobComplete: false, jobReference: { jobId: 'jobId' } },
     }).as('getQueryResults');
-    cy.route({
```
okotsopoulos, line 77: Just checking, why did this get removed? Was it not being used?

okotsopoulos (approved):

cypress/integration/errors.spec.js
nmalfroy: For some reason, it wasn't ever firing. I suspect that something changed with how axios mocks stuff.  I didn't want to spend too much on this since this is behavior that will all go alway with Shelby's changes (and I manually tested and it works just fine)

samanehsan (approved):

PR #137, opened 2019-12-06
DR-638 Michael's big table refactor

myessail:

PR #1403, opened 2023-04-27
DR-3017 Properly display timestamp values in data view

snf2ye:
### GCP
Timestamp data is not currently correctly displayed in the UI. We are displaying the long value rather than the human-readable formatted timestamp. This is a side effect of the switch from directly querying BQ to using the results from the view data TDR endpoint ([original changes here](https://github.com/DataBiosphere/jade-data-repo-ui/pull/137)).  
image
### Azure
This problem is more pervasive with Azure datasets, as it also impacts "date" and "datetime" data types. However, it does not impact array field because we process them differently in the backend. 

### GCP
When we were directly querying BQ, we were handling the same problem of converting from long to a timestamp. So, we were able to re-use this logic and format any timestamp data. Addtionally, we are now displaying the time zone for the timestamp. All timestamp data is ingested in UTC, but the UI converts it into the user's local timezone. 
image
### Azure
With a change to the backend (https://github.com/DataBiosphere/jade-data-repo/pull/1451), we will not need to do any further processing on the frontend for date/datetime/timestamp data types. 
image

### GCP
I've added a table to our integration test environment that encompasses almost all of our data types. This should help avoid errors like this in the future when we can easily see all of the different data types displayed in one table. Additionally, I've added a test that confirms that we are seeing the same timestamp both on ingest and displayed in the UI. 
### Azure
[We have a ticket](https://broadworkbench.atlassian.net/browse/DR-3024) to add an integration test for Azure datasets. Manual testing has done on my personal dev env: https://jade-sh.datarepo-dev.broadinstitute.org/ using the Azure_V2F_GWAS_Summary_Statistics dataset and the all_data_types table. 

## Note
* The changes to the integration testing table have been made to int-4 and int-5. (I have confirmed that tests pass on both int-4 & int-5 with these changes)
*  You can see the UI changes and the new table on my dev environment:
https://jade-sh.datarepo-dev.broadinstitute.org/login
Dataset: V2F_GWAS_Summary_Statistics
Table: all_data_types


src/modules/bigquery.js
```diff
@@ -9,69 +9,6 @@ export default class BigQuery {
...
-    }
-
-    return res;
-  };
-
-  commaFormatted = (amount) => new Intl.NumberFormat('en-US').format(amount);
-
-  significantDigits = (amount) =>
-    new Intl.NumberFormat('en-US', { maximumSignificantDigits: 3 }).format(amount);
-
```
snf2ye, line 74: This code is no longer used after [my last UI change](https://github.com/DataBiosphere/jade-data-repo-ui/pull/1376) to use our TDR API endpoint rather than directly querying BigQuery. 

This is the chunk of code that previously correctly formatted timestamp data. So, I'm borrowing some of the logic here, but applying it to the resulting data from the TDR endpoint.

src/reducers/query.ts
```diff
@@ -87,16 +87,32 @@ export default {
...
-        const columns = action.payload.columns.map((column: ColumnModel) => ({
+        const columns: TableColumnType[] = action.payload.columns.map((column: ColumnModel) => ({
           name: column.name,
           dataType: column.datatype,
           arrayOf: column.array_of,
           allowSort: !column.array_of,
           allowResize: true,
           width: columnsByName[column.name]?.width || TABLE_DEFAULT_COLUMN_WIDTH,
         }));
+        // We only need to re-format row data of type timestamp
```
snf2ye, line 104: Previously, we also reformatted data of type "integer" and "float". I no longer think there is need to do this.

I believe the main reason was to round to 3 significant digits so that numbers would be displayed without overflow in the UI. However, our new(er) LightTable component already handles overflow well. Additionally, we have had a user specifically request for all significant digits to be displayed in the UI.

cypress/integration/queryBuilder.spec.js
```diff
@@ -44,6 +44,29 @@ describe('test query builder', () => {
...
+    beforeEach(() => {
+      cy.get('[data-cy=selectTable]').click();
+      cy.get('[data-cy=menuItem-all_data_types]').click();
+    });
+
+    it('correctly displays timestamps', () => {
+      // Click the header to make sure we don't sort or trigger an error
+      cy.get('[data-cy=columnHeader-timestamp_column]').scrollIntoView().click({ force: true });
+
+      cy.get('[data-cy=cellValue-timestamp_column-0]').should('have.text', '5/31/2023, 9:02:40 PM');
```
nmalfroy: Sweet!  I don't know if it's necessarily worth it but I wonder if we might ever want to experiment with mocking browser info to see what happens when it's set to, say, display dates in `DD/MM/YYYY`.  Note: this isn't a thing I think we should do for this particular PR but just wanted to raise a potential thing we might want to do

src/reducers/query.ts
```diff
@@ -87,16 +87,32 @@ export default {
...
+        const timestampColumns: TableColumnType[] = [];
+        columns.forEach((col: TableColumnType) => {
+          if (col.dataType === 'timestamp') {
```
nmalfroy: We actually have a constant for this now (`TableDataType`). It's autogenerated from the OpenApi spec
```suggestion
          if (col.dataType === TableDataType.Timestamp) {
```

src/reducers/query.ts
```diff
@@ -87,16 +87,32 @@ export default {
...
+            timestampColumns.push(col);
+          }
+        });
+        const rows = action.payload.queryResults.data.result.map((row: any) => {
+          timestampColumns.forEach((col: TableColumnType) => {
+            if (col.arrayOf) {
+              row[col.name] = row[col.name].map((v: number) => new Date(v * 1000).toLocaleString());
```
nmalfroy: What happens if v is null or 0?  I think we might need a check here no?

nmalfroy (changes requested):
Thanks for doing this!  Couple of small suggestions

src/reducers/query.ts
```diff
@@ -87,16 +87,32 @@ export default {
...
+            } else {
+              row[col.name] = new Date(row[col.name] * 1000).toLocaleString();
```
okotsopoulos: Is there a way to break out this formatting so that it can be reused between the array and singleton cases?

cypress/integration/queryBuilder.spec.js
```diff
@@ -44,6 +44,29 @@ describe('test query builder', () => {
     });
   });
 
+  describe('timestamps are displayed as expected', () => {
```
okotsopoulos, line 47: Just curious, is there a reason we want / need to test timestamp display in an integration test rather than a unit test?

okotsopoulos (commented):
This looks good to me, I left a few comments but no showstoppers.  When your dev environment is working I'd be interested in poking around there :)

src/reducers/query.ts
snf2ye: Oh, great point. Yes, if it was null, I was just setting it to the Unix epoch. I also added handling for empty arrays in the LightTable code.
Fixed now:
image

cypress/integration/queryBuilder.spec.js
snf2ye: The integration test for the timestamp value is testing to make sure the combination of the output from the API, the formatting by the reducer and the displayed value all amount to the correct value. I've added a unit test for the handling of regular and null values in the Light table for both singleton and array values. But, the LightTable only passes through the value from the reducer, so we're not testing the format of the timestamp. If we added a unit test for the reducer, we could get most of the testing value without the integration test.  However, that's a bigger undertaking than I wanted to do right now because I think we may want to consider using jest instead of cypress for that.

snf2ye: I also think there is value is making sure the timestamp that we're ingesting matches the timestamp we're displaying in the UI since it's going through a couple of layers of conversion.

okotsopoulos: Well-reasoned. No objections here :)

src/components/table/LightTable.tsx
```diff
@@ -209,12 +209,20 @@ function LightTable({
...
+      (empty)
+    </span>
+  );
 
   const handleRepeatedValues = (values: Array<string>, columnName: string) => {
-    const cleanValues = values
-      .map((v) => (_.isNil(v) ? handleNullValue() : `${v}`))
-      .map((v, i) => <span key={`val-${i}`}>{v}</span>);
+    /* eslint-disable indent */
+    const cleanValues = _.isEmpty(values)
```
nmalfroy, line 220: Hmm...so if I read this right, an empty array would show up as `null`?  Do we want that behavior?

nmalfroy (changes requested):
Looks good but I wanted to just double check that I understand about the empty array handling

src/components/table/LightTable.tsx
snf2ye: No, we have special handling in the component for null values. Without this change, array fields were just blank in the table, but with this change, we now show the placeholder "(empty)" text. 
image

okotsopoulos (approved):

src/components/table/LightTable.tsx
nmalfroy: Oh, sorry, so what you are showing is the case where the array is `[null]`, as opposed to `[]`? (Just cause it says 1 item)

snf2ye: We transform null values to instead display "(empty)" -- (You can see the code for this just a couple lines up from this)

src/components/table/LightTable.tsx
```diff
@@ -209,12 +209,20 @@ function LightTable({
...
+      ? [handleNullValue()]
+      : values
+          .map((v) => (_.isNil(v) ? handleNullValue() : `${v}`))
```
snf2ye, line 223: @nmalfroy  - Actually, I think I was misunderstanding your question - If it's [null], then this code should handle it.

nmalfroy: Ah perfect

nmalfroy (approved):
LGTM!

PR #1407, opened 2023-05-01
[DR-2969] Display DUOS group's last synced time

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-2969

If a snapshot is linked to a DUOS dataset, we now display the DUOS Firecloud's last synced time in the snapshot summary panel.

We have some work coming down the pike to reorganize snapshot summary information in a more logical way within our UI, but for now this surfaces helpful information about the freshness of DUOS Firecloud groups.

Note: we're using the moment library to parse the timestamp value, ex. "an hour ago".  It might be a bit confusing to the user to see a moment on a tab that hasn't been refreshed in awhile, since a snapshot in the UI doesn't get updated periodically in the background (only on refresh or as part of action handling).  But this is consistent with our other moment usages, so I think it's okay.

**Demonstration**

My [developer environment](https://jade-ok.datarepo-dev.broadinstitute.org/snapshots/efb89ba7-8c64-4a7e-ae7a-31faea93d115) is up to date with these changes.

https://user-images.githubusercontent.com/79769153/235708277-de3b59d9-2d77-4d74-b972-1bb47938a892.mov

### DIFF
--- a/src/components/snapshot/overview/SnapshotOverviewPanel.test.tsx
+++ b/src/components/snapshot/overview/SnapshotOverviewPanel.test.tsx
@@ -14,26 +14,64 @@ const duosFirecloudGroup = {
   lastSynced: '2022-04-04T18:53:45.158566Z',
 };
 
-const snapshot = {
-  id: 'uuid',
-  name: 'Test snapshot',
-  description: 'Test description',
-  createdDate: '2022-04-04T18:53:45.158566Z',
-  source: [
-    {
-      dataset: {
-        id: 'datasetId',
-        name: 'SourceDataset',
-        storage: [
-          { cloudResource: 'bigquery', region: 'us-east4' },
-          { cloudResource: 'firestore', region: 'us-east4' },
-          { cloudResource: 'bucket', region: 'us-east4' },
-        ],
+function createSnapshot(_duosFirecloudGroup?: any): any {
+  return {
+    id: 'uuid',
+    name: 'Test snapshot',
+    description: 'Test description',
+    createdDate: '2022-04-04T18:53:45.158566Z',
+    source: [
+      {
+        dataset: {
+          id: 'datasetId',
+          name: 'SourceDataset',
+          storage: [
+            { cloudResource: 'bigquery', region: 'us-east4' },
+            { cloudResource: 'firestore', region: 'us-east4' },
+            { cloudResource: 'bucket', region: 'us-east4' },
+          ],
+        },
+      },
+    ],
+    duosFirecloudGroup: _duosFirecloudGroup,
+  };
+}
+
+function createInitialState(snapshot?: any): any {
+  return {
+    snapshots: {
+      pendingSave: {
+        consentCode: false,
+        description: false,
+        duosDataset: false,
+      },
+      snapshot,
+      snapshotPolicies: [
+        {
+          name: 'steward',
+          members: ['steward@gmail.com'],
+        },
+        {
+          name: 'reader',
+          members: ['reader@gmail.com'],
+        },
+        {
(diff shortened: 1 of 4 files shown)


src/components/snapshot/overview/SnapshotOverviewPanel.tsx
```diff
@@ -111,20 +111,30 @@ function SnapshotOverviewPanel(props: SnapshotOverviewPanelProps) {
...
+            />
+          </Grid>
+          <Grid item xs={8}>
+            {snapshot.duosFirecloudGroup && (
+              <Grid item xs={4}>
+                <Typography variant="h6">DUOS Users Last Synced:</Typography>
+                <Typography data-cy="snapshot-duos-last-synced">
+                  {snapshot.duosFirecloudGroup.lastSynced
+                    ? moment(snapshot.duosFirecloudGroup.lastSynced).fromNow()
+                    : 'Never'}
```
okotsopoulos, line 134: We shouldn't find ourselves here: when we create a Firecloud group, we sync its contents.  But figured it was an easy safeguard to put in just in case.

src/components/snapshot/overview/SnapshotOverviewPanel.test.tsx
```diff
@@ -69,26 +107,29 @@ const initialState = {
   },
 };
 
+function mockSnapshotOverviewPanel(snapshot: any): any {
```
snf2ye, line 73: Just making sure I understand - you're creating this function instead of using the "BeforeEach" method so that you can pass in a different snapshot for a given test, right?

snf2ye (approved):

src/components/snapshot/overview/SnapshotOverviewPanel.test.tsx
okotsopoulos: Exactly!  But I am open to a different way to achieve this.

src/components/snapshot/overview/SnapshotOverviewPanel.test.tsx
```diff
@@ -14,26 +14,64 @@ const duosFirecloudGroup = {
...
-    {
-      dataset: {
-        id: 'datasetId',
-        name: 'SourceDataset',
-        storage: [
-          { cloudResource: 'bigquery', region: 'us-east4' },
-          { cloudResource: 'firestore', region: 'us-east4' },
-          { cloudResource: 'bucket', region: 'us-east4' },
-        ],
+function createSnapshot(_duosFirecloudGroup?: any): any {
```
nmalfroy: I don't know if you want the `_` here right? I think that this indicates that it's an ignored field but you're using it below.  Also, I think there's a `DuosFirecloudGroupModel` type...not sure if it's the same as what's passed in here

samanehsan (approved):

nmalfroy (approved):
Looks great!  I have the one question about the test code/typescript stuff

src/components/snapshot/overview/SnapshotOverviewPanel.test.tsx
okotsopoulos: Oh, good points… I put the underscore because I had a naming conflict with a higher-level `duosFirecloudGroup` object, but I forgot that it is meant to represent an ignored field.  At any rate the conflict probably points to a code smell.

And yes, I think I can use that type here!  I might be able to use more types throughout, now that you mention it.

okotsopoulos: Glad you directed me back here -- I found a near-duplicate block of test set-up code that I could refactor out :)

PR #1417, opened 2023-05-13
DR-2993: Utilize new TDR API Column Stats endpoint for Dataset filtering on snapshot create

snf2ye:
Changes deployed here: https://jade-sh.datarepo-dev.broadinstitute.org/ 
Please try to break 😄 
Ticket: https://broadworkbench.atlassian.net/browse/DR-2993
_____________________________________________________

https://github.com/DataBiosphere/jade-data-repo-ui/assets/13254229/23604c6c-233d-4145-b513-58ec180106eb

### The Changes
* Utilize Column stats TDR endpoint rather than directly querying BQ
* If we don filter on a field due to data type or array status, a message has been added to indicate why the user cannot filter on the filed. 
image

* We now display both the filtered and total count for checkbox text fields
image

**There are 3 types of fields we can filter on:**
1. Text checkbox
 * Handles options with 30 or fewer unique text options
 * We track the total count and the filtered count of each of these unique fields
 * On applying a filter, we re-fetch the column stats and update the counts of each value. We continue to display all options if it has been filtered out of the data, we just set the filtered count to 0.
2. Free text
* We fetch all unique values upon opening the filter window. We do not update these on applying the filter. (This is a change -[ see comment](https://github.com/DataBiosphere/jade-data-repo-ui/pull/1417#discussion_r1195043891)) 
3. Numeric
* We fetch the min and max numeric value for the field upon opening the filter window. We do not update this on applying the filter. 

### Approach
I've tried to keep the changes as minimal as possible to help with the review process.  

Potential follow-on changes:
* I would like to add a follow-on PR that updates these components to typescript and into functional components. I have these changes in a separate branch. 
* Any remaining BQ references - Like removing scope? Anything else?
* Handle array fields (https://broadworkbench.atlassian.net/browse/DR-1476)
* Handle additional data types - date/time fields, boolean

### Helpful resources
[Updated understanding of useEffect](https://react.dev/learn/you-might-not-need-an-effect)


snf2ye:
Query builder tests are failing because we have not yet merged the API change to add the column stats endpoint. Manual testing of these functionalities works.

src/components/common/data/DataView.tsx
```diff
@@ -99,7 +99,6 @@ function DataView({
 }: DataViewProps) {
   // Can be removed after DR-2483
   const showPanels = panels.length > 0;
-  // Only used for Direct BQ Query
   const isDatasetFiltered = filterStatement.length > 0;
```
snf2ye, line 102: I originally added this comment, but I'm not sure why - It's not accurate as far as I can tell.

src/components/dataset/data/sidebar/filter/FreetextFilter.jsx
```diff
@@ -35,54 +33,20 @@ export class FreetextFilter extends React.PureComponent {
...
-      tableName,
-      token,
-      filterStatement,
-      joinStatement,
-    );
-
-    const transformedResponse = this.transformResponse(response);
-    this.setState({
-      options: transformedResponse,
-    });
```
snf2ye, line 85: I don't think this is necessary. 
By filtering down the responses, we remove the ability to add more options after filtering. 
Let's say there are 3 options: 'a' 'b', & 'c'. (This is a bad example because it would actually use the checkbox option, which DOES keep the original values in the list even after filtering with an updated count according to the filter set. Anyway...)
* let's say the user selects 'a' and then hits "apply filter"
* Then, the user decides they also want to include 'b'. If we kept this code in place, the the option 'b' would no longer be present. Yes, they could just type in 'b' if they knew what they were looking for, but otherwise, they would have to clear the filter and then go back and select 'a' and 'b'.

src/components/dataset/data/sidebar/DataViewSidebarItem.jsx
```diff
@@ -25,13 +25,9 @@ export class DataViewSidebarItem extends React.PureComponent {
   static propTypes = {
     classes: PropTypes.object,
     column: PropTypes.object,
-    dataset: PropTypes.object,
     filterData: PropTypes.object,
-    filterStatement: PropTypes.string,
     handleChange: PropTypes.func,
-    joinStatement: PropTypes.string,
     tableName: PropTypes.string,
-    token: PropTypes.string,
```
snf2ye, line 34: We no longer need to pass these items through:
- Dataset & filterStatement have been moved to redux state
- joinStatement is not needed for the column stats endpoint
- token was only used when directly talking to bq

src/components/dataset/data/sidebar/filter/CategoryFilter.jsx
```diff
@@ -36,7 +39,7 @@ export class CategoryFilter extends React.PureComponent {
             data-cy={`categoryFilterCheckbox-${name}`}
           />
         }
-        label={`${name} (${count})`}
+        label={`${name} (${formattedCount})`}
```
snf2ye, line 42: Slight update to the original functionality - this now displays both the filtered and total count. It originally just showed the filtered count. 
With change:
image

src/components/dataset/data/sidebar/filter/CategoryFilterGroup.jsx
```diff
@@ -45,17 +45,24 @@ export class CategoryFilterGroup extends React.PureComponent {
...
-      const count = parseInt(_.get(values, name, '0'), 10);
+    const { column, values, table, filterMap, originalValues } = this.props;
+    const checkboxes = originalValues?.map((val) => {
+      const totalCount = val.count;
+      let filteredCount = 0;
+      values.map((filteredValue) => {
+        if (filteredValue.value === val.value) {
+          filteredCount = filteredValue.count;
+        }
+      });
```
snf2ye, line 56: (Explanation because this wasn't entirely clear to me before getting into this code)
For checkbox text fields (so, text fields with fewer than 30 options), we track both the full set of options ('originalValues') and the filtered set of options ('values'). The only thing this buys us is that we can display the count of the number of times we see a given value and update this count after filtering. 
We display all of the options (all of the 'originalValues'), even if filtering has eliminated an option from the data view. If an option has been eliminated, we display the count as 0. We get the filtered count from the list of filtered 'values'.

src/components/dataset/data/sidebar/filter/CategoryWrapper.tsx
```diff
@@ -0,0 +1,89 @@
...
+        getColumnStats(
+          ResourceType.DATASET,
+          dataset.id,
+          tableName,
+          column.name,
+          ColumnDataTypeCategory.TEXT,
+        ),
+      );
+    }
+  }, [dispatch, dataset.id, tableName, column.name, filterStatement, refreshColumnStats]);
```
snf2ye: Note: We originally talked about only fetching the column stats after a user clicks on a particular field in the filter panel in order to avoid extra trips to BQ/Synapse.  I decided to leave this functionality as is for now since it would be a change in the user experience(& a little complicated to get exactly right - for example, what happens when you filter? Do you then update all fields? Or again only update on field click?), but am glad to revisit this change. 

@nmalfroy  - I'd be curious for your thoughts on this after clicking around in the updated UI.

samanehsan (approved):
This looks great! I did some manual testing on your deployment and the filtering worked as expected 😄 It would be great to add tests for the individual components as well, like the `RangeFilter`

src/components/dataset/data/sidebar/filter/CategoryWrapper.tsx
nmalfroy: I think it works in the UI.  My main concern is for azure datasets (longer query times) with a lot of columns.  I think it's fair to keep the current behavior and see how it performs once you roll out the Azure backend

src/components/dataset/data/sidebar/filter/FreetextFilter.jsx
nmalfroy: I agree.  this is a good change.  I do worry about not debouncing though.  I want to make sure we're not hammering BQ

nmalfroy: But I'll add that click testing, it seems to not be overly aggressive so I think it's fine

src/reducers/query.ts
```diff
@@ -146,6 +154,49 @@ export default {
...
           polling: { $set: false },
         }),
+      [ActionTypes.COLUMN_STATS_FAILURE]: (state, action: any) =>
+        immutable(state, {
+          error: { $set: true },
+          errMsg: { $set: action.payload },
+          polling: { $set: false },
+        }),
+      [ActionTypes.COLUMN_STATS_TEXT_SUCCESS]: (state, action: any) => {
+        const values = action.payload.queryResults.data.values;
```
nmalfroy: Might make sense to just do this now so it doesn't keep squawking at us
```suggestion
        const { values } = action.payload.queryResults.data;
```

nmalfroy (approved):
Looks great! And click testing works awesome (awesomely?)

nmalfroy (approved):
Looks great! And click testing works awesome (awesomely?)

nmalfroy (approved):
Looks great! And click testing works awesome (awesomely?)

Also thanks for the typescriptification!

src/components/dataset/data/sidebar/filter/FreetextFilter.jsx
snf2ye: It should only ever hit BQ (via API) on first load and on actually hitting the "apply filter" button. With this, do you think we're at risk for hammering the BQ?

src/reducers/query.ts
snf2ye: Ah, thanks!

src/components/dataset/data/sidebar/filter/CategoryWrapper.tsx
snf2ye: Ah, you called it. With Azure, the get stats calls are pretty slow and when you have a bunch of columns, the filter becomes unusable. In the example dataset from our integration tests that has 50+ columns, it takes 2 **minutes** to retrieve all of the stats. So, I think we need to make this change before we enable it for Azure.

snf2ye: And, actually, it's not just Azure. It totally stalls the UI for the "transcript_consequence" table in GCP. https://jade-sh.datarepo-dev.broadinstitute.org/datasets/5ed80a11-9c2e-402c-a583-fe4c85ead4b4

nmalfroy: Ah yeah, I see what you mean...So basically, we track expanded columns and any time a filter changes, only call the requests for the expanded columns?

snf2ye:
>  It would be great to add tests for the individual components as well, like the `RangeFilter`

@samanehsan  - Good point! I've added component tests for RangeFilter and CategoryWrapper with a particular focus on the logic I've added to these two components.

src/components/dataset/data/sidebar/filter/CategoryWrapper.tsx
```diff
@@ -47,9 +45,41 @@ function CategoryWrapper({
         ),
       );
     }
-  }, [dispatch, dataset.id, tableName, column.name, filterStatement, refreshColumnStats]);
+  }, [dispatch, dataset.id, tableName, column.name, column.isExpanded]);
 
-  const { originalValues, values } = column;
+  // Run when filter has been updated
+  useEffect(() => {
```
nmalfroy: This is great.  It made a very noticeable difference in the UI

src/components/dataset/data/sidebar/filter/CategoryFilter.jsx
```diff
@@ -33,7 +33,7 @@ export class CategoryFilter extends React.PureComponent {
           <Checkbox
             checked={checked}
             onChange={this.handleChange}
-            value={name}
+            value={name ?? "null"}
```
nmalfroy: Does it make sense to have this be the italic `(empty)` like we show in the table?

nmalfroy (approved):
Awesome!  Just a couple of comments that could totally be followons

src/components/dataset/data/sidebar/filter/CategoryWrapper.test.tsx
```diff
@@ -0,0 +1,189 @@
...
+describe('Expanded column after another column has updated - Freetext', () => {
+  beforeEach(() => {
+    const state = createState({
+      ...baseStringColumn,
+      isExpanded: true,
+      filterHasUpdated: true,
+      values: [{ value: 'a', count: 1 }],
+      originalValues: [
+        { value: 'a', count: 1 },
+        { value: 'b', count: 2 },
```
samanehsan, line 111: I'm not sure I understand, why does this test contain repeated values?

src/components/dataset/data/sidebar/filter/CategoryFilter.jsx
samanehsan: I agree, I think "Empty" or "None" would be clearer!

samanehsan (approved):
Thanks for adding in the tests! (I just had a question about one of them)

src/components/dataset/data/sidebar/filter/CategoryFilter.jsx
snf2ye: Great! I will cover this in my next UI PR!

src/components/dataset/data/sidebar/filter/CategoryWrapper.test.tsx
snf2ye: We don't have anything on the UI side that checks if the values are unique (We expect the API to return a list of unique values). For this test, I'm trying to just get more than 30 elements in the values list so that it switches from the checkbox display to the freetext dropdown. But, it is unclear- I'll make them unique.

samanehsan: Oh that makes sense now!

PR #1427, opened 2023-05-22
[DR-3031] Remove unused xlsx dependency

samanehsan:
https://broadworkbench.atlassian.net/browse/DR-3031

The `xlsx` library was added in one of the early Data Repo [commits](https://github.com/DataBiosphere/jade-data-repo-ui/commit/1fadea723327a1ad8fb24d069469d2f1dddd4fe5#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) and is no longer used, so this PR removes the dependency. This resolves the dependabot alert: https://github.com/DataBiosphere/jade-data-repo-ui/security/dependabot/69

When I ran `npm remove xlsx`, it generated the `package-lock.json` file using lockfileVersion 2. This formatting difference resulted in a diff of over 30k lines! Running `npm i --lockfile-version 1 --package-lock-only` changed the version back to lockfileVersion 1. Do we want to upgrade to lockfileVersion 2 at some point?


nmalfroy (approved):
Ah, this looks a lot closer. thanks!

nmalfroy:
I wonder what makes it pick between 1 and 2 and why mine was leaving it at 1? maybe node versions?

okotsopoulos (approved):
Thanks so much for revisiting and for removing this dependency!

PR #1439, opened 2023-06-01
DR-3025 Azure Snapshot Create

snf2ye:
Changes are deployed: https://jade-sh.datarepo-dev.broadinstitute.org/
Ticket: https://broadworkbench.atlassian.net/browse/DR-3025
________________________

https://github.com/DataBiosphere/jade-data-repo-ui/assets/13254229/9624fa90-82bb-4b3b-8ab1-94d40554c77d

Included Changes:
- Enable Azure snapshot create in the UI
- null fields in filter will be displayed as "(empty)" to match the data view
- Add azure dataset to integration load script
- Update V2F dataset to include more complex relationships and assets
- Put slider selector in middle of range when minValue = maxValue
image

Note: Tests are actually now passing if you click into the latest results.
_____________

Remaining todos:

- [x] Don't allow switching tables in data view until current table is loaded (Gets into strange state)
- [x] Get queryBuilder test working with both GCP and Azure versions of the V2F dataset (need to rename the azure version of the dataset)
- [x] Is there a way to share code between the queryBuilder and azureQueryBuilder test? 
- [x] Find an application deployment to use for loading the azure dataset in int-4 and int-5 (I've just been using my personal app)
- [x] Make sure GCP queryBuilder tests still pass with the new rows/assets for the V2F dataset
- [x] Re-upload GCP and azure datasets into int-4 
- [x] Re-upload GCP and azure datasets into int-5


tools/ui_integration/files/azure_billing_profile.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "bbc1684f-4b5e-45bf-8a50-64b029333731",
+  "profileName": "ui_integration_data_azure",
+  "biller": "direct",
+  "description": "Azure billing profile for UI integration test data",
+  "cloudPlatform": "azure",
+  "tenantId": "efc08443-0082-4d6c-8931-c5794c156abd",
+  "subscriptionId": "71d52ec1-5886-480a-9d6e-ed98cbf1f69f",
+  "resourceGroupName": "TDR",
+  "applicationDeploymentName": "shelby"
```
snf2ye: This currently creates a billing profile in my application deployment on azure. We should an azure application deployment to use with UI integration tests. Or is there another good one to use that already exists and shouldn't get cleaned out?

cypress/integration/azureQueryBuilder.spec.js
```diff
@@ -0,0 +1,249 @@
+describe('test query builder on Azure', () => {
```
snf2ye: This test is basically the same as the queryBuilder test, but testing against the Azure version of the same dataset. I haven't yet figured out how to share the actual tests. Any ideas?

cypress/integration/queryBuilder.spec.js
```diff
@@ -236,13 +249,15 @@ describe('test query builder', () => {
...
       cy.contains('Share Snapshot').should('be.visible');
+      const regexString = `/FROM (${cloudPlatform.datasetName}).variant /`;
       cy.window()
         .its('store')
         .invoke('getState')
         .its('snapshots')
         .its('snapshotRequest')
         .its('joinStatement')
-        .should('match', /FROM (V2F_GWAS_Summary_Stats|V2F_GWAS_Summary_Statistics).variant /);
+        .should('match', regexString);
```
snf2ye: This test is now failing - I need to figure out how to have a variable in the regex or find another workaround.

nmalfroy:
Just a few pre-review things I noticed:
first: this is so exciting!
second: I don't know if it's always been the case but when I was clicking around, I noticed that range filters where min and max are the same behave kind strangely (both drag handles are at the start).  I think we should maybe do special handling (e.g. not even show a slider in that case since the values are equal to a slider wouldn't make sense.). 
third: I also noticed that the loading mask doesn't really cover the screen.  That might be for a separate ticket though

snf2ye:
@nmalfroy  Thanks for taking a look! I'll take a look at the case where min=max on the range filter.

>  third: I also noticed that the loading mask doesn't really cover the screen. That might be for a separate ticket though

The loading mask over which component?

cypress/integration/queryBuilder.spec.js
snf2ye: Fixed

tools/ui_integration/files/azure_billing_profile.json
snf2ye: New deployment added for int ui!

nmalfroy (approved):
this looks great!  can't wait to have this available to folks!

samanehsan:
**Update:** Shelby mentioned in stand-up that this was an existing issue and will be ticketed separately

----------------
I was trying out the snapshot creation and noticed the rows don't update when I remove one of the drop-down filters. Here's what I see when I first select one of the options in the `variant_id` dropdown:

Screen Shot 2023-06-12 at 10 41 45 AM

And when I remove the filter I see the same results:
Screen Shot 2023-06-12 at 10 43 21 AM

When I apply a phenotype filter to this, the `variant_id` filter that I removed shows up again! 👻
Screen Shot 2023-06-12 at 10 44 04 AM

samanehsan (approved):
This looks great! :tada:

snf2ye:
@samanehsan - This is an existing bug. I took a stab at fixing it and it's not straightforward due to the way we've built our filter map. So, I've put this in a ticket: https://broadworkbench.atlassian.net/browse/DR-3085
Thank you for thoroughly testing and finding this!

snf2ye:
Tests are failing due to a difference in the row ordering on int-4 vs. int-5. (Tests still should be passing on int-4). I working on a fix.

PR #1468, opened 2023-08-03
[DR-3123] Replace react-scripts test server with vite

samanehsan:
The react-scripts package is no longer being maintained and contains a security vulnerability in the semver package. It is only used for testing with cypress, so this PR replaces it with the `vite-dev-server`: https://www.npmjs.com/package/@cypress/vite-dev-server/v/5.0.5


samanehsan:
Replaced by #1475

PR #1475, opened 2023-08-09
[DR-3123] Migrate to vite

samanehsan:
This PR updates the UI to use Vite instead of create react app (`react-scripts` is one of the dependencies, which has not been maintained for over a year and contains a security vulnerability: https://broadworkbench.atlassian.net/browse/DR-3123).

Some things to note:
- Vite has a built-in hot reloader
- The route proxying is defined in vite.config.js (so setupProxy is no longer needed)
- When running locally, the url needs to use "localhost" instead of "127.0.0.1" in order for log in to work

I also updated Cypress from version 9 -> 12, which supports the latest version of Vite. This involved:
- Replacing `cypress.json` with `cypress.config.ts`
- Removing the `cypress/plugins/index.js` file
- Adding a `cypress/support/component-index.html` file (where the component test html gets rendered)
- Updating the syntax in the e2e tests (e.g. cy.route is now cy.intercept) and some of the unit/component tests


samanehsan:
The test datasets/snapshots used by the e2e tests exist but for some reason are not accessible to the test user (I can't access them with my dev account either). I created new ones in int-6 as a temporary work-around so I could run/fix the e2e tests locally, so they should pass once we can sort out the issue with int-4 and int-5!

samanehsan:
The SonarCloud failure can be resolved by generating a hash for our fontawesome kit "https://kit.fontawesome.com/316fe749fb.js", but I get a 403 error when I try to do that (I used this website: https://www.srihash.org/). Does anyone have access? Or should I create a new custom kit?

okotsopoulos:
@samanehsan that FontAwesome question rings a bell, I'm not sure if this is possible :(

From @snf2ye in February '23:
> Fontawesome supports the use an SRI hash with versions <5. We are using a fontawesome kit v6. Fontawesome made a product choice to no longer support SRI hashes ([source](https://github.com/FortAwesome/Font-Awesome/issues/16691)).

https://broadworkbench.atlassian.net/browse/DR-2872

samanehsan:
Thanks @okotsopoulos! That's helpful context! I'll mark that script as excluded

samanehsan:
I think the e2e tests aren't running because it waits on `localhost:3000` but the vite dev server is using `127.0.0.1:3000`. I'll see if there's a way to change the server to use localhost.

src/components/common/TextContent.test.tsx
```diff
@@ -13,7 +13,7 @@ const markdownText = '# Lorem Ipsum \n\n[Link](https://broadinstitute.org)';
 
 const htmlText = '<a href="www.badsite.com">A bad day</a>';
 
-describe('WithoutStylesMarkdownContent', () => {
+describe('TextContent', () => {
```
s-rubenstein, line 16: Why the name change here? Is it sonar related? This seems less descriptive.

src/components/common/WithoutStylesMarkdownContent.test.tsx
```diff
@@ -39,9 +39,9 @@ describe('WithoutStylesMarkdownContent', () => {
         </Provider>
       </Router>,
     );
-    cy.get('div#__cy_root > div').children().should('have.length', 1);
-    cy.get('div#__cy_root > div > p').should('contain.text', 'A bad day');
-    cy.get('div#__cy_root > div > p').children().should('have.length', 0);
+    cy.root().should('have.length', 1);
+    cy.root().get('p').should('contain.text', 'A bad day');
```
s-rubenstein, line 43: Just as a note, do you think this sort of test could maybe be simplified by doing a findByText on the page? That is what Terra-UI does

src/constants/index.ts
```diff
@@ -1,5 +1,4 @@
 import { CloudPlatform } from 'generated/tdr';
-import _ from 'lodash';
```
s-rubenstein, line 2: I'm surprised you can remove the import from lodash without any other changes to the file. Do you have any linting for unused imports?

src/reducers/snapshot.ts
```diff
@@ -352,13 +352,13 @@ export default {
...
 
-      [LOCATION_CHANGE]: (state, action: any) => {
+      [LOCATION_CHANGE]: (state: SnapshotState, action: any) => {
         // Don't reset state if the only change was query parameters
         const path = action.payload?.location?.pathname;
         if (path === state.lastPath) {
           return immutable(state, {
-            lastPath: { $set: path },
-          });
+            lastPath: { $set: path as string },
```
s-rubenstein, line 360: I'm surprised you have to do the casts here. Does typescript not type interpolate correctly for these?

s-rubenstein (commented):
I'm noticing a lot more casting than was there before - is this a result of the libraries changing? Casting like that can sometimes be a code smell, so I am curious if you know where its coming from.

src/components/common/TextContent.test.tsx
samanehsan: `WithoutStylesMarkdownContent` is the name of a different component, I thought it was a copy-paste error

src/reducers/snapshot.ts
samanehsan: Yeah, weirdly there were a couple of places where the type couldn't be inferred and without making it explicit the server wouldn't run

s-rubenstein: Ah, in this particular case it might be that it thinks path is `string | undefined` given the ?s. Would you file a ticket to look into the specific casts? My guess is that they are resolvable using typescript but there may be some code structure changes needed.

src/store/middleware.ts
```diff
@@ -7,7 +7,7 @@ export const sagaMiddleware = createSagaMiddleware();
 const middleware = [sagaMiddleware];
 
 /* istanbul ignore next */
-if (process.env.NODE_ENV === 'development') {
+if (import.meta.env.DEV) {
```
samanehsan, line 10: This env variable is provided by Vite by default: https://vitejs.dev/guide/env-and-mode.html#env-variables

samanehsan:
The "process not defined" issue causing the e2e tests to fail was due to to the version of jsondiffpatch we were using. Updating to the latest version fixed the issue! 🎉 

- https://github.com/reduxjs/redux-devtools/issues/937#issuecomment-962520662
- https://github.com/benjamine/jsondiffpatch/issues/315

src/constants/index.ts
samanehsan: Good point! Our eslint rules don't include unused imports so I created a ticket for that: https://broadworkbench.atlassian.net/browse/DR-3176

src/reducers/snapshot.ts
samanehsan: Here's the ticket! https://broadworkbench.atlassian.net/browse/DR-3177

src/components/common/WithoutStylesMarkdownContent.test.tsx
samanehsan: Ah oops I missed this comment! It turns out `findByText` requires another dependency so I'm inclined to leave it out for now if that's okay

Dockerfile.direct
```diff
@@ -10,15 +10,15 @@ ARG TDR_OPEN_API_YAML_LOCATION
...
-FROM node:14.0-buster as build
+FROM node:16.20.2-buster as build
 # Copy the local code
 COPY . /
 # Copy the generated code
 COPY --from=codegen /local /
 # Run the build
 RUN export DISABLE_ESLINT_PLUGIN=true && \
   npm ci && \
-  npm run build-no-code-gen --production
```
s-rubenstein, line 21: Is this flag removed?

s-rubenstein (approved):
LGTM

Dockerfile.direct
samanehsan: The `--production` flag isn't needed because `vite build` uses production by default

package.json
```diff
@@ -24,6 +24,10 @@
   "dependencies": {
     "@babel/runtime": "^7.21.5",
     "@codemirror/lang-javascript": "^6.1.5",
+    "@cypress/vite-dev-server": "^5.0.5",
```
petesantos: I believe there are some deps that can be moved to `devDependencies`. This isn't a big deal, but more for `package.json` hygiene

Feel free to do this in a follow up PR

package.json
```diff
@@ -48,20 +51,19 @@
...
     "animate.css": "^4.1.1",
     "awesome-debounce-promise": "^2.1.0",
-    "axios": "^1.3.5",
-    "babel-plugin-inline-react-svg": "^2.0.2",
+    "axios": "^1.4.0",
     "classlist-polyfill": "^1.2.0",
     "clsx": "^1.1.0",
     "connected-react-router": "^6.9.3",
-    "cypress": "^9.7.0",
+    "cypress": "^12.17.3",
```
petesantos: This can be moved to `devDependencies`

package.json
```diff
@@ -97,25 +98,27 @@
     "tss-react": "^4.8.4",
     "typescript": "^5.0.4",
     "validator": "^13.9.0",
+    "vite": "^4.4.9",
```
petesantos: I believe `vite` can be moved to `devDepenendcies`

petesantos (approved):

PR #1482, opened 2023-09-08
[DR-2849][DR-3179][DR-3087] Select DUOS dataset from dropdown and more!

nmalfroy:
This PR does a few things:
- It adds a drop down to the snapshot details page to allow you to select a DUOS dataset

image

- It adds a loading spinner to the dataset and snapshot details pages
- It adds more information on the dataset and snapshot details pages:
- Datasets:
  - For Azure, show the storage account
  - For Azure, don't show the service account
- Snapshots:
  - Is secure monitoring enabled?
  - What is the cloud platform?

Feel free to click around at:
https://jade-nm.datarepo-dev.broadinstitute.org/


src/sagas/repository.ts
```diff
@@ -909,6 +916,26 @@ export function* changePage(page: number): any {
...
+    yield put({
+      type: ActionTypes.GET_DUOS_DATASETS_SUCCESS,
+      datasets: { data: response },
+    });
+  } catch (err) {
+    yield put({
+      type: ActionTypes.GET_DUOS_DATASETS_FAILURE,
+    });
+    showNotification(err);
+  }
```
snf2ye: When I navigate to the snapshot page, I'm getting a 404 error. I think it's because I'm not registered in DUOS. 
I don't think we can expect that every TDR user will also be registered in DUOS. 

`GET https://consent.dsde-dev.broadinstitute.org/api/dataset/v2 404`
image

snf2ye (changes requested):

src/sagas/repository.ts
nmalfroy: oooh interesting.  Good catch.  I'll add something to handle that gracefully.  Perhaps I'll just grey out the drop down in that case (since only users with access to DUOS should be linking DUOS datasets

src/components/snapshot/overview/SnapshotOverviewPanel.tsx
```diff
@@ -150,44 +155,53 @@ function SnapshotOverviewPanel(props: SnapshotOverviewPanelProps) {
...
-                    ...params.inputProps,
-                    value: pendingSave?.duosDataset ? 'Saving...' : params.inputProps.value,
+            {_.isEmpty(duosDatasets) &&
+              renderTextFieldValue('DUOS Dataset', snapshot.duosFirecloudGroup?.duosId)}
+            {!_.isEmpty(duosDatasets) && (
+              <>
+                <Typography variant="h6">
+                  DUOS Dataset:
+                  <InfoHoverButton infoText={duosInfoButtonText} fieldName="DUOS ID" />
+                </Typography>
```
snf2ye: Should we add some text to the info hover button to explain why the dropdown is not showing up?

nmalfroy: Ah good point!  I'll add that

src/sagas/repository.ts
nmalfroy: ok, should be good to go!

src/components/snapshot/overview/SnapshotOverviewPanel.tsx
samanehsan: Should we should show "Duos Users Last Synced" (none) in that case too so that there isn't a space in the middle? I'm also okay with saving that for the bigger UI overhaul that we wanted to do later.

samanehsan (approved):
Love all of these improvements! 🎉

samanehsan:
Do you know how we can get rid of the old 14.x tests under the PR checks?

nmalfroy:
> Do you know how we can get rid of the old 14.x tests under the PR checks?

Ah I was wondering about that.  Lemme see

snf2ye (approved):

src/components/snapshot/overview/SnapshotOverviewPanel.tsx
nmalfroy: We actually show that regardless (only when there is a DUOS dataset linked)

nmalfroy:
I think Cypress was having an issue yesterday which might have led to the extraneous Cypress 14 thing.  I'll ignore for now

PR #1483, opened 2023-09-15
DR-3223: Fix Snapshot Create on paste with return carriage

snf2ye:
The changes: Switch to single quotes to match other filtering commands (And combine these functions since they're now the same); Add return carriage handling to paste into filter operation.

**The Problem**

A user ran into an issue creating a snapshot in the UI. The query for the snapshot create looked like the following example. This causes a "BigQueryException Syntax error - Unclosed string literal" error on snapshot create.

```sql
 SELECT <dataset>.<table>.datarepo_row_id FROM  <dataset>.<table> WHERE  <dataset>.<table>.<column> 
 IN (\"sample1\r\", \"sample2\r\", \"sample3\r\", \"sample1\r\")
```

I still haven't quite figured out how the escaped return carriage "\r" ended up in the query. 

With the the included code changes, we should instead get something that looks like the following example. 

```sql
SELECT <dataset>.<table>.datarepo_row_id FROM  <dataset>.<table>  WHERE  <dataset>.<table>.<column> 
IN ('sample1', 'sample2', 'sample3', 'sample4')
```

Example snapshot create by query with changes:
```sql
SELECT V2F_GWAS_Summary_Statistics.variant.datarepo_row_id FROM V2F_GWAS_Summary_Statistics.variant
WHERE V2F_GWAS_Summary_Statistics.variant.id  IN ('1:98233405:A:G','1:155630619:A:C')
```

### Testing

Changes have been deployed to my dev env: https://jade-sh.datarepo-dev.broadinstitute.org/datasets/5ed80a11-9c2e-402c-a583-fe4c85ead4b4

### DIFF
--- a/cypress/integration/queryBuilder.spec.js
+++ b/cypress/integration/queryBuilder.spec.js
@@ -133,7 +133,7 @@ testPlatforms.forEach((testPlatform) => {
             .its('snapshots')
             .its('snapshotRequest')
             .its('filterStatement')
-            .should('contain', 'ancestry_specific_meta_analysis.ancestry IN ("EU")');
+            .should('contain', "ancestry_specific_meta_analysis.ancestry IN ('EU')");
         });
       });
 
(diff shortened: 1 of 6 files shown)


samanehsan (approved):
Looks good to me!

nmalfroy (approved):
This looks great!

PR #1509, opened 2023-10-06
DR-3100: Cherry-pick dev image to prod/public gcr on merge to dev

snf2ye:
Part of work to move away from two GCRs - https://github.com/DataBiosphere/jade-data-repo/pull/1521
### The Changes
- Output tag during bump action
- Call jade-data-repo's cherry-pick action
- Wait to send slack notification until cherry-pick completes

### Testing
- Testing out running the cherry-pick action by copying the job code over to the lint action and giving it an image tag of an existing tag (rather than a version tag). This successfully ran!
image
image

### Remaining points of failure
- For the UI, I'm having to pull the tag version from the bumper action. From looking at the output from previous runs, I'm pretty sure this is the tag I want, but I could be wrong
- As true with the API, I don't want to run the full dev-image-update action off my PR because versions could get out of wack. But, I think I've tested almost all of the individual components.

### DIFF
--- a/.github/workflows/dev-image-update.yaml
+++ b/.github/workflows/dev-image-update.yaml
@@ -23,6 +23,8 @@ env:
 
 jobs:
   update_image:
+    outputs:
+      ui_image_tag: ${{ steps.bumperstep.outputs.tag }}
     strategy:
       matrix:
         os: [ubuntu-latest]
@@ -43,6 +45,7 @@ jobs:
           path: datarepo-helm-definitions
       - name: 'Bump the tag to a new version'
         uses: broadinstitute/datarepo-actions/actions/main@0.67.0
+        id: bumperstep
         with:
           actions_subcommand: 'bumper'
           role_id: ${{ secrets.ROLE_ID }}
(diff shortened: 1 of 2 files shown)


s-rubenstein (approved):
This looks fine to me!

.github/workflows/dev-image-update.yaml
```diff
@@ -95,12 +98,21 @@ jobs:
     needs: update_image
     uses: ./.github/workflows/helmtagbump.yaml
     secrets: inherit
+  cherry_pick_image_to_production_gcr:
+    needs: update_image
+    uses: DataBiosphere/jade-data-repo/.github/workflows/cherry-pick-image.yaml@1.536.0
```
nmalfroy, line 103: Food for thought: would we want this to switch to an `@latest` version to avoid needing to have to update if we ever make a change to the cherry-pick-image action?

nmalfroy (approved):
Looks good.  The question shouldn't block the PR but more of a philosophical question (and potential followup ticket)

.github/workflows/dev-image-update.yaml
snf2ye: This follows the convention that we use with the datarepo-actions repo. While, yes, it requires an extra PR to update the version, it also allows us to make breaking changes to the jade-data-repo action without immediately breaking the ui action (We can merge an API change, then merge the UI change with no downtime). It is also slightly more secure (Although, given that we control both repos, this really shouldn't be an issue).

nmalfroy: yeah, that's totally fair

PR #1513, opened 2023-10-11
DR-3178 - Add note that *Azure* snapshot import into Terra is not yet fully supported 

snf2ye:
Right now, when an **Azure** snapshot is exported into Terra, only the table names imported into the Terra workspace. This change adds a note just to help alleviate some potential confusion around this partially implemented functionality. Once this feature is implemented in Terra, we should revert this change. 
image


nmalfroy (approved):
lgtm!

samanehsan (approved):

PR #1523, opened 2023-10-19
DR-3164: Handle null values in Free Text Tabular data filtering

snf2ye:
Before this change, the UI when totally blank if there was a null value in the list of freetext value on attempting to filter. See video:

https://github.com/DataBiosphere/jade-data-repo-ui/assets/13254229/9fa77c38-afc0-4887-a3fe-d943095a8517

After:

https://github.com/DataBiosphere/jade-data-repo-ui/assets/13254229/f7916c3b-a63b-4fdf-9bd8-1f6984c3c26b


okotsopoulos (approved):
Thanks for fixing!  Two things:

1. It looks like the linter picked up a failure and a warning.
2. Is this worth the effort to cover in additional unit testing?

README.md
```diff
@@ -86,7 +86,7 @@ gcloud auth login --no-activate
 
 To run end-to-end tests: `npx cypress run` or `npx cypress open` (interactive mode)
 
-To run unit tests: `npx cypress run-ct` or `npx cypress open-ct` (interactive mode)
+To run unit tests: `npx cypress run --component` or `npx cypress open --component` (interactive mode)
```
nmalfroy, line 89: ah interesting.  Didn't realize that this had changed

nmalfroy (approved):
looks good! Same question about the linter

README.md
snf2ye: Yeah, the old command didn't work for me and there was an error message saying to use the new commands. I believe the old ones will soon be retired.

snf2ye:
> Thanks for fixing! Two things:
> 
> 1. It looks like the linter picked up a failure and a warning.
> 2. Is this worth the effort to cover in additional unit testing?

@okotsopoulos  - Lint issues should be fixed now. And yes, good point. I've converted this file to TS and added component testing. It would be good to one day add testing on the filter function, but I'm not quite sure how to proceed there.

PR #1524, opened 2023-10-19
[DDO-3250] Don't update datarepo-helm-definitions's dev anymore

jack-r-warren:
Related to https://github.com/broadinstitute/datarepo-helm-definitions/pull/474, there's no reason to update these files anymore.

I'm not touching the other parts of this repo's GHA, which seem to:
- update datarepo-helm's UI chart with the new versions
- update integration environments with the new UI versions
- propagate UI images across the different GCRs

### DIFF
--- a/.github/workflows/dev-image-update.yaml
+++ b/.github/workflows/dev-image-update.yaml
@@ -79,9 +79,13 @@ jobs:
           gcloud auth configure-docker --quiet
           docker push gcr.io/${{ env.gcr_google_project }}/jade-data-repo-ui:${GCR_TAG}
           gcloud container images \
-          add-tag \
-          gcr.io/${{ env.gcr_google_project }}/jade-data-repo-ui:"${GCR_TAG}" \
-          gcr.io/${{ env.gcr_google_project }}/jade-data-repo-ui:"${{ steps.uiprevioustag.outputs.tag }}" --quiet
+            add-tag \
+            gcr.io/${{ env.gcr_google_project }}/jade-data-repo-ui:"${GCR_TAG}" \
+            gcr.io/${{ env.gcr_google_project }}/jade-data-repo-ui:"${{ steps.uiprevioustag.outputs.tag }}" --quiet
+          gcloud container images \
+            add-tag \
+            gcr.io/${{ env.gcr_google_project }}/jade-data-repo-ui:"${GCR_TAG}" \
+            gcr.io/${{ env.gcr_google_project }}/jade-data-repo-ui:"${GCR_TAG}-develop" --quiet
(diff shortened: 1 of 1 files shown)


.github/workflows/dev-image-update.yaml
```diff
@@ -88,12 +82,6 @@ jobs:
           add-tag \
           gcr.io/${{ env.gcr_google_project }}/jade-data-repo-ui:"${GCR_TAG}" \
           gcr.io/${{ env.gcr_google_project }}/jade-data-repo-ui:"${{ steps.uiprevioustag.outputs.tag }}" --quiet
-      - name: 'Check and edit Helm definition for dev'
-        uses: broadinstitute/datarepo-actions/actions/main@0.67.0
-        with:
-          actions_subcommand: 'deploytagupdate'
-          helm_env_prefix: dev
-          helm_imagetag_update: ui
```
snf2ye, line 96: This actions appears to also tag the images with the new version - https://github.com/broadinstitute/datarepo-actions/blob/master/actions/main/src/deploytagupdate.sh#L11
We can just pull this into a new step (We're moving away from datarepo-actions when possible, anyway)

jack-r-warren: I did not see that, that's a really good catch -- I thought tag just meant git tag

snf2ye: Hmm... Maybe we don't need it. Looks like it already has the version tagged image (e.g. 1.538.0), but we're adding a githash-develop tag to the image as well.. I wonder if we still need that if we're not using it in the dev helm definitions anymore.

snf2ye: Well, we are still using it as the tag. https://github.com/broadinstitute/terra-helmfile/blob/master/values/app/datarepo/live/dev.yaml#L79

One day, we should swap to using the version, but let's just change one thing at a time.

jack-r-warren: It seems like the only effect that we actually want from that section is https://github.com/broadinstitute/datarepo-actions/blob/master/actions/main/src/deploytagupdate.sh#L17 -- it takes `gcr.io/broad-jade-dev/jade-data-repo-ui:${GCR_TAG}` and adds `gcr.io/broad-jade-dev/jade-data-repo-ui:${GCR_TAG}-develop`. I can add this to the command above

snf2ye: That's perfect! Thanks!

snf2ye (approved):

okotsopoulos (approved):

PR #1540, opened 2023-11-13
DR-3313: Point to POST Query Data Endpoints

snf2ye:
### Merge-Order requirement
- https://github.com/DataBiosphere/jade-data-repo/pull/1535 Must be merged first. Tests will fail until this is merged.

### Included Changes
We should not notice any changes in the UI except that the UI can handle larger filter requests when creating a snapshot.

### Testing
While the UI still struggles a little bit, I was able to filter on 1000 items and create a snapshot.

https://github.com/DataBiosphere/jade-data-repo-ui/assets/13254229/6c98685a-eac7-4225-a37b-f633989bf16f

### DIFF
--- a/src/sagas/repository.ts
+++ b/src/sagas/repository.ts
@@ -787,16 +787,19 @@ export function* watchGetDatasetByIdSuccess(): any {
 
 export function* previewData({ payload }: any): any {
   const queryState = yield select(getQuery);
-  const lookupDataRequest = {
+  const queryDataRequest = {
     offset: queryState.page * queryState.rowsPerPage,
     limit: queryState.rowsPerPage,
     sort: queryState.orderProperty === undefined ? DbColumns.ROW_ID : `${queryState.orderProperty}`,
-    sortDirection: queryState.orderDirection === undefined ? SqlSortDirection.Asc : `${queryState.orderDirection}`,
-    filter: queryState.filterStatement === undefined ? '' : `${queryState.filterStatement}`
-  }
+    sortDirection:
+      queryState.orderDirection === undefined
+        ? SqlSortDirection.Asc
+        : `${queryState.orderDirection}`,
+    filter: queryState.filterStatement === undefined ? '' : `${queryState.filterStatement}`,
+  };
   const query = `/api/repository/v1/${payload.resourceType}s/${payload.resourceId}/data/${payload.table}`;
   try {
-    const response = yield call(authPost, query, lookupDataRequest);
+    const response = yield call(authPost, query, queryDataRequest);
     yield put({
       type: ActionTypes.PREVIEW_DATA_SUCCESS,
       payload: {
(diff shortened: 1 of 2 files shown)


src/sagas/repository.ts
```diff
@@ -787,16 +787,19 @@ export function* watchGetDatasetByIdSuccess(): any {
 
 export function* previewData({ payload }: any): any {
```
nmalfroy, line 788: While you're in here, should we rename the method viewData? Or something that doesn't have preview in the name?

nmalfroy (approved):
lgtm!

src/sagas/repository.ts
```diff
@@ -787,16 +787,19 @@ export function* watchGetDatasetByIdSuccess(): any {
...
-  const query = `/api/repository/v1/${payload.resourceType}s/${payload.resourceId}/data/${payload.table}?offset=${offset}&limit=${limit}${sort}${sortDirection}${filter}`;
+  const queryDataRequest = {
+    offset: queryState.page * queryState.rowsPerPage,
+    limit: queryState.rowsPerPage,
+    sort: _.isEmpty(queryState.orderProperty) ? DbColumns.ROW_ID : `${queryState.orderProperty}`,
+    direction:
+      queryState.orderDirection === undefined
+        ? SqlSortDirection.Asc
+        : `${queryState.orderDirection}`,
+    filter: queryState.filterStatement === undefined ? '' : `${queryState.filterStatement}`,
```
samanehsan: I think you can use `_.isEmpty` to check whether `queryState.orderDirection` and `queryState.filterStatement` are undefined as well

src/sagas/repository.ts
```diff
@@ -851,7 +852,7 @@ export function* getColumnStats({ payload }: any): any {
         break;
       }
       case ColumnStatsRetrievalType.RETRIEVE_FILTERED_TEXT: {
-        const filteredResponse = yield call(authGet, filteredQuery);
+        const filteredResponse = yield call(authPost, baseQuery, { filter: filterStatement });
```
samanehsan: Do we still need to check if the `filterStatement` is `undefined`?

src/sagas/repository.ts
snf2ye: Good point!

src/sagas/repository.ts
snf2ye: Oh yeah, good point. Thanks!

src/sagas/repository.ts
snf2ye: I think it would be hard to actually remove all of the references to "preview" data. I think I'd rather leave it all as preview rather than changing only some of the references to the new name, especially since this would only be a code-facing change rather than user-facing.

snf2ye:
I've deployed the latest to my personal namespace and confirmed that it works for large filter queries.

samanehsan (approved):
Thanks for making those changes! 💯

PR #1572, opened 2023-12-20
DR-3408: SqlSortDirection Model no longer generated

snf2ye:
https://broadworkbench.atlassian.net/browse/DR-3408
____________
With a [recent TDR change](https://github.com/DataBiosphere/jade-data-repo/pull/1541), we no longer generate the "SqlSortDirection" model via swagger. Instead we generate "SqlSortDirectionAscDefault" and "SqlSortDirectionDescDefault" models. 

Since we are now setting a default for the query data endpoint in the TDR code, we can pass through an empty sort direction and it will default to "asc." So, we can remove the empty string handling and the reference to the SQLSortDirection model.

A few notes:
**How we sort in the UI**
- By default, we sort by the datarepo_row_id "desc" when viewing data in the UI
- However, we do not show the datarepo_row_id, so the order looks random.

This is not ideal.  We should instead default to sorting by the first non-datarepo-row-id, default to "asc". But, this needs to be communicated well with the users so that they're not confused as to why their data is suddenly appearing different to them. [Added ticket for this - DR-3407](https://broadworkbench.atlassian.net/browse/DR-3407) 

**We missed that the UI code was no longer building**
- Our UI e2e tests have been failing for a week, but we haven't noticed because we do not send nightly test failure alerts to the sack. We should add a slack notification on failure. https://broadworkbench.atlassian.net/browse/DR-3406

### DIFF
--- a/src/sagas/repository.ts
+++ b/src/sagas/repository.ts
@@ -791,7 +791,7 @@ export function* previewData({ payload }: any): any {
     offset: queryState.page * queryState.rowsPerPage,
     limit: queryState.rowsPerPage,
     sort: _.isEmpty(queryState.orderProperty) ? DbColumns.ROW_ID : `${queryState.orderProperty}`,
-    direction: queryState.orderDirection, // query data enndpoint defaults sort direction to asc
+    direction: queryState.orderDirection, // query data endpoint defaults sort direction to asc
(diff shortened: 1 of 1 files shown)


okotsopoulos (approved):
Thanks for fixing and filing those follow-on tickets!

rushtong (approved):
LGTM 👍🏽

PR #1573, opened 2023-12-21
DR-3359: Add Data Access Control Groups to Roles & Membership Tab

snf2ye:
https://broadworkbench.atlassian.net/browse/DR-3359

https://github.com/DataBiosphere/jade-data-repo-ui/assets/13254229/3be3992d-a639-4a99-9d4d-196addae30d3

__________
### Included Changes

- Update ManageUsersView component: Typescript-ify, convert to functional component and clean up (it had state & props it was not using) 
- An update to how we're displaying users when they cannot be removed (see comment on UserList component)
- Add list of auth domains for the snapshot to the redux store
- new DataAccessControlGroup component that shares the UserList component
- Add "Roles" header to existing membership section
- Minor cleanup of SnapshotAccess Component

### Remaining questions around UX

1) [Answer: no, hide if there are no auth domains] Do we want to always show the "Data Access Control Groups" even when there are no auth domains on the snapshot? 
image
2) [Answer: okay] I added the "Roles" header to the second section because the "people" header is really tied to the text box to add new users (And technically wouldn't be shown if the user cannot manage users... but I don't actually think we can get to that state.) Is this okay? Would we want to name it something different?
image


src/components/UserList.tsx
```diff
@@ -60,14 +60,8 @@ class UserList extends React.PureComponent<UserListProps> {
...
         >
-          {canManageUsers && <ManageUsersView removeUser={removeUser} users={users} />}
+          <ManageUsersView removeUser={canManageUsers ? removeUser : undefined} users={users} />
           {users.length === 0 && <Typography className={classes.noUsers}>(None)</Typography>}
-          {!canManageUsers &&
-            users.map((user) => (
-              <Typography noWrap key={user}>
-                {user}
-              </Typography>
-            ))}
```
snf2ye: Note for reviewers: If the user cannot manage users, then instead of just displaying the emails/groups in plain text, we will still display the users in a "chip" format. It just won't have the little "x" to delete the group. 

**Before** this change:
image

**After** this change:
image

okotsopoulos:
Hi @snf2ye --
> 1) Do we want to always show the "Data Access Control Groups" even when there are no auth domains on the snapshot?

We do not.  That's in part because the language around data access control groups is liable to change a bit more and the flux could confuse general users.  For now the primary aim is to give Nate an easy way to validate that his AnVIL snapshots have been set up correctly.  

> 2) I added the "Roles" header to the second section because the "people" header is really tied to the text box to add new users (And technically wouldn't be shown if the user cannot manage users... but I don't actually think we can get to that state.) Is this okay? Would we want to name it something different?

That seems reasonable to me!

src/components/snapshot/DataAccessControlGroup.tsx
```diff
@@ -0,0 +1,30 @@
...
+type DataAccessControlGroupProps = {
+  authDomains: Array<string>;
+};
+
+function DataAccessControlGroup(props: DataAccessControlGroupProps) {
+  const { authDomains } = props;
+  return (
+    <Grid container spacing={1} data-cy="data-access-control-container">
+      <Typography variant="h6">Data Access Control Groups</Typography>
+      <Grid item xs={12} data-cy="data-access-controls">
```
okotsopoulos, line 41: Could we also include subtext that matches the description given in workspaces?  "Collaborators must be a member of all of these [groups](https://bvdp-saturn-dev.appspot.com/#groups) to access this snapshot." ideally with a link to the corresponding Terra environment's groups page.

Screenshot 2024-01-02 at 10 37 47 AM

Often we include details like this when hovering on an info icon, but my preference would be to fully render it as is done in workspaces.

src/components/UserList.tsx
okotsopoulos: Nice!

Will you remind me, what's the case when a user will see this tab, but not be able to manage users?  Is that when a user is a snapshot custodian but not a steward?

okotsopoulos (approved):
Looks good to me, pending the conditional rendering of the DAC groups panel and adding a description!

Thanks also for Typescript-ifying along the way.

src/components/snapshot/DataAccessControlGroup.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+  it('Displays snapshot policies and emails', () => {
+    cy.get('[data-cy="data-access-control-container"]').should(
+      'contain.text',
+      'Data Access Control Groups',
+    );
+    cy.get('[data-cy="data-access-controls"]')
+      .should('contain.text', 'Data Access Controls')
+      .within(() => {
+        cy.get('[data-cy="user-email"]').then((user) => {
+          cy.wrap(user[0]).should('contain.text', 'authdomain1');
```
samanehsan: Could we also add this check for `authdomain2` and `authdomain3`?

samanehsan (approved):
Thanks for making this change! I had one comment about a test and also +1 Olivia's suggestion to add the same description from the workspaces auth domain section!

src/components/UserList.tsx
snf2ye: As far as I can tell, there isn't a scenario where a user can see the membership on dataset/snapshot but not edit it. 

However, I'm using this setting to disable all users' ability to edit the auth domain groups. 

____
More on the first point:
- We only show the Roles & membership tab if the user is a [steward on the snapshot](https://github.com/DataBiosphere/jade-data-repo-ui/blob/develop/src/components/snapshot/overview/SnapshotOverviewPanel.tsx#L121), so they will be able to manage users
- We also reference this code when sharing a snapshot, but in that case it's in "create mode" no matter the user role on the dataset, so they will always be able to manage users on the snapshot they're currently creating.

src/components/snapshot/DataAccessControlGroup.tsx
snf2ye: Added!

PR #1588, opened 2024-01-09
DR-3406: Update slack notify action: Correct branch; Move status check to action call

snf2ye:
### The changes
- Point to the correct dev branch (We use 'develop' in jade-data-repo-ui, not 'dev' like in drs-hub). 
- It appears that the reusable workflow cannot correctly assess the status of the calling workflow. So, we need to move this logic into the calling workflow. 
      -> I don't love the way we're now passing the status: "needs.update_image.result == 'success'" appears to be the right move from some online research but definitely is verbose 😞. We don't have access to the "success()" and "failure()" functions when setting the arguments (at least as far as I can tell).  

### Testing
With change - successfully sending failure notification on job failure:
image

and success notification on job success:
image

### DIFF
--- a/.github/workflows/dev-image-update.yaml
+++ b/.github/workflows/dev-image-update.yaml
@@ -102,6 +102,8 @@ jobs:
     needs: [ update_image, helm_tag_bump, cherry_pick_image_to_production_gcr ]
     uses: ./.github/workflows/notify-slack.yaml
     secrets: inherit
+    if: ${{ !cancelled() }}
     with:
       workflow_name: Dev Image Update
       notify_on_success: true
+      success: ${{ needs.update_image.result == 'success' && needs.helm_tag_bump.result == 'success' && needs.cherry_pick_image_to_production_gcr.result == 'success' }}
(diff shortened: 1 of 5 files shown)


.github/workflows/dev-image-update.yaml
```diff
@@ -102,6 +102,8 @@ jobs:
     needs: [ update_image, helm_tag_bump, cherry_pick_image_to_production_gcr ]
     uses: ./.github/workflows/notify-slack.yaml
     secrets: inherit
+    if: ${{ !cancelled() }}
     with:
       workflow_name: Dev Image Update
       notify_on_success: true
```
pshapiro4broad, line 108: Does this parameter still make sense? In the false case wouldn't the workflow no longer be called?

pshapiro4broad (approved):

rjohanek (approved):
Looks good to me

samanehsan (approved):
I had the same question as Phil, but besides that this looks good! :octocat:

.github/workflows/dev-image-update.yaml
snf2ye: There are three states that we can check for here - success, failure or cancelled. (Cancelled being when the job is canceled manually by a user). So, the !cancelled() check means that it will run either on success on failure. 

It's a little clunky, but it is the [suggested option in the github actions docs](https://docs.github.com/en/actions/learn-github-actions/expressions#always). 

>  If you want to run a job or step regardless of its success or failure, use the recommended alternative: if: ${{ !cancelled() }}

But, given that it's confusing, I think I'll just switch to always() and then cancelled jobs will just get the failure notification.

PR #1593, opened 2024-01-12
[DR-3427] Use Dependabot grouped version updates

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3427
Reference: https://github.blog/changelog/2023-08-24-grouped-version-updates-for-dependabot-are-generally-available/
Slack discussion: https://broadinstitute.slack.com/archives/C01VBGH431S/p1678899992604489

Our Dependabot configuration will now create up to two new PRs weekly:
- A grouping of all major dependency updates
- A grouping of all minor/patch dependency updates
I elected to split these groupings because minor/patch dependency updates are more likely to 

I have capped our open Dependabot PR limit to 4, which allows us to reenable auto-rebase.  We previously disabled auto-rebase to avoid too many concurrent integration test attempts against our 2 available namespaces (a subset of unlucky attempts would time out waiting for a namespace to become available).

We can certainly tweak settings further as needed (e.g. we may wish to ungroup major dependency updates) but I think this is a good starting point to hopefully make it easier for us to stay on top of dependency bumps.

DUOS team has been using this functionality for several months, starting from when it was in preview (it is now GA).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,10 +3,24 @@ updates:
 - package-ecosystem: npm
   directory: "/"
   schedule:
-    interval: daily
-  # Disable automatic rebasing to avoid tests kicking off all at once
-  rebase-strategy: "disabled"
-  open-pull-requests-limit: 10
+    interval: weekly
+  # See workflows/test-e2e.yml for the current list of k8s namespaces available for integration testing.
+  # Presently, there are only 2.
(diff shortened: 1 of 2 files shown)


fboulnois (approved):
Looks good 👍

fboulnois:
You may want to add a node Docker image update check too in the future.

snf2ye (approved):

PR #1598, opened 2024-01-16
[DR-3429] Expand Dependabot ignore list

okotsopoulos:
https://broadworkbench.atlassian.net/browse/DR-3429

These are all of the major dependency updates that I couldn't get [working](https://github.com/DataBiosphere/jade-data-repo-ui/pull/1599), using Dependabot's [last grouped major update PR](https://github.com/DataBiosphere/jade-data-repo-ui/pull/1597) as a reference.

Many are related to React 18 / Redux 8 upgrades or linting.

I filed tickets for the upgrades or referenced existing ones:
- https://broadworkbench.atlassian.net/browse/DR-2953
- https://broadworkbench.atlassian.net/browse/DR-2952
- https://broadworkbench.atlassian.net/browse/DR-3428
- https://broadworkbench.atlassian.net/browse/DR-2799
- https://broadworkbench.atlassian.net/browse/DR-3432
- https://broadworkbench.atlassian.net/browse/DR-3433

We no longer group major version updates: they will be PR-ed individually.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -23,9 +23,28 @@ updates:
         - "patch"
   ignore:
   # newer versions of prettier seem to format the code in a way that is incompatible with eslint (DR-2953)
+  # more generally, eslint and prettier upgrades are likely to need developer intervention
+  - dependency-name: "@typescript-eslint/eslint-plugin"
+  - dependency-name: "eslint-config-prettier"
+  - dependency-name: "eslint-plugin-prettier"
   - dependency-name: "prettier"
   # newer versions of material ui (MUI) cause our styling to not respect theming or straight up break the UI (DR-2952)
   - dependency-name: "@mui/*"
   # @mui/styles is incompatible with react 18, we should replace @mui/styles with tss-react (DR-3428)
-  - dependency-name: "@types/react"
+  # more generally, react 18 / redux 8 upgrades are a large lift (DR-2799)
+  - dependency-name: "@reduxjs/tookit"
+    update-types: [ "version-update:semver-major" ]
+  - dependency-name: "@testing-library/react"
+    update-types: [ "version-update:semver-major" ]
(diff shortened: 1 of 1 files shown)


snf2ye (approved):

.github/dependabot.yml
```diff
@@ -26,3 +26,6 @@ updates:
   - dependency-name: "prettier"
   # newer versions of material ui (MUI) cause our styling to not respect theming or straight up break the UI (DR-2952)
   - dependency-name: "@mui/*"
+  # @mui/styles is incompatible with react 18, we should replace @mui/styles with tss-react (DR-3428)
+  - dependency-name: "@types/react"
+    update-types: [ "version-update:semver-major" ]
```
snf2ye: Do you know if just pinning just the react types library going to be sufficient to keep react on v17? I see in the package.json that we also have a reference to react itself.

nmalfroy: I _think_ that the typescript version of the react dependency is basically the shim to pull in react?

okotsopoulos: Great point, @snf2ye -- we are going to have to pin quite a bit more here.  I've converted the PR back to draft.

rjohanek (approved):
this looks good to me just one comment - I agree I think we should shy away from grouping major updates. However, I fear that if we just disable all major updates for now then we may fall behind in a way that would make it harder when we do need to make updates. Is there a dependabot setting for if there is a major update, make a singular PR for just that update without grouping in others?

okotsopoulos:
> this looks good to me just one comment - I agree I think we should shy away from grouping major updates. However, I fear that if we just disable all major updates for now then we may fall behind in a way that would make it harder when we do need to make updates. Is there a dependabot setting for if there is a major update, make a singular PR for just that update without grouping in others?

@rjohanek , yes there is!  That's the default setting.  So if I remove the major grouping (which sounds like the right next move) without also ignoring major updates in general, new major updates outside of the existing ignore list will be PR-ed individually.

PR #1600, opened 2024-01-17
DR-3431: Switch to use sherlock for slack notifications

snf2ye:
https://broadworkbench.atlassian.net/browse/DR-3431
_____
# Custom Slack Integration: Lots of edge cases

### Issue in current implementation: Skipped steps cause entire workflow to show up as failed
You'll see in this PR that we remove the "reusable workflow" for our custom slack notification integration (originally added to drshub [here](https://github.com/DataBiosphere/terra-drs-hub/pull/100)). The code there works except if there is a step that is skipped - it has a bug that will report the entire workflow as failed if a step is skipped. This could be fixed, but with a much more verbose calling for every call of the slack notifications. 

Example:
image
image

For our use case, we don't need custom slack notifications, but if there is a case, then the removed code here can be used as a template to get started with. 

### Roadbump
- A reusable workflow doesn't know the status of the caller workflow. So, you have to determine this status in the calling workflow and github does not provide helpful helper methods for arguments passed to reusable workflows.
- Workarounds:
    1) Have a separate job for failing vs. successful slack notifications. This allows you to use the if: success() or if: failure() conditionals on the job, allowing for a concise job definition in the calling workflow. But, if you always want to report the status, then there is duplication
    2) Put the slack notification as a step instead of job - Steps would already know the status of the job. Con: Very verbose GHA and still can't inherently collect the status of other jobs in the workflow.
_____
# Solution: Switch to use Sherlock's implementation for slack notifications
More details here: https://docs.google.com/document/d/1G6-whnNJvON6Qq1b3VvRJFC7M9M-gu2dAVrQHDyp9Us/edit
I found that it just worked out of the box. I didn't need to do any further setup than just adding the notification step. 

### Where are we sending the notifications?
- E2E tests - send to #dsde-qa and #jade-alerts (Since we're sending these alerts to QA, we should make sure we're getting notified if they fail)
- Unit tests - Send to #jade-spam
- Image update - Send to #jade-spam 

### Testing
- forced successful run without the "develop" branch check
Slack notification went to #jade-spam
image

- forced failure run without the "develop" branch check
Slack notification went to #jade-alerts
image

- Test with "develop" branch check - no notification

### Custom Icons
The notifications with have custom icons based on the workflow. For our tests, we'll use the :terra-horse: icon, and for our dev-image-update & helm action, we'll use the :github_merged: icon. 
image
image
^ Note: this says it was the unit test workflow, but it will actually be used for our image update actions.


.github/workflows/test-e2e.yml
```diff
@@ -113,14 +113,11 @@ jobs:
...
+  report-workflow:
+    uses: broadinstitute/sherlock/.github/workflows/client-report-workflow.yaml@main
+    if: ${{ github.ref == 'refs/heads/develop' }}
     with:
-      workflow_name: Nightly E2E Tests
-      notify_on_success: true
-      failure_slack_channel_id: "CMYTGJVFY,C53JYBV9A" # #jade-alerts & #dsde-qa
-      success_slack_channel_id: "CK9M0ENRJ,C53JYBV9A" # #jade-spam & #dsde-qa
-      success: ${{ needs.e2e_test.result == 'success' }}
+      notify-slack-channels-upon-workflow-success: "#jade-spam"
```
okotsopoulos: Just curious, is there a reason that you are only notifying #jade-spam on success rather than completion?

.github/workflows/test-e2e.yml
```diff
@@ -113,14 +113,11 @@ jobs:
...
+      notify-slack-channels-upon-workflow-failure: "#jade-alerts"
```
okotsopoulos: Does this mean that a rerun of a failed GHA won't emit a notification to #jade-alerts if it succeeds?

okotsopoulos (approved):
Looks awesome… it's fantastic to be able to leverage Sherlock here, standardization for the win!  I especially like how you can reference the channel name rather than the ID.

I imagine you'll also be linking in #dsde-qa channel throughout?

.github/workflows/test-e2e.yml
snf2ye: I found this wording confusing as well - from the docs:

> “Upon completion” means “upon success or failure.” It’s like setting both the “upon success” and “upon failure” fields to the same channel.

.github/workflows/test-e2e.yml
snf2ye: That's correct and a good point. This was a part of the effort to reduce the number of notifications going to #jade-alerts, but I'm not set on this approach and am glad to revisit this conversation.

snf2ye (commented):
> I imagine you'll also be linking in #dsde-qa channel throughout?

Ah, thanks for the reminder! I meant to add this back.

.github/workflows/test-e2e.yml
okotsopoulos: Right!  That's what I figured it meant, thanks for confirming.

I'd think that we'd want to notify #jade-spam on both success and failure (so on completion), what do you think?

.github/workflows/test-e2e.yml
okotsopoulos: Sounds good!  I'm happy to keep it as-is.  (I'd be curious to know what other teams that use Sherlock for Slacking are doing in that regard.)

.github/workflows/test-e2e.yml
snf2ye: Ah, ok - That would be fine. I was just thinking that we send our nightly test results for the api to #jade-alerts, but sending them both to spam would be fine.

snf2ye: Probably better to just pick a channel rather than them going to different channels. I can just switch this to #jade-spam.

snf2ye: I've updated the channels for [almost all of the alerts](https://github.com/DataBiosphere/jade-data-repo-ui/pull/1600/commits/87db32250bc67215d50c38b758404feb907c6ce1):
- If we're also sending the alert to QA, then we probably should have the result in jade-alerts. So, e2e is sent to #jade-alerts
- Otherwise, send to jade-spam. I think that's what we were originally doing with our dev image updates. And, I think that's a fine place to report the unit tests.

.github/workflows/dev-image-update.yaml
```diff
@@ -98,12 +98,9 @@ jobs:
...
-    if: ${{ always() }}
+  report-workflow:
+    uses: broadinstitute/sherlock/.github/workflows/client-report-workflow.yaml@main
     with:
-      workflow_name: Dev Image Update
-      notify_on_success: true
-      success: ${{ needs.update_image.result == 'success' && needs.helm_tag_bump.result == 'success' && needs.cherry_pick_image_to_production_gcr.result == 'success' }}
+      notify-slack-channels-upon-workflow-completion: "#jade-spam"
+    permissions:
+      id-token: write
```
rjohanek, line 107: nit newline

rjohanek (approved):
this looks good to me - to recap, #jade-alerts and #dsde-qa are only notified for end to end tests, on both failure and success. All other updates go to #jade-spam. I see this conversation throughout but I'm not sure what the conclusion is - I thought I remembered from your previous PR that only failures (not all completions) were sent to #jade-alerts. Is that still desired?

snf2ye:
> this looks good to me - to recap, #jade-alerts and #dsde-qa are only notified for end to end tests, on both failure and success. All other updates go to #jade-spam. I see this conversation throughout but I'm not sure what the conclusion is - I thought I remembered from your previous PR that only failures (not all completions) were sent to #jade-alerts. Is that still desired?

Yes, sorry, there was a lot of back and forth. [This message](https://github.com/DataBiosphere/jade-data-repo-ui/pull/1600#discussion_r1457482848) summarizes where I landed and the PR description is up to date on the latest state. 

I'm changing course slightly from the last PR. My thought is that if a notification is important enough to send over to #dsde-qa, then we probably also should see it in #jade-alerts. Additionally, it would be confusing if we re-run a failure and it the successful notification ends up in a different slack channel. We're only sending notifications for e2e tests, so correct - we'll now see both success and failure in #jade-alerts. All other notifications are getting routing to #jade-spam. 

I'll post an additional message in slack to more widely publicize this.

pshapiro4broad (approved):
I'm approving but I'm not sure we're better off using Sherlock for this use case. Using slack-github-action gives us more control over the slack message content which Sherlock may never support. The overhead of having our own action vs using Sherlock doesn't seem too great to me. On the other hand, if the slack API changes a lot over time it could become a burden to keep our code working.

snf2ye:
> I'm approving but I'm not sure we're better off using Sherlock for this use case. Using slack-github-action gives us more control over the slack message content which Sherlock may never support. The overhead of having our own action vs using Sherlock doesn't seem too great to me. On the other hand, if the slack API changes a lot over time it could become a burden to keep our code working.

@pshapiro4broad  - At least in these cases, I don't think that we need to customize the slack message. The two most important pieces of information are provided: which workflow failed and a link to the workflow. 
But, we could definitely revisit this reusable workflow code for other use cases!

PR #1618, opened 2024-02-07
DC-833: Update slack channel for E2E tests

snf2ye:
https://broadworkbench.atlassian.net/browse/DC-833

Any notifications sent to the qa channel are considered release blockers if they fail. We do not consider E2E tests release blockers, so we’re going to remove notifications sent to QA. 

Since it’s no longer sent to QA, I will also move the destination channel to #jade-spam

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -117,7 +117,7 @@ jobs:
     uses: broadinstitute/sherlock/.github/workflows/client-report-workflow.yaml@main
     if: ${{ github.ref == 'refs/heads/develop' }}
     with:
-      notify-slack-channels-upon-workflow-completion: '#jade-alerts,#dsde-qa'
+      notify-slack-channels-upon-workflow-completion: '#jade-spam'
       notify-slack-custom-icon: ':terra-horse:'
     permissions:
       id-token: write


.github/workflows/test-e2e.yml
```diff
@@ -117,7 +117,7 @@ jobs:
     uses: broadinstitute/sherlock/.github/workflows/client-report-workflow.yaml@main
     if: ${{ github.ref == 'refs/heads/develop' }}
     with:
-      notify-slack-channels-upon-workflow-completion: '#jade-alerts,#dsde-qa'
+      notify-slack-channels-upon-workflow-completion: '#jade-spam'
```
okotsopoulos, line 120: I could see sending all alerts to #jade-spam, but would we also like to send alerts on workflow failure to #jade-alerts still, since a failure is something that we'd likely want to look at even if not officially a release blocker?

okotsopoulos (approved):
Preemptive approval, and my apologies for yet another question related to our many Slack channels.

.github/workflows/test-e2e.yml
rjohanek: I have the same question- why remove from jade-alerts?

rjohanek (approved):

.github/workflows/test-e2e.yml
snf2ye: I would prefer to have both of the UI test outputs in the same spot and there was pressure to move things out of jade-alerts. I personally check both channels. 

I definitely am not tied to this setup, but I think I will push ahead with this change.  If we find that the results are getting ignored or we want to change it for any reason, I'm glad to put in another PR!

PR #1631, opened 2024-03-14
DC-707: Add tables to match new snapshotbuilder settings

snf2ye:


snf2ye:
Oh, right! This change now needs to happen in jade-data-repo. I'll move these changes to that repo.

PR #1634, opened 2024-03-18
[DC-929] remove ui_integration folder and add pointer

s-rubenstein:


tools/ui_integration.md
```diff
@@ -0,0 +1 @@
+Looking for setup_ui_integration.py script? Try checking [here](https://github.com/DataBiosphere/jade-data-repo/tree/develop/tools/setupResourceScripts)
```
s-rubenstein: Do you think this is helpful? I thought it might be nice to add a pointer where people expected the scripts to be.

pshapiro4broad: This is fine although maybe instead reference the PR that merged the files in? The path to the code could change over time so unless you pin the version in the path, it could get stale. A PR link would always exist. Or you could link to the repo as whole and let the reader look for the script by name.

pshapiro4broad (approved):
Looks OK, although I wonder what we should do about `tools/ingest.js`. Maybe the whole `tools` directory should be retired now, unless `ingest.js` still has a use for us.

rjohanek (approved):

s-rubenstein:
I don't really know enough about `tools/injest.js` to know what to do there, so my inclination is to leave it and if someone feels strongly they can always move it later.

s-rubenstein:
I have verified that these tests are also failing on develop. Given that we are getting failures on the develop branch, and the PR itself doesn't have any code changes (just removing scripts and JSON files), I am going to merge this in.

PR #1642, opened 2024-04-12
DC-994 - Fix remove user from role on datasets/snapshots

snf2ye:
https://broadworkbench.atlassian.net/browse/DC-994
______
### Original bug
Removing a dataset or snapshot policy member in the TDR UI no longer works, and yields an “InvalidMemberEmail” toast notification.

https://github.com/DataBiosphere/jade-data-repo-ui/assets/13254229/b2b5fbdd-edef-48bc-adc2-1bb3b09052ce

### With fix

https://github.com/DataBiosphere/jade-data-repo-ui/assets/13254229/792905fc-dea3-4132-9594-672fe9f4751a

### Code Changes
- Pass in the user email to be removed
- Define function parameters in typescript
- Add E2E test for dataset add and remove policies

### DIFF
--- a/cypress/integration/datasetSharing.spec.js
+++ b/cypress/integration/datasetSharing.spec.js
@@ -1,3 +1,4 @@
+const newUser = 'voldemort.admin@test.firecloud.org';
 describe('test dataset sharing', () => {
   beforeEach(() => {
     cy.intercept('GET', 'api/repository/v1/datasets/**').as('getDataset');
@@ -22,6 +23,26 @@ describe('test dataset sharing', () => {
     cy.get('a > .MuiButtonBase-root').click();
 
     cy.wait(['@getDataset', '@getDatasetPolicies', '@getBillingProfileById']);
-    cy.get('[data-cy="manageAccessContainer"]').should('not.exist');
+    cy.get('[data-cy="roles-tab"]').should('not.exist');
+  });
+  it('A steward can add and remove a user to a dataset', () => {
+    cy.get('[placeholder="Search keyword or description"]').type('V2F_GWAS');
+    cy.contains(/V2F_GWAS_Summary_Stats|V2F_GWAS_Summary_Statistics/g).should('be.visible');
+    cy.contains(/V2F_GWAS_Summary_Stats|V2F_GWAS_Summary_Statistics/g).click({ force: true });
+
+    cy.wait(['@getDataset', '@getDatasetPolicies', '@getBillingProfileById']);
+    cy.get('[data-cy="roles-tab"]').click();
+    // Add user as a custodian on the dataset
+    cy.get('[data-cy="enterEmailBox"]').type(newUser);
+    cy.get('[data-cy="roleSelect"]').click();
+    cy.get('[data-cy="roleOption-custodian"]').click();
+    cy.get('[data-cy="inviteButton"]').click();
+    // Confirm the user was added
+    cy.get('[data-cy="user-list-Custodians"]').click();
+    cy.get(`[data-cy="chip-${newUser}"]`).should('be.visible');
+
+    // Now, let's remove this user
+    cy.get(`[data-cy="chip-${newUser}"]`).find('[data-testid="CancelIcon"]').click();
+    cy.get(`[data-cy="chip-${newUser}"]`).should('not.exist');
   });
 });
--- a/src/components/ManageUsersView.tsx
+++ b/src/components/ManageUsersView.tsx
@@ -14,7 +14,7 @@ const styles = (theme: CustomTheme) => ({
 });
 
 interface ManageUsersProps extends WithStyles<typeof styles> {
-  removeUser?: () => void;
+  removeUser?: (removableEmail: string) => void;
   users: Array<string>;
 }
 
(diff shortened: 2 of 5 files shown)


cypress/integration/datasetSharing.spec.js
```diff
@@ -22,6 +23,26 @@ describe('test dataset sharing', () => {
...
+    cy.get('[data-cy="enterEmailBox"]').type(newUser);
+    cy.get('[data-cy="roleSelect"]').click();
+    cy.get('[data-cy="roleOption-custodian"]').click();
+    cy.get('[data-cy="inviteButton"]').click();
+    // Confirm the user was added
+    cy.get('[data-cy="user-list-Custodians"]').click();
+    cy.get(`[data-cy="chip-${newUser}"]`).should('be.visible');
+
+    // Now, let's remove this user
+    cy.get(`[data-cy="chip-${newUser}"]`).find('[data-testid="CancelIcon"]').click();
```
rjohanek, line 45: nice test!

src/components/ManageUsersView.tsx
```diff
@@ -28,8 +28,9 @@ function ManageUsersView({ classes, removeUser, users }: ManageUsersProps) {
           color="primary"
           label={user}
           key={user}
-          onDelete={removeUser}
+          onDelete={() => (removeUser ? removeUser(user) : null)}
```
nawatts: If desired, this could be simplified a little using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).

```suggestion
          onDelete={() => removeUser?.(user)}
```

src/components/ManageUsersView.tsx
```diff
@@ -28,8 +28,9 @@ function ManageUsersView({ classes, removeUser, users }: ManageUsersProps) {
...
           variant="outlined"
+          data-cy={`chip-${user}`}
```
rjohanek, line 33: what does this line do?

rjohanek (approved):

cypress/integration/datasetSharing.spec.js
okotsopoulos, line 1: Just curious, I'm happy to see a new test for this fix but is there a reason that it couldn't be verified using a component test instead?

okotsopoulos (approved):
Thanks very much for the fix!

src/components/ManageUsersView.tsx
snf2ye: Oh, nice! That is much cleaner, thank you!

cypress/integration/datasetSharing.spec.js
snf2ye: I thought an E2E test was appropriate here because we weren't seeing the error until it hit the API. But, there probably is also a way to test with component test.

src/components/ManageUsersView.tsx
fboulnois: I didn't know you could do this on functions. Is this syntax typically used on functions (e.g. in the Terra UI codebase)?

fboulnois (approved):
👍

cypress/integration/datasetSharing.spec.js
okotsopoulos: That's fair!  One component test idea that came to mind was verifying that the API endpoint URL was constructed as expected, with the email appended to the end.  But again, I'm happy to see test coverage of any kind for this.

cypress/integration/datasetSharing.spec.js
```diff
@@ -22,6 +23,26 @@ describe('test dataset sharing', () => {
     cy.get('a > .MuiButtonBase-root').click();
 
     cy.wait(['@getDataset', '@getDatasetPolicies', '@getBillingProfileById']);
-    cy.get('[data-cy="manageAccessContainer"]').should('not.exist');
+    cy.get('[data-cy="roles-tab"]').should('not.exist');
+  });
+  it('A steward can add and remove a user to a dataset', () => {
+    cy.get('[placeholder="Search keyword or description"]').type('V2F_GWAS');
+    cy.contains(/V2F_GWAS_Summary_Stats|V2F_GWAS_Summary_Statistics/g).should('be.visible');
+    cy.contains(/V2F_GWAS_Summary_Stats|V2F_GWAS_Summary_Statistics/g).click({ force: true });
```
rushtong, line 31: We've had issues with `click({ force: true })` hiding errors in some cases - I'm not entirely sure we've come down on one side or the other of this question.

rushtong (approved):
👍🏽

src/components/ManageUsersView.tsx
nawatts: `git grep '?\.(' | wc -l` shows it's used 20 times in Terra UI.

cypress/integration/datasetSharing.spec.js
snf2ye: Yeah, that probably would work.
I think that this is an important workflow for our UI so a E2E test is justified.

cypress/integration/datasetSharing.spec.js
snf2ye: Ah, yeah, it does seem a little hacky. This was copied from another spot in our E2E tests, so I'm guessing we've run into issues with successfully clicking this particular spot so I thought I would keep it. But, good to call out.

PR #1643, opened 2024-04-12
DC-988: Remove http-proxy-middleware dependency

snf2ye:
https://broadworkbench.atlassian.net/browse/DC-988
________
Dependabot recently opened a [PR ](https://github.com/DataBiosphere/jade-data-repo-ui/pull/1641)for a major upgrade to the http-proxy-middleware dependency that according to the release notes had breaking changes. After looking through our code, it does not appear that we have any references to this dependency. 

http-proxy-middleware was [first added back in 2019](https://github.com/DataBiosphere/jade-data-repo-ui/commit/ecb0990c755b5339409b634d18e77b97b4c948e9) to a file that is no longer in the repo. I'm having a little trouble tracking down when it was removed, but I don't think it is used anymore.

Tests are passing and changes have been deployed to my dev environment (https://jade-sh.datarepo-dev.broadinstitute.org/) demonstrating some basic click testing still works.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "jade-data-repo-ui",
-  "version": "0.299.0",
+  "version": "0.301.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "jade-data-repo-ui",
-      "version": "0.299.0",
+      "version": "0.301.0",
       "license": "BSD-3-Clause",
       "dependencies": {
         "@babel/runtime": "^7.24.4",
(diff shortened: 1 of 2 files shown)


evansuslovich (approved):
Looks good!

pshapiro4broad (approved):
Great!

It does make you wonder how many other libraries we have that aren't being used. Is there a way to generate a report that would answer this question?

okotsopoulos (approved):
🎉 thank you for removing this!

A data point from another repo: `duos-ui` has this dependency too but uses it still, for local development: https://github.com/DataBiosphere/duos-ui/blob/ebb25c64971d373511d9a78ea4b2b8474a9849ba/src/setupProxy.js

rjohanek (approved):
Nice!

snf2ye:
@pshapiro4broad 

> It does make you wonder how many other libraries we have that aren't being used. Is there a way to generate a report that would answer this question?

I'm not sure - I tried using the depcheck library but it lists quite a few libraries that are not used including typescript. So, that doesn't seem to be accurate.

PR #1655, opened 2024-05-20
[DC-1076] Working snapshot access request table

s-rubenstein:
This adds the request table to the Jade Data Repo UI. The request table has seven columns:
* Name
* Request ID
* Source Snapshot Id
* Created By
* Date Requested
* Status
* (Action column)

Explicitly not including/doing anything with the approve button yet - we want to wait for the approve journey to be more fleshed out first.

Screenshot 2024-05-21 at 3 52 53 PM
Screenshot 2024-05-21 at 3 53 02 PM


src/components/SnapshotAccessRequestView.test.tsx
```diff
@@ -25,24 +25,18 @@ const initialState = {
 };
 
 describe('SnapshotAccessRequestView', () => {
-  it('Renders an unordered list of snapshot access request ids', () => {
+  it('Renders', () => {
```
s-rubenstein, line 28: For now there is very little testing we can do here - it is mostly a passthrough for state. It may grow, though, as we end up needing to do more searching and filtering serverside.

src/components/table/SnapshotAccessRequestTable.tsx
```diff
@@ -0,0 +1,115 @@
...
+      name: 'Actions',
+      render: (row: SnapshotAccessRequestResponse) => (
+        <div>
+          <Button
+            color="primary"
+            onClick={() => row.id && dispatch(rejectSnapshotAccessRequest(row.id))}
+          >
+            Reject
+          </Button>
+          <Button color="primary">Approve</Button>
```
s-rubenstein, line 94: This will eventually call `createSnapshot` with the request id.

src/components/table/SnapshotAccessRequestTable.tsx
```diff
@@ -0,0 +1,115 @@
...
+      containsSearchString(snapshotAccessRequest),
+  );
+
+  const filteredAndSortedRequests = _.sortBy(
+    filteredRequests,
+    (snapshotAccessRequest: SnapshotAccessRequestResponse) =>
+      Date.parse(
+        snapshotAccessRequest.createdDate ? snapshotAccessRequest.createdDate : _.now().toString(),
+      ) * -1,
+  );
```
snf2ye, line 44: Would it make sense to put each of these in a useeffect?

s-rubenstein: Because there is no async and very limited cost, I think we can safely have them outside of a `useEffect`. The advantage of having them as derived values that are inlined, as opposed to within a `useEffect`, is that we don't need to manage them as state - the scoping is handled for us.

s-rubenstein: The time when you really want to `useEffect` is when you want to store values in state, because you don't want a `setState` call within the render loop.

snf2ye (approved):
Can you add a description with a screenshot? Otherwise, LGTM!

rjohanek (approved):
Nice! Looks good and only appears when snapshot requests exists!

PR #1664, opened 2024-06-11
[DC-1094] Add approve snapshot flow

s-rubenstein:


tsconfig.json
```diff
@@ -1,9 +1,10 @@
 {
   "compilerOptions": {
-    "target": "es2016",
+    "target": "es2021",
```
s-rubenstein, line 3: Change is necessary to use `.replaceAll`. It does not seem to have broken anything, and still builds okay.

src/libs/utilsTs.ts
```diff
@@ -18,3 +18,25 @@ export const urlEncodeParams = (params: Record<string, string | number | boolean
...
+
+/**
+ * @param id {string} the id of the snapshot access request
+ * @param snapshotName {string} the user specified name of the snapshot access request
+ *
+ * @return {string} An underscore joined name and id with all dashes and spaces converted to
+ * underscores and all non-alphanumeric or underscore character stripped out.
+ * It also trims all leading underscores
+ */
+export const generateSnapshotNameFromAccessRequestInformation = (
```
rjohanek, line 30: instead of handling on the front end - should we handle it on the back end? because if we just handle it here it will still be a problem for API users. I made a ticket for this recently: https://broadworkbench.atlassian.net/browse/DC-1128

src/sagas/repository.ts
```diff
@@ -1023,6 +1028,40 @@ export function* rejectSnapshotAccessRequest({ payload }: any): any {
...
+    ],
+  };
+  try {
+    yield call(authPut, `/api/repository/v1/snapshotAccessRequests/${id}/approve`);
+    yield call(authPost, '/api/repository/v1/snapshots', snapshotRequest);
+    yield put({
+      type: ActionTypes.APPROVE_SNAPSHOT_ACCESS_REQUEST_SUCCESS,
+    });
+    yield put({
+      type: ActionTypes.GET_SNAPSHOT_ACCESS_REQUESTS,
```
rjohanek: why get snapshot access requests here? and like not create?

src/libs/utilsTs.ts
s-rubenstein: I think API users could use whatever name they want, and may prefer to get an exception then have their name they provide be changed out from under them.

On the frontend I think we can more freely futz with the name, because the users aren't entering their own input.

src/sagas/repository.ts
s-rubenstein: We want to refresh the page so it gets the new status

rjohanek: Oh is it the page refresh? Do we not want to do that for the failure case as well?

src/sagas/repository.ts
```diff
@@ -1023,6 +1028,40 @@ export function* rejectSnapshotAccessRequest({ payload }: any): any {
...
+        datasetName: 'unused',
+        mode: SnapshotRequestContentsModelModeEnum.ByRequestId,
+        requestIdSpec: {
+          snapshotRequestId: id,
+        },
+      },
+    ],
+  };
+  try {
+    yield call(authPut, `/api/repository/v1/snapshotAccessRequests/${id}/approve`);
```
rjohanek, line 1049: I see approve called here, and the payload for the create endpoint constructed above, but where is the create endpoint actually called?

s-rubenstein: The line below! https://github.com/DataBiosphere/jade-data-repo-ui/pull/1664/files#diff-b0cc26de535914428b254ecf0a5ed0e23a1c0e9753bbcb59a2af7167f6cd37ceR1050

src/libs/utilsTs.ts
rjohanek: Okay, but if it is stewards creating the requests on behalf of researchers, they may not care if the name is changed and may be frustrated by failures. And then naming will be consistent as well. Because researchers won't be able to use this endpoint right?

src/sagas/repository.ts
rjohanek: ohh lol thank you

src/sagas/repository.ts
s-rubenstein: Hmmm... Yeah, I think that makes sense. I'll add it.

src/libs/utilsTs.ts
s-rubenstein: They can if they are also a steward on the snapshot. Curious other people's thoughts.

rjohanek (approved):

snf2ye (approved):

PR #1670, opened 2024-06-27
[DC-1152] Remove client side name generation

s-rubenstein:
We are now generating the name server-side for byRequestId snapshots, so we can remove that here.


pshapiro4broad (approved):
LGTM. I realize we can't make name optional without changing the API, but it's something we should keep in mind for the future.

s-rubenstein:
Yeah, I am still in favor of a V2 snapshot api. That's something that we have a ticket in our backlog for.

snf2ye (approved):

PR #1671, opened 2024-06-28
DCJ-476: Bypass vault in actions

rushtong:
## Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DCJ-476

## Summary of changes
This PR passes a token to the actions to bypass vault
Requires the following PRs:
* ~https://github.com/broadinstitute/datarepo-actions/pull/85~ (merged)


okotsopoulos (approved):
Looks good -- thank you!  I see that E2E tests ran successfully with the new credentials which gives me confidence that the other workflows will also work, but I'm sure we'll keep an eye out following merge.

Very much out of scope, but seeing all of the places where you've had to change an action version in code made me think that setting the `datarepo-actions` version in a [repo environment variable](https://github.com/DataBiosphere/jade-data-repo-ui/settings/variables/actions) would be an improvement.  Probably not worth the effort if our goal is to remove `datarepo-actions` soon.

rushtong:
> Very much out of scope, but seeing all of the places where you've had to change an action version in code made me think that setting the `datarepo-actions` version in a [repo environment variable](https://github.com/DataBiosphere/jade-data-repo-ui/settings/variables/actions) would be an improvement. Probably not worth the effort if our goal is to remove `datarepo-actions` soon.

Yes, this could be a great follow-on task. I think the conversation around this would depend on what our deployment process should look like - my opinion is that it should not look like what it does now.

aarohinadkarni (approved):
👍

okotsopoulos:
> Yes, this could be a great follow-on task. I think the conversation around this would depend on what our deployment process should look like - **my opinion is that it should not look like what it does now.**

@rushtong I share your opinion!

fboulnois (approved):
I agree with the other comments 👍

PR #1681, opened 2024-07-29
DC-1048: Add aggregate reader role to UI

snf2ye:
### Summary
We now have a new role "aggregate data reader" on snapshots. We want to enable snapshot stewards to be able to give users this role in the TDR UI. Users can also add aggregate data readers on snapshot create. 

https://github.com/user-attachments/assets/601f0bf6-b107-4f07-b988-258d56769424

https://github.com/user-attachments/assets/32968ce0-945d-4529-baa3-e1973feb3b74

### Testing
- Added component test for this role
- Manual testing

### Super quick additional fix
We previously displayed the "x" button that would imply that you can remove a data access control (DAC) group. However, we do not support deleting DACs from a snapshot. So, I've remove it. 
**Before:**
Screenshot 2024-07-29 at 11 40 54 AM
**With Fix:**
Screenshot 2024-07-29 at 11 40 32 AM

### DIFF
--- a/src/components/ManageUsersView.tsx
+++ b/src/components/ManageUsersView.tsx
@@ -28,7 +28,7 @@ function ManageUsersView({ classes, removeUser, users }: ManageUsersProps) {
           color="primary"
           label={user}
           key={user}
-          onDelete={() => removeUser?.(user)}
+          onDelete={removeUser ? () => removeUser?.(user) : undefined}
           variant="outlined"
           data-cy={`chip-${user}`}
         />
--- a/src/components/snapshot/SnapshotAccess.tsx
+++ b/src/components/snapshot/SnapshotAccess.tsx
@@ -40,6 +40,8 @@ function SnapshotAccess({
   userRoles,
   createMode,
 }: SnapshotAccessProps) {
+  const transformRoleToCreatePolicy = (role: string): string => `${_.camelCase(role)}s`;
+
   const addUsers = (role: string, usersToAdd: string[]) => {
     if (createMode) {
       const existingEmails = _.get(snapshotRequest, ['policies', `${role}s`], []);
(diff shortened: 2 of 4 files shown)


src/components/snapshot/SnapshotAccess.tsx
```diff
@@ -47,7 +47,12 @@ function SnapshotAccess({
 
       // needs this manual conversion because the permissions are different for
       // editing existing snapshot policies vs creating a new snapshot
-      if (role === 'steward' || role === 'reader' || role === 'discoverer') {
+      if (
+        role === 'steward' ||
+        role === 'reader' ||
+        role === 'discoverer' ||
+        role === 'aggregate_data_reader'
+      ) {
```
rjohanek: do we want to use the enum of these roles here?

rjohanek (approved):
looks good, one question

src/components/ManageUsersView.tsx
```diff
@@ -28,7 +28,7 @@ function ManageUsersView({ classes, removeUser, users }: ManageUsersProps) {
           color="primary"
           label={user}
           key={user}
-          onDelete={() => removeUser?.(user)}
+          onDelete={removeUser ? () => removeUser?.(user) : undefined}
```
s-rubenstein, line 31: Why this change? I would expect `removeUser?.(user)` to handle the undefined check.

snf2ye: This is for the additional fix I mentioned in the PR description. The "Chip" component is a material UI component and according to its documentation, it determines whether or not to display the "x" option based on whether or not the "onDelete" prop is defined.  So, even though removeUser was undefined in the DAC case, it was still returning a function  (something like`() => undefined`) and we instead want it to return just`undefined`. 

Before:
Screenshot 2024-07-29 at 1 44 03 PM

With Change:
Screenshot 2024-07-29 at 1 42 43 PM

src/components/snapshot/SnapshotAccess.tsx
snf2ye: Good point - Looking at this reminded me that these components are used in another spot: on snapshot create. That workflow was broken with this change. So, I've now fixed that.

s-rubenstein (approved):

PR #1682, opened 2024-08-01
[DCJ-572] Hide Activity tab

okotsopoulos:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/DCJ-572

## Addresses

Removes the Activity tab on TDR's top navigation bar.

For non-Admin users, it's currently timing out and driving up load balancer latency stats, which pages our On Call engineers.

This tab is clicked on only sporadically: this PR hides it so that users don't stumble upon it while we debug the underlying slowness.

Screenshot 2024-08-01 at 4 47 16 PM

Users can if needed still navigate to the /activity URL to view all jobs, but only TDR Admins / debuggers are likely to actually do that.

Screenshot 2024-08-01 at 4 47 30 PM


pshapiro4broad (approved):

LizBaldo (approved):

fboulnois (approved):
👍

PR #1684, opened 2024-08-08
[DC-1139]Add a Snapshot Summary to the Request Tab in TDR UI

cinyecai:
Added a modal that pops up when you click on the Request Name in the Snapshot Access Request Table that displays the summary of that Snapshot

https://github.com/user-attachments/assets/d2a9712d-47fb-40b4-9317-4fe6b4319341


src/components/common/Modal.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+const initialState: ModalState = {
+  open: false,
+};
+
+export class Modal extends React.PureComponent<ModalProps, ModalState> {
+  constructor(props: ModalProps) {
+    super(props);
+    this.state = initialState;
+  }
```
s-rubenstein: Rather than using a constructor and typing state, we can do this using a functional component. 
```suggestion
export const Modal = (props: ModalProps) => {
  const [open, setOpen] = useState(false)
```

src/components/common/Modal.tsx
```diff
@@ -0,0 +1,166 @@
...
+    });
+  };
+
+  handleClose = () => {
+    this.setState({ open: false });
+  };
+
+  render() {
+    const { classes, modalText, modalContent, modalHeading, isLoading } = this.props;
+    const { open } = this.state;
```
s-rubenstein: If you use a functional component, render as a method goes away, and the entire function is the render method. So you can delete the top line, and then move the destructuring to the top of the component.

src/components/table/SnapshotAccessRequestTable.tsx
```diff
@@ -58,7 +59,13 @@ function SnapshotAccessRequestTable({
     {
       label: 'Request Name',
       name: 'snapshotName',
-      render: (row: SnapshotAccessRequestResponse) => row.snapshotName || '',
+      render: (row: SnapshotAccessRequestResponse) => (
+        <Modal
```
s-rubenstein: I am not sure `Modal` is the right name for this component. Maybe `ModalTrigger`? The reason I say that is that `Modal` very specifically refers to the dialog that opens on the screen. This also handles the rendering of the text, and open state, so its purview is larger.

s-rubenstein (approved):

src/components/table/SnapshotAccessRequestTable.tsx
```diff
@@ -58,7 +59,13 @@ function SnapshotAccessRequestTable({
     {
       label: 'Request Name',
       name: 'snapshotName',
-      render: (row: SnapshotAccessRequestResponse) => row.snapshotName || '',
+      render: (row: SnapshotAccessRequestResponse) => (
+        <ModalTrigger
```
pshapiro4broad: I agree that this isn't `Modal` but `ModalTrigger` also doesn't seem right to me. This is rendering a prompt (which looks like a hyperlink) that renders its content into a modal dialog. Maybe `ModalLink` would be better?

src/components/table/SnapshotAccessRequestTable.tsx
```diff
@@ -58,7 +59,13 @@ function SnapshotAccessRequestTable({
...
+          modalText={row.snapshotName}
```
pshapiro4broad, line 64: I don't think this is the "modal text". It's the prompt which is a link to the modal. So maybe `linkText` or `triggerPrompt`, if the name `trigger` is kept.

src/components/common/ModalTrigger.tsx
```diff
@@ -0,0 +1,146 @@
...
+      backgroundColor: theme.palette.common.white,
+      zIndex: 100,
+    },
+  });
+
+interface ModalTriggerProps extends WithStyles<typeof styles> {
+  modalText: string;
+  modalContent: string;
+  modalHeading: string;
+  isLoading: boolean;
```
pshapiro4broad: This code supports async loading but that appears to be unused here. If it's not used by the current code, I would recommend removing it for now.

src/components/common/ModalTrigger.tsx
```diff
@@ -0,0 +1,146 @@
...
+}
+
+export function ModalTrigger(props: ModalTriggerProps) {
```
pshapiro4broad: I like the idea of a general purpose modal but I wonder if that makes sense in the TDR UI. There are already a number of modals in TDR but each are done as a separate custom class. It would be more compelling if this PR could refactor an existing model to also use this code, to show that there's value in having a general class for modal displays.

pshapiro4broad (approved):
This looks OK although I would remove the async code, if it's unused, and would also like to know if there's a possibility to use this code in a more general way. I'm skeptical that a general approach is needed here, as so far TDR hasn't done this in a general way.

Also, regarding the UI itself, I wonder how a user would know that clicking on the name would bring up the summary. It isn't something I would expect and I don't see how a user would know that a request summary exists at all.

cinyecai:
> I'm skeptical that a general approach is needed here, as so far TDR hasn't done this in a general way.

@pshapiro4broad, after browsing through the other modal code, I see what you mean. Would you suggest making this component specific to this purpose then (also, what rewriting would that involve, if any? Or would it just be giving the file a more specific name and location)?

> Also, regarding the UI itself, I wonder how a user would know that clicking on the name would bring up the summary. It isn't something I would expect and I don't see how a user would know that a request summary exists at all.

I see this point as well. Maybe worth getting designer input on?

src/components/common/ModalTrigger.tsx
```diff
@@ -0,0 +1,130 @@
...
+    closeButton: {
+      position: 'absolute',
+      right: theme.spacing(1),
+      top: theme.spacing(1),
+      color: theme.palette.grey[500],
+    },
+    dialogInstructions: {
+      margin: 0,
+      padding: `${theme.spacing(2)} !important`,
+      whiteSpace: 'pre-wrap',
```
s-rubenstein:
```suggestion
      whiteSpace: 'pre-wrap',
      height: '10rem',
      overflowY: 'scroll',
```

PR #1692, opened 2024-08-29
[DC-1221] Add summary view for tdr ui

s-rubenstein:


src/components/common/TextWithModalDetails.tsx
```diff
@@ -73,54 +73,26 @@ const styles = (theme: CustomTheme) =>
...
-        maxWidth="md"
-        onClose={handleClose}
-        aria-labelledby="customized-dialog-title"
-        open={open}
-      >
+      <Dialog fullWidth maxWidth="md" onClose={onDismiss} open={true}>
         <DialogTitle className={classes.dialogTitle} id="customized-dialog-title">
           {modalHeading}
-          <IconButton aria-label="Close" className={classes.closeButton} onClick={handleClose}>
+          <IconButton aria-label="Close" className={classes.closeButton} onClick={onDismiss}>
```
rjohanek, line 89: why is the onDismiss method in the props now instead of defined in the component? And why the name change from handleClose?

src/components/table/SnapshotAccessRequestTable.tsx
```diff
@@ -55,16 +92,29 @@ function SnapshotAccessRequestTable({
...
-        />
+        <Button
+          className={classes.openButton}
+          aria-label={row.snapshotName}
+          onClick={() => openDetailsModal(row)}
+          disableFocusRipple={true}
+          disableRipple={true}
+        >
+          {row.snapshotName}
+        </Button>
```
rjohanek, line 117: why change to a button here? How was the modal working before?

rjohanek (commented):
looks good overall, just a couple questions for my own understanding, and in the description can you provide a picture or video of the functionality/modal?

src/components/common/TextWithModalDetails.tsx
s-rubenstein: In order to have the flow work for making the API call, I had to pull the button out. So now the open state is handled by the parent, rather than the component itself. (I could fold the API call into this component, but it would make it less generalizeable.) It may be worth renaming the component?

src/components/table/SnapshotAccessRequestTable.tsx
s-rubenstein: Previously the button was within the modal - that worked because there was no additional API call. There is now, so we want to manage the async behavior separately

s-rubenstein:
Uploading Screen Recording 2024-08-29 at 3.21.51 PM.mov…

s-rubenstein:
https://github.com/user-attachments/assets/45308304-c66c-497d-b34c-6d51637c41e6

rjohanek (approved):

cinyecai (approved):
LGTM - I had similar questions to Rae, so those answers were helpful and thank you for including the video!

PR #1706, opened 2024-09-30
[DC-1269] Remove limit on dependabot and integration references

s-rubenstein:
I believe these should all get removed with the move to point at dev, rather than integration for the tests


snf2ye (approved):
🎉 Thanks!

fboulnois (approved):
yes! 👍

PR #1725, opened 2024-12-17
DT-1108: Manually bump cherry pick action version 

snf2ye:
So that it picks up this change: https://github.com/DataBiosphere/jade-data-repo/pull/1874

We have seen that the action works in TDR. We'll see this in action on merge of this PR.


fboulnois (approved):
👍

pshapiro4broad (approved):

PR #1727, opened 2024-12-19
[DT-984]Create Full View Snapshot Button in TDR UI

rjohanek:
Ticket: https://broadworkbench.atlassian.net/browse/DT-984

## Summary
Adds a create full-view snapshot button in two locations, on the dataset overview page and on the side panel of the data view page for a dataset.

## Addresses
Helps lighten the burden of self-service ingest by removing the need to create an asset for a snapshot.

## Changes
1. Creates new component FullViewSnapshotButton
2. Adds FullViewSnapshotButton to DatasetOverviewPanel
3. Adds FullViewSnapshotButton to FilterPanel
4. Adds SnapshotPopup to DatasetOverview (only needs to be added in one location because it already exists in DataView because the side panel on the data page is already used to create snapshots)
5. Converts SnapshotPopup to tsx

## Tests
Adds component tests for new FullViewSnapshotButton component
Adds component tests for FilterPanel and DatasetOverviewPanel to test that the button appears

## Visual Aids
Here is a video showing the two buttons and what happens when you click one. The video showing the entire snapshot creation process is too long to post here, but I added it to our shared Data Team drive. Additionally, the mocks are linked here: https://www.figma.com/design/e8UwJPyz7d9mFviSHZ1nNl/Create-Full-Snapshot-view-in-TDR?node-id=2-122&t=FlhLoSSxxGnEy3og-1

https://github.com/user-attachments/assets/9e49d9ec-1f38-4e4f-9f4b-a9ea653ee3ae


snf2ye (approved):
Looks great! Awesome to see lots more test coverage and typescript!

One sanity check request -- Do the snapshots created by full view and by query look as expected? For by query, we could filter down to like 1 row and make sure the row count matches. And then make sure it looks like it has correctly created a full view snapshot.

src/components/common/FullViewSnapshotButton.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+      id: 'uuid',
+      name: 'Test Snapshot',
+    },
+  },
+  profiles: {
+    profiles: [{ id: 'profile1', name: 'Test Profile 1' }],
+  },
+};
+
+describe('FullViewSnapshotButton component with permission', () => {
```
s-rubenstein: I think you can nest describes, so you can have a `describe('FullViewSnapshotButton'` and then within it `describe('with permission'` and `describe('without permission'`. Alternatively, you could make the beforeEach lambda a method that takes  a dataset and an initial state and that way you can avoid repeating code and only have one describe.

In general as a rule we want to only have one top level `describe` per component.

src/components/snapshot/SnapshotPopup.tsx
```diff
@@ -0,0 +1,247 @@
...
+    },
+    jadeLink: {
+      ...theme.mixins.jadeLink,
+    },
+  });
+
+interface SnapshotPopupProps {
+  classes: ClassNameMap;
+  dispatch: Dispatch<Action>;
+  filterData: any;
```
s-rubenstein, line 88: Could we use a better type here than `any`?

src/components/snapshot/SnapshotPopup.tsx
```diff
@@ -0,0 +1,247 @@
...
+            SnapshotIncludeOptions.RELATIONSHIPS,
+            SnapshotIncludeOptions.ACCESS_INFORMATION,
+            SnapshotIncludeOptions.PROFILE,
+            SnapshotIncludeOptions.DATA_PROJECT,
+          ],
+        }),
+      );
+      dispatch(getSnapshotPolicy(snapshot.id));
+    }
+  }, [snapshot, dispatch]);
```
s-rubenstein: Do we want to subscribe on `dispatch` here? I know we ran into some issues with that elsewhere

src/sagas/repository.ts
```diff
@@ -228,34 +228,49 @@ export function* createSnapshot(): any {
...
     assetName,
     filterStatement,
     joinStatement,
     policies,
   } = snapshots.snapshotRequest;
 
   const datasetName = dataset.name;
-  const mode = 'byQuery';
-  const selectedAsset = _.find(dataset.schema.assets, (asset) => asset.name === assetName);
+  let snapshotRequest;
```
s-rubenstein: nit: Rather than a let with some assignments here, I'd love to try to use a switch expression on the `SnapshotRequestContentsModelModeEnum` - I think that will let us be a bit more declarative about things and maintain immutability

src/components/common/FullViewSnapshotButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+  billingProfiles,
+}: FullViewSnapshotButtonProps) {
+  useOnMount(() => {
+    dispatch(getBillingProfiles());
+  });
+
+  const defaultBillingProfile = dataset.defaultProfileId;
+  // if user does not have permission on billing profile, disable button and show tooltip
+  const isDisabled = !billingProfiles.map((model) => model.id).includes(defaultBillingProfile);
+  const tooltipText = 'You do not have access to the billing profile associated with this dataset.';
```
pshapiro4broad: The mocks show a step where the UI prompts for a billing profile, which (I believe) would allow a full view snapshot even if the user doesn't have access to the dataset billing profile. Was it decided to put off that work for now?

src/components/common/FullViewSnapshotButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const handleCreateFullViewSnapshot = () => {
+    const name = `Full_View_Snapshot_of_${dataset.name}_${now()}`;
```
pshapiro4broad, line 43: What is the use case for creating a snapshot with an automatically generated name? For many users, the snapshot name is used to keep track of a specific data release, and I believe that providing a custom name is an important part of that process. Have we validated the use case where we don't allow the users to name the snapshot?

src/components/common/FullViewSnapshotButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+  return (
+    <TerraTooltip title={isDisabled ? tooltipText : ''}>
+      <span>
+        <Button
+          variant="outlined"
+          disableElevation
+          onClick={() => handleCreateFullViewSnapshot()}
+          disabled={isDisabled}
+        >
+          Create Full-View Snapshot
```
pshapiro4broad: Is `full-view` the preferred way to write this phrase? I would expect it to be written without a hyphen, as `Create a full view snapshot`. This is the terminology is see here https://support.terra.bio/hc/en-us/articles/31203245858715-One-click-data-releases-through-full-view-snapshots-Releasing-01-06-2025 and here https://support.terra.bio/hc/en-us/articles/4407243163291-How-to-create-snapshots-in-TDR-GCP

src/components/common/FullViewSnapshotButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+      ),
+    );
+    dispatch(createSnapshot());
+  };
+
+  return (
+    <TerraTooltip title={isDisabled ? tooltipText : ''}>
+      <span>
+        <Button
+          variant="outlined"
```
pshapiro4broad, line 60: Why is this button an outline button? I would expect it to have the same style as the `Create Snapshot` button, since it's an equivalent operation in this side panel. That button uses `variant="contained"`.

src/components/common/FullViewSnapshotButton.tsx
snf2ye: I believe that users will be able to enter their own name with this next ticket: https://broadworkbench.atlassian.net/browse/DT-1001

src/components/common/FullViewSnapshotButton.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+    beforeEach(() => {
+      const dataset = { defaultProfileId: 'profile1' };
+      mountFullViewSnapshotButton(dataset);
+    });
+
+    it('Displays the button with correct text', () => {
+      cy.get('button').should('contain.text', 'Create Full-View Snapshot');
+    });
+
+    it('Button is clickable, calls createSnapshot, and does not have a tooltip', () => {
```
pshapiro4broad: Does this assert that it doesn't have a tooltip?

src/components/common/FullViewSnapshotButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+  dispatch: Dispatch<Action>;
+  dataset: DatasetModel;
+  billingProfiles: Array<BillingProfileModel>;
+}
+
+function FullViewSnapshotButton({
+  dataset,
+  dispatch,
+  billingProfiles,
+}: FullViewSnapshotButtonProps) {
```
pshapiro4broad: Sonar flags this as a case where the props should be marked as readonly 
```suggestion
}: Readonly<FullViewSnapshotButtonProps>) {
```

I see that this PR [adds new sonar issues](https://sonarcloud.io/project/issues?id=DataBiosphere_jade-data-repo-ui&pullRequest=1727&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true). I don't think there's a quality gate set up for this repo so we need to enforce it ourselves if we think that fixing sonar issues is important. Do we want to fix sonar issues like this?

src/components/common/FullViewSnapshotButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+  useOnMount(() => {
+    dispatch(getBillingProfiles());
+  });
+
+  const defaultBillingProfile = dataset.defaultProfileId;
+  // if user does not have permission on billing profile, disable button and show tooltip
+  const isDisabled = !billingProfiles.map((model) => model.id).includes(defaultBillingProfile);
```
pshapiro4broad: You could use `some()` here. Also note that `defaultBillingProfile` could be `undefined`
```
  const isDisabled = !billingProfiles.some(model => model.id == defaultBillingProfile);
```

src/components/common/FullViewSnapshotButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+    const name = `Full_View_Snapshot_of_${dataset.name}_${now()}`;
+    const description = `Full-View Snapshot of Dataset with Dataset name ${dataset.name}, and Dataset id ${dataset.id}.`;
+    dispatch(
+      snapshotCreateDetails(
+        name,
+        description,
+        SnapshotRequestContentsModelModeEnum.ByFullView,
+        null,
+        null,
+        dataset,
```
pshapiro4broad: I believe if you move the `null` parameters to the end, you can omit them in this case.
```suggestion
        dataset,
```

```
export const { snapshotCreateDetails } = createActions({
  [ActionTypes.SNAPSHOT_CREATE_DETAILS]: (
    name,
    description,
    mode,
    dataset,
    assetName,
    filterData,
  ) => ({
```

src/components/dataset/data/sidebar/panels/CreateSnapshotPanel.jsx
```diff
@@ -66,7 +67,16 @@ export class CreateSnapshotPanel extends React.PureComponent {
   saveNameAndDescription = () => {
     const { dispatch, switchPanels, filterData, dataset } = this.props;
     const { name, description, assetName } = this.state;
-    dispatch(snapshotCreateDetails(name, description, assetName, filterData, dataset));
+    dispatch(
+      snapshotCreateDetails(
+        name,
+        description,
+        SnapshotRequestContentsModelModeEnum.ByQuery,
```
pshapiro4broad, line 74: There are now two modes used in this API, which have different required parameters. Would it be better to have two different methods instead of requiring `mode` to be passed in?

src/reducers/snapshot.ts
```diff
@@ -291,16 +294,21 @@ export default {
...
+        const { name, description, mode, assetName, filterData, dataset } = action.payload;
+        let snapshotRequest = {
           ...state.snapshotRequest,
           name,
           description,
-          assetName,
-          joinStatement,
+          mode,
         };
+        if (mode === SnapshotRequestContentsModelModeEnum.ByQuery) {
```
pshapiro4broad, line 312: Would it be a lot more work to split this into two action types? I'm not sure what's typically done here but maybe it would be better to avoid conditional code based on `mode`.

src/components/common/FullViewSnapshotButton.tsx
rjohanek: yes that work is also going to be done in the follow up ticket

src/components/common/FullViewSnapshotButton.tsx
rjohanek: I don't have a strong opinion on this. I followed the mocks which show an outlined button. The button is added in two places, on the data view side panel and the snapshots tab of the dataset. I'm not sure what would look most consistent in both locations. I'll follow up with Lou about this.

src/components/common/FullViewSnapshotButton.tsx
rjohanek: I see inconsistent use between full-view and full view even within the article. I'll follow up with Lou about this as well to see if he has any thoughts.

src/components/snapshot/SnapshotPopup.tsx
```diff
@@ -0,0 +1,247 @@
...
+          <div className={clsx(classes.centered, classes.content)}>
+            <CircularProgress />
+          </div>
+        </DialogContent>
+      </Dialog>
+    );
+  }
+
+  // this number represents the total rows in the snapshot, i.e. the sum over all tables
+  const rows = snapshot.tables?.map((t) => t.rowCount).reduce((a, b) => a + b);
```
pshapiro4broad: I think this is not quite correct as `rowCount` is `number?`. Something like this should work
```suggestion
  const rows = snapshot.tables?.map((t) => t?.rowCount ?? 0).reduce((a, b) => a + b);
```

Also I think `reduce()` might return `number?` as well without an initial value, so this would let `rows` be `number`:
```
  const rows = snapshot.tables?.map((t) => t?.rowCount ?? 0).reduce((a, b) => a + b) ?? 0;
```

src/components/snapshot/SnapshotPopup.tsx
```diff
@@ -0,0 +1,247 @@
...
+        <div>{properties}</div>
+      </div>
+    );
+  });
+
+  const readers = policies.find((p) => p.name === 'reader')?.members;
+
+  return (
+    <Dialog open={isOpen} onClose={handleClose}>
+      <div className={classes.exitButton} onClick={handleClose}>
```
pshapiro4broad: Does this handle keyboard exit as well as mouse exit? Since it's not a Button or Link I think it needs a `onKeyPress` to handle `ESC` correctly.

src/components/snapshot/SnapshotPopup.tsx
rjohanek: Everywhere that filterData is typed, it uses type any. Since it isn't documented anywhere else in the code, I think it might be out of scope to figure out the correct typing of this object.

src/components/snapshot/SnapshotPopup.tsx
rjohanek: When I remove it, I get a missing dependency eslint error. I could add a comment to suppress the error, but that seems maybe unnecessary when it works correctly with dispatch in the dependency array. Is there a reason we would want to do this given there is not an issue with it here? Like is it a principle that we don't want to subscribe to dispatch?

src/sagas/repository.ts
pshapiro4broad: I agree, also it looks like the two objects are the same except for the object inside `contents`,  so you could have the two cases in the switch generate only that part of the request object.

src/sagas/repository.ts
```diff
@@ -228,34 +228,49 @@ export function* createSnapshot(): any {
...
+  if (mode === SnapshotRequestContentsModelModeEnum.ByFullView) {
+    snapshotRequest = {
+      name,
+      profileId: dataset.defaultProfileId,
+      description,
+      contents: [
```
pshapiro4broad: Is this omitting `policies` on purpose?

pshapiro4broad (approved):
Looks OK, a few comments

src/sagas/repository.ts
rjohanek: currently, via the UI, the option to add policies upon creation is not available for the full-view snapshot, so I could include an empty array, but it isn't required/passing in nothing has the same effect

src/components/common/FullViewSnapshotButton.test.tsx
rjohanek: no, good catch. I was trying to test that and then decided it wasn't necessary so I just need to update the description.

src/components/common/FullViewSnapshotButton.tsx
rjohanek: yes I think it's a good idea, I'll take a look through the sonar issues.

src/components/common/FullViewSnapshotButton.tsx
rjohanek: good point, I'll add a case (and a test) for when there is no `defaultBillingProfile`

src/components/common/FullViewSnapshotButton.tsx
rjohanek: nice!

src/components/dataset/data/sidebar/panels/CreateSnapshotPanel.jsx
rjohanek: I'm kind of in favor of keeping the number of actions as low as we can, and reusing them where possible. My original approach was creating a new action that did the work of snapshotCreateDetails and createSnapshot in one action for Full-View snapshots specifically. However, the new action added a lot of new code in multiple places, was making the same API calls, and storing a lot of the same information. So, then I decided to simply reuse the existing actions because they already accomplish what is needed.

src/reducers/snapshot.ts
rjohanek: Rather than having an action for each mode (if we add support for other modes in the future), I think I prefer having one action that encompasses the action of snapshot create details for any mode. To me, this seems like the simpler approach. Is there a reason we wouldn't want conditionals based on mode? In the future, if we have more than one case, we could change to using a switch statement and it may look even clearer.

src/components/snapshot/SnapshotPopup.tsx
rjohanek: ah good catch, thank you!

src/components/snapshot/SnapshotPopup.tsx
rjohanek: yes it does handle keyboard esc! It does something weird to the button that goes away when you mouseover, but I checked and this happens on dev as well, so it isn't a performance degradation. Here is a video from dev of what I am talking about: 

https://github.com/user-attachments/assets/2ec20a19-4290-4c19-b887-70265544f2e1

pshapiro4broad: Interesting, maybe this is something that `?react` is doing for us

src/reducers/snapshot.ts
pshapiro4broad: It looks like having one action handle two different modes means that the code needs to check for different modes in a few places, which may make it hard to keep the code updated if one of the modes changes in the future (such as a new parameter is required). But if there's a lot of overhead to add a new action, it's probably not worth doing it that way.

src/components/snapshot/SnapshotPopup.tsx
s-rubenstein: I believe that adding the comment to suppress the error is the standard practice because we've run into issues with it making a large number of api calls if it triggers whenever dispatch updates. Because we don't have polling it might be okay?

src/components/snapshot/SnapshotPopup.tsx
s-rubenstein: I'm okay with that for now, would you make sure there's a ticket to improve the typing though?

src/components/common/FullViewSnapshotButton.tsx
rjohanek: I talked to Lou, and he said: "As far as button styling, the blue colored button is only used when we want that to be the primary action. The outlined button is used when it’s not the primary or absolute thing we want users to do. So it’s not essential that a user may want or need to create a full-view snapshot so a Blue Primary button is not needed."

src/components/common/FullViewSnapshotButton.tsx
rjohanek: I checked with Lou and he said User Ed decided the hyphen is not necessary. So I've removed it.

src/components/dataset/data/sidebar/SidebarDrawer.jsx
```diff
@@ -8,13 +8,13 @@ const styles = (theme) => ({
...
     position: 'absolute',
-    top: '112px',
+    top: '0px',
     right: '0px',
     bottom: '0px',
     zIndex: 10,
   },
   drawer: {
-    top: '112px',
+    top: '0px',
```
rjohanek, line 17: fix to make the panel take up the whole page, as reflected in the mocks

src/reducers/snapshot.ts
```diff
@@ -277,10 +280,18 @@ export default {
...
+          immutable(state, {
+            snapshot: { $set: {} },
+            dialogIsOpen: { $set: false },
+          });
+        } else {
+          immutable(state, {
+            dialogIsOpen: { $set: action.payload },
+          });
+        }
+      },
```
rjohanek, line 294: I made this change to clear the snapshot out of state when the dialog is closed because without it the name and description text fields were pre-populated with the name and description of the last snapshot created. I tested and this fixes that and I don't see any other side effects.

s-rubenstein (approved):

src/components/snapshot/SnapshotPopup.tsx
```diff
@@ -121,7 +121,8 @@ function SnapshotPopup({
       );
       dispatch(getSnapshotPolicy(snapshotId));
     }
-  }, [snapshot.id, dispatch]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [snapshot.id]);
```
snf2ye, line 125: Why is this getting removed?

src/reducers/snapshot.ts
snf2ye: I think we always want to clear the snapshot state on opening the dialog. With the useEffect logic we just changed, we are expecting that the snapshot state is empty to start (like it is for full view snapshot button case - https://github.com/DataBiosphere/jade-data-repo-ui/blob/d79cbeb1f886d235abcf6eda0c7751c682a32a8b/src/reducers/snapshot.ts#L147) 

We also could make this only for closing the snapshot dialog. We're not using the open dialog option anywhere, I don't think.

src/components/snapshot/SnapshotPopup.tsx
rjohanek: I removed dispatch as a dependency because Sky said it often causes problems and it standard to not include it and just disable the exhaustive dependency requirement. This works the same with or without it.  This is the conversation: https://github.com/DataBiosphere/jade-data-repo-ui/pull/1727/files/f1ca98e2ddb99a146e614e9bf1d815e23d322247#r1904232191

src/reducers/snapshot.ts
rjohanek: I think we always want to clear the snapshot state on opening the dialog.
> I agree, I think that code you just sent already does that. 
So, because of that code you just sent (line 147) in snapshot.ts, I figured we didn't need to clear the snapshot for the open dialog case because it is cleared there every time the snapshot create job is launched and the dialog is opened. So, to clear it again once the dialog is opened would be at best redundant, and at worst, could clear something that we don't want to clear (since the job has already been submitted if it for some reason reason returned faster than the modal opened we would clear information we wanted to keep. I think this is very unlikely/maybe a case we don't need to consider). 

We also could make this only for closing the snapshot dialog. We're not using the open dialog option anywhere, I don't think.
> I think the open dialog option is only used when the snapshot create job is submitted. I think you're right that if we clear the snapshot on close, it should be empty on open already, so we don't necessarily need both. But, since we want to guarantee that it is empty upon opening the dialog as well, I feel comfortable having it in both locations.

src/components/snapshot/SnapshotPopup.tsx
snf2ye: Ok, thanks!

src/reducers/snapshot.ts
snf2ye: I'm worried about the case where someone adds code that utilizes the open snapshot action without the snapshot create job (and therefore the clear snapshot state call). But, you're right, "at worst, could clear something that we don't want to clear". I'm trying to save our future selves, but we don't know how they would use this in the future. 
I'm good with leaving it as is.

snf2ye (approved):

PR #1730, opened 2025-01-03
DT-1110: Migrate away from @mui/styles Part 1

snf2ye:
## Background
We must discontinue use of @mui/styles in order to migrate from React 17 to 18. ([source](https://mui.com/system/styles/api/))

>  @mui/styles was deprecated with the release of MUI Core v5 in late 2021. It is not compatible with [React.StrictMode](https://react.dev/reference/react/StrictMode) or React 18+, and it will not be updated. Please use [@mui/system](https://mui.com/system/getting-started/) instead.

There is guide for this migration here: https://mui.com/material-ui/migration/migrating-from-jss

Options considered:
- [`sx` prop](https://mui.com/system/getting-started/the-sx-prop/)  + styled components from `@mui/matertial/styles`  -  From migration guide - "We recommend sx API over styled for creating responsive styles or overriding minor CSS". We can use styled components for components with a lot of styles (4+) and for styling that is reused. 
- tss-react - Easiest change, but still would require using copilot per component. It would be more of a lateral step.

This PR uses the first option and uses the `sx` prop alongside styled components. Given that we have to touch every file manually and I think we can utilize copilot for the bulk of the work, I would prefer to do the harder change rather than the lateral change. 

## Steps for Upgrades
**Pre-checks:**
- Review the styles listed at the beginning of the file. Are they all referenced in the component? (Apparently if there are styles that don't match up to a component, it will try to infer where the component with those styles should go and add it) 

1) Run copilot command, specifically in VSCode with Claude 3.5 Sonnet.

>Can you migrate this component away from using JSS and references to '@mui/styles'? We want to instead use a combination of inline styling with the "sx" prop on Material UI components and styled components from @mui/material/styles.

>Let me explain something about how our code is set up: In the existing code, all of the styles currently live in a "styles" object. It contains a list of classes each with their own styles.

>Consider the following example:
```
const styles = (theme: CustomTheme) =>
  createStyles({
    pageRoot: {
      padding: '16px 24px',
      display: 'flex',
      flexDirection: 'column',
      height: '100%',
    },
    jadeTableSpacer: {
      paddingBottom: theme.spacing(12),
    }
});
```

>pageRoot and jadeTableSpacer are classes that we reference in the component. pageRoot has 4 styles and jadeTableSpacer has 1 style. We don't want to use inline styling with the sx for classes that have more than 3 styles, so we want to handle these two classes in different manners. Since pageRoot has 4 styles, we want to use a styled component. Since jadeTableSpacer only has 1 style, we want to use the inline styling with an "sx" prop. We can evaluate other style classes with this same rule: if there are 3 or fewer styles, use the "sx" prop. If there are more than 3 styles, then convert to a styled component.

>When using the "sx" prop, then we must use the material UI version of the component (e.g. Box instead of div and add an import statement at the beginning of the file). We must also switch any css references of direct numbers to include px (e.g. '10px' instead of 10)

2) Apply prettier
3) Review UI and code generated
4) Compare UI/styling between local UI and dev

### Common Issues
- It doesn't handle theme changes well - https://github.com/DataBiosphere/jade-data-repo-ui/commit/30ca7c2d2a6c15a6f8f8c6690781714a04c17cf2

## What step is the copilot upgrade taking? (example output from copilot)
I'll help migrate the SnapshotAccessRequestTable from JSS to use styled components and sx props. 

Let's analyze the styles:
1) More than 3 styles:
-  overlaySpinner (6 styles)
-  openButton (7 styles)
2) Convert to sx props:
- jadeLink (1 style with theme mixin)

Key changes made:
- Removed withStyles HOC and styles object
- Added styled import from @mui/material/styles
- Created styled components for complex styles: OverlaySpinner for loading overlay, RequestButton for the request name button, Converted jadeLink to use sx prop with theme mixin
- Removed classes prop and ClassNameMap interface
- Added Box component for styled span
- Simplified component exports
- Maintained all functionality while modernizing the styling approach

The component maintains the same styling and behavior while using modern MUI patterns.

**Example**
**Before with @mui/styles**
```
const styles = (theme: CustomTheme) => ({
  wrapper: {
    display: 'flex',
    justifyContent: 'center',
    marginTop: '1em',
  }
);
....
 <div className={classes.wrapper}>
...
</div>
```
**With sx prop**
```
 <Box sx={{ display: 'flex', justifyContent: 'center', marginTop: '1em' }}>
...
</Box>
```

##TODO - add example for styled component

## Components Impacted in this PR
**Updated screenshots as of 01-11-2024**
1) HomeView and 2) DatasetView
image

3) JobView
image

4) SnapshotView
image

5) DatasetOverview and 6) DatasetOverviewPanel
image
image

7) DatasetAccess and 8) UserList
image

9) SnapshotAccessRequestTable
image

### DIFF
--- a/src/components/DatasetView.tsx
+++ b/src/components/DatasetView.tsx
@@ -2,13 +2,15 @@ import React, { Dispatch } from 'react';
 import { connect } from 'react-redux';
 import { Action } from 'redux';
 import { Box } from '@mui/material';
+import { styled } from '@mui/system';
 import { getDatasets, addDatasetPolicyMember } from 'actions/index';
 import { DatasetSummaryModel } from 'generated/tdr';
 import { TdrState } from 'reducers';
 import { OrderDirectionOptions } from 'reducers/query';
-import theme from 'modules/theme';
 import DatasetTable from './table/DatasetTable';
 import { DatasetRoles } from '../constants';
+
+const Container = styled('div')(({ theme }) => theme.mixins.containerWidth);
 
 interface IProps {
   datasets: Array<DatasetSummaryModel>;
@@ -55,7 +57,7 @@ function DatasetView({
         mt: '1em',
       }}
     >
-      <Box sx={{ ...theme.mixins.containerWidth }}>
+      <Container>
         <Box>
           {datasets && (
             <DatasetTable
@@ -71,7 +73,7 @@ function DatasetView({
             />
           )}
         </Box>
-      </Box>
+      </Container>
     </Box>
   );
 }
--- a/src/components/DatasetView.tsx
+++ b/src/components/DatasetView.tsx
@@ -1,11 +1,12 @@
 import React, { Dispatch } from 'react';
 import { connect } from 'react-redux';
 import { Action } from 'redux';
-import { Box, Typography } from '@mui/material';
+import { Box } from '@mui/material';
 import { getDatasets, addDatasetPolicyMember } from 'actions/index';
 import { DatasetSummaryModel } from 'generated/tdr';
 import { TdrState } from 'reducers';
 import { OrderDirectionOptions } from 'reducers/query';
+import theme from 'modules/theme';
 import DatasetTable from './table/DatasetTable';
 
 import { DatasetRoles } from '../constants';
@@ -49,7 +50,7 @@ function DatasetView({
 
   return (
     <Box sx={{ display: 'flex', justifyContent: 'center', marginTop: '1em' }}>
-      <Box sx={{ width: '100%', maxWidth: '1200px' }}>
+      <Box sx={{ ...theme.mixins.containerWidth }}>
         {datasets && (
           <DatasetTable
             datasets={datasets}
--- a/src/components/DatasetView.tsx
+++ b/src/components/DatasetView.tsx
@@ -8,7 +8,6 @@ import { TdrState } from 'reducers';
 import { OrderDirectionOptions } from 'reducers/query';
 import DatasetTable from './table/DatasetTable';
 import { DatasetRoles } from '../constants';
-import Container from './common/Container';
 
 interface IProps {
   datasets: Array<DatasetSummaryModel>;
@@ -55,7 +54,7 @@ function DatasetView({
         mt: '1em',
       }}
     >
-      <Container>
+      <Box sx={{ width: '100%' }}>
         <Box>
           {datasets && (
             <DatasetTable
@@ -71,7 +70,7 @@ function DatasetView({
             />
           )}
         </Box>
-      </Container>
+      </Box>
     </Box>
   );
 }
--- a/src/components/DatasetView.tsx
+++ b/src/components/DatasetView.tsx
@@ -2,7 +2,7 @@ import React, { Dispatch } from 'react';
 import { connect } from 'react-redux';
 import { Action } from 'redux';
 import { Box } from '@mui/material';
-import { styled } from '@mui/system';
+import { CustomTheme, styled } from '@mui/material/styles';
 import { getDatasets, addDatasetPolicyMember } from 'actions/index';
 import { DatasetSummaryModel } from 'generated/tdr';
 import { TdrState } from 'reducers';
@@ -10,7 +10,7 @@ import { OrderDirectionOptions } from 'reducers/query';
 import DatasetTable from './table/DatasetTable';
 import { DatasetRoles } from '../constants';
 
-const Container = styled('div')(({ theme }) => theme.mixins.containerWidth);
+const Container = styled('div')(({ theme }: { theme: CustomTheme }) => theme.mixins.containerWidth);
 
 interface IProps {
   datasets: Array<DatasetSummaryModel>;
--- a/src/components/DatasetView.tsx
+++ b/src/components/DatasetView.tsx
@@ -2,15 +2,13 @@ import React, { Dispatch } from 'react';
 import { connect } from 'react-redux';
 import { Action } from 'redux';
 import { Box } from '@mui/material';
-import { CustomTheme, styled } from '@mui/material/styles';
 import { getDatasets, addDatasetPolicyMember } from 'actions/index';
 import { DatasetSummaryModel } from 'generated/tdr';
 import { TdrState } from 'reducers';
 import { OrderDirectionOptions } from 'reducers/query';
 import DatasetTable from './table/DatasetTable';
 import { DatasetRoles } from '../constants';
-
-const Container = styled('div')(({ theme }: { theme: CustomTheme }) => theme.mixins.containerWidth);
+import Container from './common/Container';
 
 interface IProps {
   datasets: Array<DatasetSummaryModel>;
--- a/src/components/DatasetView.tsx
+++ b/src/components/DatasetView.tsx
@@ -50,16 +50,6 @@ function DatasetView({
   return (
     <Box sx={{ display: 'flex', justifyContent: 'center', marginTop: '1em' }}>
       <Box sx={{ width: '100%', maxWidth: '1200px' }}>
-        <Typography
-          sx={{
-            color: 'primary.main',
-            fontSize: '54px',
-            lineHeight: '66px',
-            paddingBottom: '64px',
-          }}
-        >
-          Datasets
-        </Typography>
         {datasets && (
           <DatasetTable
             datasets={datasets}
--- a/src/components/DatasetView.tsx
+++ b/src/components/DatasetView.tsx
@@ -8,7 +8,6 @@ import { TdrState } from 'reducers';
 import { OrderDirectionOptions } from 'reducers/query';
 import theme from 'modules/theme';
 import DatasetTable from './table/DatasetTable';
-
 import { DatasetRoles } from '../constants';
 
 interface IProps {
@@ -49,21 +48,29 @@ function DatasetView({
   };
 
   return (
-    <Box sx={{ display: 'flex', justifyContent: 'center', marginTop: '1em' }}>
+    <Box
+      sx={{
+        display: 'flex',
+        justifyContent: 'center',
+        mt: '1em',
+      }}
+    >
       <Box sx={{ ...theme.mixins.containerWidth }}>
-        {datasets && (
-          <DatasetTable
-            datasets={datasets}
-            datasetRoleMaps={datasetRoleMaps}
-            datasetsCount={datasetsCount}
-            handleFilterDatasets={handleFilterDatasets}
-            handleMakeSteward={handleMakeSteward}
-            filteredDatasetsCount={filteredDatasetsCount}
-            searchString={searchString}
-            loading={loading}
-            refreshCnt={refreshCnt}
-          />
-        )}
+        <Box>
+          {datasets && (
+            <DatasetTable
+              datasets={datasets}
+              datasetRoleMaps={datasetRoleMaps}
+              datasetsCount={datasetsCount}
+              handleFilterDatasets={handleFilterDatasets}
+              handleMakeSteward={handleMakeSteward}
+              filteredDatasetsCount={filteredDatasetsCount}
+              searchString={searchString}
+              loading={loading}
+              refreshCnt={refreshCnt}
+            />
+          )}
+        </Box>
       </Box>
     </Box>
   );
--- a/src/components/DatasetView.tsx
+++ b/src/components/DatasetView.tsx
@@ -51,7 +51,7 @@ function DatasetView({
       sx={{
         display: 'flex',
         justifyContent: 'center',
-        mt: '1em',
+        marginTop: '1em',
       }}
     >
       <Box sx={{ width: '100%' }}>
--- a/src/components/DatasetView.tsx
+++ b/src/components/DatasetView.tsx
@@ -1,34 +1,16 @@
 import React, { Dispatch } from 'react';
 import { connect } from 'react-redux';
 import { Action } from 'redux';
-import { WithStyles, withStyles } from '@mui/styles';
+import { Box, Typography } from '@mui/material';
 import { getDatasets, addDatasetPolicyMember } from 'actions/index';
 import { DatasetSummaryModel } from 'generated/tdr';
-import { CustomTheme } from '@mui/material';
 import { TdrState } from 'reducers';
 import { OrderDirectionOptions } from 'reducers/query';
 import DatasetTable from './table/DatasetTable';
 
 import { DatasetRoles } from '../constants';
 
-const styles = (theme: CustomTheme) => ({
-  wrapper: {
-    display: 'flex',
-    justifyContent: 'center',
-    marginTop: '1em',
-  },
-  width: {
-    ...theme.mixins.containerWidth,
-  },
-  title: {
-    color: theme.palette.primary.main,
-    fontSize: '54px',
-    lineHeight: '66px',
-    paddingBottom: theme.spacing(8),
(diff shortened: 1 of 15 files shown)


src/components/JobView.tsx
```diff
@@ -1,86 +1,64 @@
...
-    lineHeight: '66px',
-    paddingBottom: theme.spacing(8),
-  },
-});
-
-interface IProps extends WithStyles<typeof styles> {
+interface IProps {
   jobs: Array<JobModel>;
-  dispatch: Dispatch<Action>;
+  dispatch: Dispatch<any>;
```
snf2ye: TODO: Why is this now any instead of Action?

src/components/dataset/overview/DatasetOverviewPanel.jsx
```diff
@@ -294,26 +241,45 @@ function DatasetOverviewPanel(props) {
...
+              theme.transitions.create('width', {
+                easing: theme.transitions.easing.sharp,
+                duration: theme.transitions.duration.leavingScreen,
+              }),
+            overflowX: 'hidden',
+            width: '0px',
+            minWidth: '0px',
+            [theme.breakpoints.up('sm')]: {
+              width: '0px',
+            },
```
snf2ye: For more complicated styles like this, it may make sense to pull it into a styled component to help with code readability. But, this styling is still only used in one spot so I'm a torn over which is better (leave as sx prop vs. pull into styled component)

src/components/UserList.tsx
```diff
@@ -1,75 +1,53 @@
...
     <Accordion defaultExpanded={defaultOpen}>
       <AccordionSummary
-        expandIcon={<ExpandMore className={classes.expandIcon} />}
-        className={classes.header}
+        expandIcon={<ExpandMore sx={{ color: 'primary.main' }} />}
+        sx={{
+          fontSize: '14px',
+          lineHeight: '22px',
+          fontWeight: '600',
+          color: 'primary.main',
```
fboulnois, line 24: Do these need to `import theme from 'modules/theme';` and use `theme.palette.primary.main` instead like in `src/components/DatasetView.tsx`? If so, a number of components in this PR have this issue with `color`.

src/components/UserList.tsx
```diff
@@ -1,75 +1,53 @@
...
+          paddingTop: canManageUsers ? 0 : undefined,
+        }}
       >
         <ManageUsersView removeUser={canManageUsers ? removeUser : undefined} users={users} />
-        {users.length === 0 && <Typography className={classes.noUsers}>(None)</Typography>}
+        {users.length === 0 && (
+          <Typography
+            sx={{
+              fontStyle: 'italic',
+              color: 'error.contrastText',
```
fboulnois, line 36: The original component had `colorPrimary` as well, should this be captured?

fboulnois (approved):
One main question, otherwise lgtm:

src/components/dataset/overview/DatasetOverviewPanel.jsx
fboulnois: Since the refactoring is being done in stages, we can always pull this out into a separate style in the future.

src/components/UserList.tsx
snf2ye: Great question - I believe it is correct as is - Here's[ the documentation](https://mui.com/system/getting-started/the-sx-prop/?srsltid=AfmBOor9PjYY4oizLrQBEB2KSea8tGUdEU5VACrtpu6Js6RSEbVKr837#basic-example) and the basic example has something very similar. We have ["ThemeProvider" set up in the index](https://github.com/DataBiosphere/jade-data-repo-ui/blob/829daf5eb149d9896b9903ce9cf4c297e8090104/src/index.jsx#L162), so the theme should be available to all components in the repo. 

I'm possibly not referencing the theme correctly in DatasetView, but I wasn't quite sure how to do the expansion of the custom mixin.

src/components/UserList.tsx
snf2ye: Oh I missed that, thanks for catching it. 
It appears the the "(None)" text is correctly picking up the contrast text color, so I don't think we should need both.

src/components/UserList.tsx
fboulnois: Does that also mean that it isn't necessary for `DatasetView.tsx`?

src/components/DatasetView.tsx
```diff
@@ -40,56 +23,51 @@ interface IProps extends WithStyles<typeof styles> {
...
-              />
-            )}
-          </div>
-        </div>
-      </div>
-    );
-  },
-);
+  return (
+    <Box sx={{ display: 'flex', justifyContent: 'center', marginTop: '1em' }}>
```
s-rubenstein: If you want to reduce the diff here, and keep the object abstractions, this could be `<Box sx={styles(theme).wrapper}>` just as easily, and that way we keep our object abstractions while moving to the `sx` property?

To do this you would need to have `styles` be a raw object rather than a method (or call it once if we have a theme we could pass), but that should be okay

rushtong (approved):
👍🏽

src/components/DatasetView.tsx
```diff
@@ -1,34 +1,17 @@
...
-import { CustomTheme } from '@mui/material';
 import { TdrState } from 'reducers';
 import { OrderDirectionOptions } from 'reducers/query';
+import theme from 'modules/theme';
 import DatasetTable from './table/DatasetTable';
 
 import { DatasetRoles } from '../constants';
 
-const styles = (theme: CustomTheme) => ({
-  wrapper: {
```
rjohanek, line 15: so just a general comment, this is moving us away from having separate style objects and towards inline styling. Is that the principle that we want to follow? I thought that separate objects were preferred because they are reusable and easier to maintain. But maybe that isn't the case. What are the guidelines from the UI working group on this?

snf2ye: My goal in this PR is to use best practices for Material UI. We're moving from an old Material UI concept (@mui/styles), to a new one - sx prop + styled components. I don't think Terra UI uses Material UI, but I'm glad to still run this by them. 

RE: separate objects that are reusable/easier to maintain: We should use styled components for this case. The "sx prop" should be used for styles that are used in one spot, styled components are reusable and then we use the global theme to share styling variables. I think this is a helpful guide here: https://mui.com/material-ui/customization/how-to-customize

src/components/DatasetView.tsx
snf2ye: That's a really interesting idea, and I'm torn about it. 
I think I'm less interested in making the diff smaller (copilot is doing the heavy lifting here to move things around), and want to think about how we want this to look long term. We often use existing code as a model for new code, so what do we want our best practices to be going forward? 
I think I generally like the sx prop styles to be inline for a small set of changes, but we may want to consider using styled components more frequently to keep a pseudo object abstraction. I could actually build that into the copilot request - like "if there are more than 3 css styles, use a styled component instead of the sx prop".

s-rubenstein: That's my ultimate preference as well. 3 css styles is fine for inline, more gets bulky. I also think if we have a often reused width or something - with lots of constants reused, that I'd abstract.

snf2ye: Turns out that its tricky to get copilot to handle these two cases 🫠 It's not great at counting, apparently (It would catch some components with >3 styles, but not all). 
We may have to do this in two steps (convert to sx prop or styled and then move the rest). I think I'm also having better luck with a newer gpt model in vs code (as opposed to intellij).

src/components/dataset/overview/DatasetOverviewPanel.jsx
```diff
@@ -231,14 +196,26 @@ function DatasetOverviewPanel(props) {
...
               <Accordion defaultExpanded={false}>
                 <AccordionSummary
-                  expandIcon={<ExpandMore className={classes.expandIcon} />}
-                  className={classes.header}
+                  expandIcon={<ExpandMore sx={{ color: 'primary.main' }} />}
+                  sx={{
+                    fontSize: '14px',
+                    lineHeight: '22px',
+                    fontWeight: '600',
+                    color: 'primary.main',
```
rjohanek, line 204: just curious, why is this not a styled component? Is something like this just a symptom of copilot not counting very well?

src/components/DatasetView.tsx
```diff
@@ -1,34 +1,18 @@
...
-    ...theme.mixins.containerWidth,
-  },
-  title: {
-    color: theme.palette.primary.main,
-    fontSize: '54px',
-    lineHeight: '66px',
-    paddingBottom: theme.spacing(8),
-  },
-});
+const Container = styled('div')(({ theme }: { theme: CustomTheme }) => theme.mixins.containerWidth);
```
rjohanek: for variables like container that are used in multiple files do we want to define them in one location and import them?

rjohanek (approved):
I think this looks good overall, just a couple non-blocking questions

src/components/HomeView.tsx
```diff
@@ -18,76 +15,42 @@ import {
...
-      marginLeft: theme.spacing(2),
-      height: '45px',
-      textTransform: 'none',
-    },
-    buttonIcon: {
-      marginRight: 6,
-      fontSize: '1.5rem',
-    },
-  });
+const RootContainer = styled(Box)({
```
s-rubenstein, line 25: I really like this approach - very clean.

src/components/UserList.tsx
```diff
@@ -1,75 +1,48 @@
...
       >
         {typeOfUsers}
       </AccordionSummary>
-      <AccordionDetails
-        data-cy="user-email"
-        className={clsx({
-          [classes.canManageUsers]: canManageUsers,
-        })}
-      >
+      <AccordionDetails data-cy="user-email" sx={canManageUsers ? { pt: 0 } : undefined}>
```
s-rubenstein: Is `pt: 0` the same as not rendering? Should that be paddingTop?

src/components/UserList.tsx
```diff
@@ -1,75 +1,48 @@
...
         <ManageUsersView removeUser={canManageUsers ? removeUser : undefined} users={users} />
-        {users.length === 0 && <Typography className={classes.noUsers}>(None)</Typography>}
+        {users.length === 0 && (
+          <Typography
+            sx={{
+              fontStyle: 'italic',
+              color: 'error.contrastText',
+              pb: '8px',
```
s-rubenstein: Same with `pb`? Is that shorthand for paddingBottom? If so, can we use the full CSS version?

src/components/dataset/overview/DatasetOverviewPanel.jsx
```diff
@@ -259,7 +236,7 @@ function DatasetOverviewPanel(props) {
                         </dt>
                       </dl>
                       {i < dataset.schema.relationships.length - 1 && (
-                        <Divider className={classes.divider} />
+                        <Divider sx={{ mt: '14px', mb: '14px' }} />
```
s-rubenstein: Same with `mt` and `mb` - I'd prefer the longer versions even if that is functional shorthand.

s-rubenstein (commented):
I am seeing a lot of `p` or `m` as a shorthand for margin/padding - I don't know if that is an affordance that material provides, but it seems less readable to me than the CSS property - I would rather we stick with the CSS property versions? I commented a few examples

src/components/dataset/overview/DatasetOverviewPanel.jsx
snf2ye: Yeah, it really seems like copilot can't compare numbers all that well 😭  

Here's copilot's notes from this update:

> I'll help migrate the DatasetOverviewPanel from JSS to use styled components and sx props. Let's analyze the styles:
> 
> 1. More than 3 styles:
> - drawer (5 styles)
> - drawerOpen (4 styles with transitions)
> - drawerClose (5 styles with transitions and breakpoints)
> 
> 2. Convert to sx props:
> - root (1 style)
> - helpOverlayCloseButton (1 style)
> - helpText (1 style)
> - drawerPosition (1 style)
> - **header (4 styles)**
> - expandIcon (1 style)
> - noRelationships (3 styles)
> - divider (2 styles)
> - fullViewButton (1 style)
> - fullViewText (1 style)
>
> Key changes made:
> - Removed withStyles HOC and styles object
> - Added styled import from @mui/material/styles
> - Created StyledDrawer component for complex drawer styling
> - Converted all simple styles to sx props
> - Added Box component as root container
> - Removed classes prop and references
> - Simplified component exports
> - Used theme spacing units where appropriate
> - Maintained all functionality while modernizing the styling approach
> - The component maintains the same styling and behavior while using modern MUI patterns.

.

It clearly notes that the "header" class has 4 styles and notes that the first group is for components with more than 3 styles, but still groups it with the sx props. So, I really don't understand why it groups it with the sx props group. But, it still does catch the more complex styles and generally seems to move the components with lots of styles to styled components. I think as long as it is loosely following this rule, then it is okay.

src/components/DatasetView.tsx
snf2ye: Excellent point - I think we should use this as an opportunity for identifying components that could be shared. 

I made this a shared component and then realized that "containerWidth" is just setting width to 100%. That seems unnecessary to be a shared theme, so I removed these styled components and the theme property and instead am setting an sx prop here.

snf2ye:
> I am seeing a lot of `p` or `m` as a shorthand for margin/padding - I don't know if that is an affordance that material provides, but it seems less readable to me than the CSS property - I would rather we stick with the CSS property versions? I commented a few examples

@s-rubenstein  Yeah, this is specifically a feature of the sx prop for mui components: https://mui.com/system/getting-started/the-sx-prop/#spacing
I'm glad to switch these back to the CSS property. If we were really all-in on mui styling, it seems possible that these shortcuts could help with readability of bigger sx props. But, css properties are more readable for me.
Copilot was randomly deciding between using the shorthands and the regular css properties, so we'll have to watch out for this happening in future PRs. We can add a note in the copilot prompt.

src/components/UserList.tsx
snf2ye: Yeah, I'll switch back to paddingTop

s-rubenstein (approved):
LGTM - there are a number of places where we have `14px` vs `14` for styles, but tbh I think that's an existing issue and out of scope to unify to one or the other here.

snf2ye:
> LGTM - there are a number of places where we have `14px` vs `14` for styles, but tbh I think that's an existing issue and out of scope to unify to one or the other here.

Oh, good catch - That's something I paid attention to in the first round, but forgot about in the second round. It's actually important that we pay attention to this for spacing styles because of how mui does theme. And, unfortunately, it looks different for styled vs. sx prop 🫠 I'll go back through the changes again.

### sx Prop spacing

With the sx prop, if you leave off "px", then it defaults to it theme spacing, which multiples the value by 8px. So, `sx={margin: 2}` is actually equivalent to `sx={margin: '16px'}`
image

### Styled spacing
With styled components, you have to add `theme.spacing(2)` to use their multiplier shorthand.
styled component: 
* `margin: 2` = `margin: 2px`
* `margin: theme.spacing(2)` = `margin: 16px` = `margin: 16`

image

### Other sx prop styles
To make things even more confusing, these rules appear to not be consistent across sx props. Like sizing (so, height, width, etc.) does NOT default to the theme sizing, but grid does. These guides on sx prop and styled components are vital to getting through these changes. 
https://mui.com/system/getting-started/the-sx-prop
https://mui.com/system/styled/

PR #1731, opened 2025-01-03
DT-1125: Migrate away from @mui/styles Part 2 - ThemeProvider to use @mui/system

snf2ye:
### Background
More details about this migration here: https://github.com/DataBiosphere/jade-data-repo-ui/pull/1730

Breaking out this change to help simplify other PRs. 

### Included Changes
Switch references from `@mui/styles` to `@mui/material/styles`
______
From migration guide: https://mui.com/material-ui/migration/v5-style-changes/#update-themeprovider-import

image

______

### Note
I don't totally understand why this appears to still be working in combination with the sx components. We may need to eventually migrate away from this as well. But, I think that this ThemeProvider is providing the theme even in the changes in [this PR](https://github.com/DataBiosphere/jade-data-repo-ui/pull/1730/files) to use the sx prop because things break if the theme is not referenced correctly. 

But, I want to go ahead and remove all references directly to @mui/styles.

### DIFF
--- a/src/components/AppBreadcrumbs/AppBreadcrumbs.test.tsx
+++ b/src/components/AppBreadcrumbs/AppBreadcrumbs.test.tsx
@@ -3,7 +3,7 @@ import { mount } from 'cypress/react';
 import { Router } from 'react-router-dom';
 import history from 'modules/hist';
 import globalTheme from 'modules/theme';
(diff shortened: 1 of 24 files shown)


fboulnois (approved):
This one seems pretty straightforward 👍

s-rubenstein (approved):

PR #1733, opened 2025-01-08
[DT-1122] Apply `zizmor` suggestions

fboulnois:
## Addresses

https://broadworkbench.atlassian.net/browse/DT-1122

## Summary

- Do not persist credentials if not needed
- Use shell variable expansion instead of [template expansion](https://woodruffw.github.io/zizmor/audits/#template-injection)

### DIFF
--- a/.github/workflows/dev-image-update.yaml
+++ b/.github/workflows/dev-image-update.yaml
@@ -13,7 +13,7 @@ on:
       - 'Dockerfile'
       - '.github/workflows/dev-image-update.yaml'
 env:
-  gcr_google_project: 'broad-jade-dev'
+  GCR_GOOGLE_PROJECT: 'broad-jade-dev'
   google_sdk_version: '290.0.1'
   DEV_PROJECT: broad-jade-dev
   # This must be defined for the bash redirection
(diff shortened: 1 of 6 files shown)


.github/workflows/dev-image-update.yaml
```diff
@@ -69,21 +70,23 @@ jobs:
...
+            gcr.io/${gcr_google_project}/jade-data-repo-ui:"${GCR_TAG}" \
+            gcr.io/${gcr_google_project}/jade-data-repo-ui:"${PREV_TAG}" --quiet
           gcloud container images \
             add-tag \
-            gcr.io/${{ env.gcr_google_project }}/jade-data-repo-ui:"${GCR_TAG}" \
-            gcr.io/${{ env.gcr_google_project }}/jade-data-repo-ui:"${GCR_TAG}-develop" --quiet
+            gcr.io/${gcr_google_project}/jade-data-repo-ui:"${GCR_TAG}" \
+            gcr.io/${gcr_google_project}/jade-data-repo-ui:"${GCR_TAG}-develop" --quiet
+        env:
+          PREV_TAG: ${{ steps.uiprevioustag.outputs.tag }}
```
snf2ye: Were you able to test these changes? I'm a little worried that it won't pick up on these variables with the changes - I've found it to be really picky in the past. 
I think I've been able to test a subset of this action by commenting out the bumperstep, helm_tag_bump an cherry_pick_image_to_production_gcr tasks and then kicking off the run on your branch.

rushtong (approved):
👍🏽

snf2ye (approved):

.github/workflows/dev-image-update.yaml
```diff
@@ -69,21 +69,23 @@ jobs:
           DISABLE_ESLINT_PLUGIN: true
         run: |
           rm -rf jade-dev-account.pem
-          docker build -t gcr.io/${{ env.gcr_google_project }}/jade-data-repo-ui:${GCR_TAG} --build-arg CACHEBUST=$(date +%s) .
+          docker build -t gcr.io/${gcr_google_project}/jade-data-repo-ui:${GCR_TAG} --build-arg CACHEBUST=$(date +%s) .
```
pshapiro4broad: As I understand it, the reason for this fix is to replace github workflows template expansion with bash shell expansion, since bash shell supports quoting to prevent code injection. But this new expression doesn't use bash quoting, so I believe it would also be vulnerable. Adding bash quoting (as recommended by `shellcheck`) produces
```suggestion
          docker build -t gcr.io/"${gcr_google_project}"/jade-data-repo-ui:"${GCR_TAG}" --build-arg CACHEBUST="$(date +%s)" .
```

It's probably good to apply all recommended fixes, although in these cases I think you could use logic to determine that the expansions only refer to objects which we control and therefore wouldn't be a possible injection attack vector.

.github/workflows/dev-image-update.yaml
fboulnois: Yup, I confirmed it works!

PR #1734, opened 2025-01-10
[DT-1115]Flaky TDR UI integration test

rjohanek:
Summary:
Adds a thirty second timeout period to the before each of the first integration test suite because we occasionally see issues where the page is still loading for the first test.
Note: Another approach we could take is adding a rerun to the integration tests in github actions
Testing:
Rerunning test runs a few times to see if we see any failures.

### DIFF
--- a/cypress/integration/datasetSharing.spec.js
+++ b/cypress/integration/datasetSharing.spec.js
@@ -9,8 +9,7 @@ describe('test dataset sharing', () => {
     );
 
     cy.visit('/login/e2e');
-    cy.get('#tokenInput').type(Cypress.env('GOOGLE_TOKEN'), {
-      timeout: 30000,
+    cy.get('#tokenInput', { timeout: 60000 }).type(Cypress.env('GOOGLE_TOKEN'), {
       log: false,
       delay: 0,
     });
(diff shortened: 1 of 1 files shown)


cypress/integration/datasetSharing.spec.js
```diff
@@ -16,15 +16,22 @@ describe('test dataset sharing', () => {
...
-    cy.get('[placeholder="Search keyword or description"]').type('NonStewardDataset');
-    cy.contains(/NonStewardDataset/g).should('be.visible');
-    cy.contains(/NonStewardDataset/g).click();
-    cy.get('a > .MuiButtonBase-root').click();
+  it(
+    'does not have manage users buttons when the user is not a steward',
+    // this is the first integration test run, and it occasionally fails on setup
+    {
+      retries: 2,
+    },
```
snf2ye: Do you know if it is still going to retry if fails in the "beforeEach" part of the test?

rjohanek: Oh no, according to the documentation a failure in the before or after hook will not trigger a rerun. I'll look into another solution. Good catch.

pshapiro4broad (approved):

fboulnois (approved):
👍

PR #1735, opened 2025-01-10
[DT-1001] select billing profile

s-rubenstein:
https://github.com/user-attachments/assets/671d0f84-ab02-4eba-9c3a-6a01b69de9e5

### DIFF
--- a/src/components/common/FullViewSnapshotButton.test.tsx
+++ b/src/components/common/FullViewSnapshotButton.test.tsx
@@ -19,9 +19,6 @@ const initialState = {
       name: 'Test Snapshot',
     },
   },
-  profiles: {
-    profiles: [{ id: 'profile1', name: 'Test Profile 1' }],
-  },
   user: _.cloneDeep(initialUserState),
   query: _.cloneDeep(initialQueryState),
   router: { location: {} },
--- a/src/components/common/FullViewSnapshotButton.test.tsx
+++ b/src/components/common/FullViewSnapshotButton.test.tsx
@@ -82,33 +82,15 @@ describe('FullViewSnapshotButton', () => {
   });
 
   describe('FullViewSnapshotButton component without permission', () => {
-    beforeEach(() => {
+    it('Button is disabled and has tooltip with the no access message', () => {
       const dataset = { defaultProfileId: 'profile2' };
       const profiles: BillingProfileModel[] = [];
       mountFullViewSnapshotButton(dataset, profiles);
-    });
-
-    it('Button is disabled and has tooltip with the no access message', () => {
       cy.get('button').should('be.disabled');
       cy.get('button').trigger('mouseover', { force: true });
       cy.contains('You do not have access to any billing profiles to create a snapshot').should(
         'be.visible',
       );
-    });
-  });
-
-  describe('FullViewSnapshotButton component without default billing profile', () => {
-    beforeEach(() => {
-      const dataset = {};
-      const profiles = [
-        { id: 'profile1', profileName: 'profile1' },
-        { id: 'profile2', profileName: 'profile2' },
-      ];
-      mountFullViewSnapshotButton(dataset, profiles);
-    });
-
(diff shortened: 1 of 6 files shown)


src/components/common/FullViewSnapshotButton.test.tsx
```diff
@@ -18,11 +22,17 @@ const initialState = {
   profiles: {
     profiles: [{ id: 'profile1', name: 'Test Profile 1' }],
```
rjohanek: so do you still need this ?

rjohanek (commented):
this looks good, but looking at the ticket, I think we talked about adding the name and description of the snapshot to the modal as well. Did we still want to do that as part of this work?

s-rubenstein:
@rjohanek oh I see that now - we should work on updating ticket names when we make those sorts of changes as well.

rushtong (approved):
Looks reasonable to me 👍🏽

src/components/common/FullViewSnapshotButton.test.tsx
s-rubenstein: Yes, because it provides a default if it is undefined.

s-rubenstein:
https://github.com/user-attachments/assets/2f995a57-1054-4110-abb5-9595670afd6e

src/components/common/FullViewSnapshotButton.test.tsx
```diff
@@ -42,46 +49,66 @@ describe('FullViewSnapshotButton', () => {
...
+      mountFullViewSnapshotButton(dataset, profiles);
     });
 
     it('Button is disabled and has tooltip with the no billing profile message', () => {
-      cy.get('button').should('be.disabled');
-      cy.get('button').trigger('mouseover', { force: true });
-      cy.contains('There is no default billing profile associated with this dataset.').should(
-        'be.visible',
-      );
+      cy.get('button').click();
```
rjohanek: can you explain the changes to this test? Also do we want to add tests for the name and description functionality?

s-rubenstein: The changes for the test are to accommodate the additional modal in the flow. We could add tests for name and description, but given the defaults are already populated I personally think its probably okay?

rjohanek: okay, I was thinking we could test that you can change name and description, but not a blocker. and for my first question, I meant specifically the 'Button is disabled and has tooltip with the no billing profile message' test. Maybe the test description just needs to be updated? If I'm understanding correctly, the body of the test is just a click, so I'm just wondering what this is testing now?

src/components/common/FullViewSnapshotButton.tsx
```diff
@@ -28,44 +37,145 @@ function FullViewSnapshotButton({
...
+          <FormLabel sx={{ fontWeight: 600, color: 'black' }} htmlFor="snapshot-name" required>
+            Snapshot Name
+          </FormLabel>
+          <TextField
+            fullWidth
+            margin="normal"
+            id="snapshot-name"
+            label="Snapshot Name"
+            value={snapshotName}
+            onChange={(e) => setSnapshotName(e.target.value)}
```
rjohanek, line 110: do we need to do input validation here for name and description because we are sending it to the backend?

s-rubenstein: The backend has validation - I checked other places and TDR UI doesn't currently validate inputs - it just pushes the problem to the API. We probably want clientside validation at some point, but that should be out of scope of this. (That's also when I would add more testing)

lstrano-broad:
LGTM! Thx

src/components/common/FullViewSnapshotButton.test.tsx
s-rubenstein: Oh, that's not a worthwhile test now, I'll delete it.

rjohanek (approved):
thanks!

PR #1751, opened 2025-02-06
DT-1152: Migrate Components from @mui/styles

rushtong:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-1152

## Summary
For each component, followed the following process:
* Use https://github.com/DataBiosphere/jade-data-repo-ui/pull/1730 as a model for the copilot conversion
* Apply prettier formatting
* Ensure that we import `styled` from `'@mui/material/styles';` instead of `'@mui/system';`. Copilot doesn't catch that.
* Refactor the cases where props are destructured inside the function definition. Instead, ensure that `props` is the passed in function argument and destructure on a separate line in the function. This helps avoid a `readonly` warning.
* Ensure that the props have `readonly` modifiers
* Add `// @ts-ignore` where necessary. I found that we needed to do this for a variety of theme-related fields.

Other notes:
* `Failure.jsx` is unused, but I'm not aware of the history for this so it is included as part of the migration.
* `InfoModal` - Need some guidance on how to find where this is in use.
* `TextContent` is visible on all tables

### AddUserAccess
Screenshot 2025-02-11 at 7 17 50 AM

### FullViewSnapshotButton
Screenshot 2025-02-11 at 7 21 38 AM

### LoadingSpinner
https://github.com/user-attachments/assets/385b98fd-8c72-414c-a473-27916fc08139

### TabPanel, TabWrapper
https://github.com/user-attachments/assets/4c50c9d0-80cf-4e9d-bda3-8c938dc987f7

### TerraTooltip
Screenshot 2025-02-11 at 7 31 30 AM

### DIFF
--- a/src/components/common/AddUserAccess.tsx
+++ b/src/components/common/AddUserAccess.tsx
@@ -1,7 +1,6 @@
 import React, { useState } from 'react';
 import _ from 'lodash';
-import { CustomTheme } from '@mui/material/styles';
-import { createStyles, WithStyles, withStyles } from '@mui/styles';
+import { styled } from '@mui/system';
 import {
   Typography,
   Autocomplete,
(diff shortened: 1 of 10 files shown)


src/components/common/AddUserAccess.tsx
```diff
@@ -167,24 +146,31 @@ function AddUserAccess({ classes, permissions, onAdd }: AddUserAccessProps) {
...
+        </Box>
+        <Box sx={{ paddingTop: '22px' }}>
           <Button
             variant="contained"
             color="primary"
             disableElevation
             disabled={!isEmail(addEmailInput) && !_.some(usersToAdd, (user) => isEmail(user))}
-            className={clsx(classes.button, classes.section)}
+            sx={{
+              backgroundColor: 'common.link',
```
s-rubenstein, line 157: This one is borderline. It only has three properties, but it also has the hover logic? What do you think about pulling this out?

s-rubenstein (approved):
LGTM pending the conversations about theme. I really would love to make that better, but maybe we do it the way that works right now and then file a follow up?

snf2ye (approved):
LGTM! Thank you!
I think we can make a new ticket to look into a more sharable theme util and do that in the future.

src/components/common/AddUserAccess.tsx
snf2ye: That would totally be reasonable to pull it into a styled component. 
The approach I took with my PR was just to go with whichever option (sx or styled) copilot picked. Especially for cases where a selection is borderline, I would like to defer to copilot's pick mostly to help expedite these PRs. We can always revisit a given pick if we're working with a component in the future.

rushtong: I'm happy with the current approach, and I think once more components are converted, we'll have a better handle on how to generify this to handle more cases. For instance, our `CustomTheme` + `useTheme()` approach did not work in [`TextContent.jsx`](https://github.com/DataBiosphere/jade-data-repo-ui/pull/1751/files#diff-ef8e72215eb7c2d427cb87b04e2301402dc3537ad2735400e68b0f3c4abae68eR7) due to the `span` as an argument to `styled`.

PR #1752, opened 2025-02-06
[DT-1171] Add auth domain to snapshot access

fboulnois:
## Addresses

https://broadworkbench.atlassian.net/browse/DT-1171

## Summary of changes

* Minor fix for front page
* Add auth domain to dataset snapshot create

Screenshot 2025-02-06 at 2 01 29 PM

## Testing Strategy

Unit tests and created snapshots with auth domains manually.

### DIFF
--- a/src/components/dataset/data/sidebar/panels/ShareSnapshot.jsx
+++ b/src/components/dataset/data/sidebar/panels/ShareSnapshot.jsx
@@ -18,6 +18,7 @@ import { MoreVert } from '@mui/icons-material';
 import { isEmail } from 'validator';
 import { createSnapshot } from 'actions/index';
 import SnapshotAccess from 'components/snapshot/SnapshotAccess';
+import AuthDomain from 'src/components/snapshot/AuthDomain';
 
 const drawerWidth = 600;
 const sidebarWidth = 56;
@@ -95,6 +96,7 @@ export class ShareSnapshot extends React.PureComponent {
       anchor: null,
       hasError: false,
       errorMsg: '',
+      authDomain: undefined,
     };
   }
 
@@ -105,6 +107,10 @@ export class ShareSnapshot extends React.PureComponent {
     onDismiss: PropTypes.func,
     readers: PropTypes.arrayOf(PropTypes.string),
     setIsSharing: PropTypes.func,
+  };
+
+  setAuthDomain = (domain) => {
+    this.setState({ authDomain: domain });
   };
 
   saveSnapshot = () => {
@@ -126,18 +132,21 @@ export class ShareSnapshot extends React.PureComponent {
         </Typography>
         <SnapshotAccess createMode={true} />
         {!isModal && (
-          <div className={classes.bottom}>
-            <Button
-              variant="contained"
-              color="primary"
-              disableElevation
-              className={clsx(classes.button, classes.section)}
-              onClick={this.saveSnapshot}
-              data-cy="releaseDataset"
-            >
-              Create Snapshot
-            </Button>
-          </div>
+          <>
+            <AuthDomain setParentAuthDomain={this.setAuthDomain} />
+            <div className={classes.bottom}>
+              <Button
+                variant="contained"
+                color="primary"
+                disableElevation
+                className={clsx(classes.button, classes.section)}
+                onClick={this.saveSnapshot}
+                data-cy="releaseDataset"
+              >
+                Create Snapshot
+              </Button>
+            </div>
+          </>
         )}
         {isModal && (
           <div className={classes.modalBottom}>
--- a/src/components/dataset/data/sidebar/panels/ShareSnapshot.jsx
+++ b/src/components/dataset/data/sidebar/panels/ShareSnapshot.jsx
@@ -102,11 +102,14 @@ export class ShareSnapshot extends React.PureComponent {
 
   static propTypes = {
     classes: PropTypes.object,
+    dataset: PropTypes.object,
     dispatch: PropTypes.func,
+    filterData: PropTypes.object,
     isModal: PropTypes.bool,
     onDismiss: PropTypes.func,
     readers: PropTypes.arrayOf(PropTypes.string),
     setIsSharing: PropTypes.func,
+    snapshotRequest: PropTypes.object,
   };
 
   setAuthDomain = (domain) => {
@@ -114,10 +117,16 @@ export class ShareSnapshot extends React.PureComponent {
   };
 
   saveSnapshot = () => {
-    const { dispatch } = this.props;
+    const { dispatch, snapshotRequest, dataset, filterData } = this.props;
     const { authDomain } = this.state;
     dispatch(
       snapshotCreateDetails({
+        name: snapshotRequest.name,
+        description: snapshotRequest.description,
+        mode: snapshotRequest.mode,
+        assetName: snapshotRequest.assetName,
+        dataset,
+        filterData,
         authDomain,
       }),
     );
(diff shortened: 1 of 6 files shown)


src/components/dataset/data/sidebar/panels/ShareSnapshot.jsx
```diff
@@ -107,6 +108,10 @@ export class ShareSnapshot extends React.PureComponent {
     setIsSharing: PropTypes.func,
   };
 
+  setAuthDomain = (domain) => {
+    this.setState({ authDomain: domain });
+  };
+
   saveSnapshot = () => {
     const { dispatch } = this.props;
     dispatch(createSnapshot(undefined));
```
fboulnois, line 133: How do I pass `authDomain` to `createSnapshot`? It isn't clear to me.

src/components/snapshot/AuthDomain.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+const JadeLink = styled('span')(({ theme }) => theme.mixins.jadeLink);
+
+type AuthDomainProps = {
+  dispatch: AppDispatch;
+  userGroups: Array<ManagedGroupMembershipEntry>;
+  setParentAuthDomain: (domain: string) => void;
+};
+
+function AuthDomain({ dispatch, userGroups, setParentAuthDomain }: AuthDomainProps) {
```
pshapiro4broad: I see a sonar warning for this line, did you try using its suggested fix?
```suggestion
function AuthDomain({ dispatch, userGroups, setParentAuthDomain }: Readonly<AuthDomainProps>) {
```

src/components/snapshot/AuthDomain.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  useOnMount(() => {
+    dispatch(getUserGroups());
+  });
+
+  return (
+    <>
+      <FormLabel sx={{ fontWeight: 700, color: '#333f52' }} htmlFor="authorization-domain">
+        Authorization Domain
+        <span style={{ fontWeight: 400, fontStyle: 'italic' }}> - (optional)</span>
```
pshapiro4broad, line 34: This line also triggers a sonar warning, the idea being that a reader might think that the newline here is preserved. Their fix doesn't seem great to me though so I'd probably skip this one.

src/components/snapshot/AuthDomain.tsx
```diff
@@ -0,0 +1,72 @@
...
+      </FormLabel>
+      <Box sx={{ mb: 1 }}>
+        Authorization Domains restrict data access to only specified individuals in a group and are
+        intended to fulfill requirements you may have for data governed by a compliance standard,
+        such as federal controlled-access data or HIPAA protected data. They follow all snapshot
+        copies and cannot be removed. For more details, see{' '}
+        <Link
```
pshapiro4broad, line 41: This uses `Link` but the other use of `JadeLink` uses `<a href`. Is there a reason to have different forms in these two cases?

Could a more general component be defined that parameterizes the `href` and prompt text and shares the rest of the code in both places?

src/components/snapshot/AuthDomain.tsx
```diff
@@ -0,0 +1,72 @@
...
+        sx={{ height: '2.5rem' }}
+        disabled={userGroups ? userGroups.length < 1 : true}
+        options={userGroups ? userGroups.map((group) => group.groupName) : []}
+        name="authorization-domain"
+        onSelectedItem={(event) => {
+          const authDomain = event.target.value;
+          setParentAuthDomain(authDomain);
+          setSelectedAuthDomain(authDomain);
+        }}
+        value={selectedAuthDomain || ''}
```
pshapiro4broad: This is also flagged by sonar, its suggested fix:
```suggestion
        value={selectedAuthDomain ?? ''}
```

src/components/snapshot/AuthDomain.tsx
```diff
@@ -0,0 +1,72 @@
...
+            When to use an Authorization Domain
+            <LaunchOutlined fontSize="small" />
+          </JadeLink>
+        </Link>
+        .
+      </Box>
+      <JadeDropdown
+        sx={{ height: '2.5rem' }}
+        disabled={userGroups ? userGroups.length < 1 : true}
+        options={userGroups ? userGroups.map((group) => group.groupName) : []}
```
pshapiro4broad: Since `userGroups` is not optional I don't think you need to check for `undefined` here.
```suggestion
        disabled={userGroups.length < 1}
        options={userGroups.map((group) => group.groupName)}
```

src/components/snapshot/AuthDomain.tsx
```diff
@@ -0,0 +1,72 @@
...
+          <JadeLink>
+            When to use an Authorization Domain
+            <LaunchOutlined fontSize="small" />
+          </JadeLink>
+        </Link>
+        .
+      </Box>
+      <JadeDropdown
+        sx={{ height: '2.5rem' }}
+        disabled={userGroups ? userGroups.length < 1 : true}
```
pshapiro4broad: I see that this is enabled even if there's only one element. Does the UI let you clear this field if the user selects the only item in the list? Or, do we need to add an element to the list to represent "No selection"?

src/components/snapshot/AuthDomain.tsx
```diff
@@ -0,0 +1,72 @@
...
+        options={userGroups ? userGroups.map((group) => group.groupName) : []}
+        name="authorization-domain"
+        onSelectedItem={(event) => {
+          const authDomain = event.target.value;
```
pshapiro4broad, line 59: I think if you support removing the selection, this would need to support that too?

pshapiro4broad (approved):
Looks OK, but I have a few questions

rushtong (approved):
LGTM 👍🏽

src/components/snapshot/AuthDomain.tsx
pshapiro4broad: I think the overall UX question here is, if there is no checkbox to toggle this feature on/off, then the user's selection of a group is how the option is toggled. So there needs to be a way to unselect a group after it's been selected to opt out. Also even if there are groups, a user doesn't need to select one of them, so we wouldn't want to select the first one by default.

Maybe the UI already does this, I'm not familiar with how `JadeDropdown` works.

fboulnois: I need to rebase on Shelby's changes to make the optional setting clearer.

src/components/snapshot/AuthDomain.tsx
fboulnois: I don't have a strong opinion on this, perhaps something we can discuss as a team?

src/components/snapshot/AuthDomain.tsx
fboulnois: That is fixed in Shelby's PR 👍

src/components/snapshot/AuthDomain.tsx
fboulnois: I'm not sure about this, since the `userGroups` loads during the mount step, so there is a possibility it is momentarily `undefined`?

src/components/snapshot/AuthDomain.tsx
pshapiro4broad: No strong opinion here either -- as a non-UI expert I'm usually looking at other code to see how to do something, and when I see things done in two (superficially?) different ways it can be confusing. Refactoring the code duplication also has the benefit that if a fix is needed, it only needs to be done in one place. On the other hand, it's not a much code and maybe this kind of small duplication is typical in the UI code.

src/components/snapshot/AuthDomain.tsx
pshapiro4broad: Since `initialUserState` in `user.ts` provides a default value for `userGroups` (`[]`), it should  never be `undefined` when `AuthDomain` is rendered.

src/components/snapshot/AuthDomain.tsx
fboulnois: I created a ticket for this since it's inconsistent across the repo.

PR #1757, opened 2025-02-21
[DDO-4001] Deprecate datarepo-helm (UI GHA)

em-may:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/DDO-4001

## Addresses

`datarepo-helm` deprecation.

## Summary of changes

Removes the GHA steps to bump the UI app version in the `datarepo-api` chart and adds a step to report new UI app versions to Sherlock instead. Additionally, since the `datarepo` umbrella Helm chart (and subcharts) are moving to `terra-helmfile`, these steps would touch a repo which will no longer be used (`datarepo-helm`).

Also moves updates the build, tag, and publish steps to report versions for branches/PRs similar to https://github.com/DataBiosphere/jade-data-repo/pull/1911, but with the addition of branch names tags in GAR. The logic is slightly different and some files have been renamed, but the overall effect is the same.

Example new commit image in GAR and its tags:

Captura de pantalla 2025-02-24 a la(s) 17 00 34

Also updates `actions/checkout@v3` to `actions/checkout@v4`.

## Testing Strategy

Tested in my BEE by setting Sherlock/Beehive to follow versions from this branch. The app version was propagated to the API deployment in k8s correctly and TDR came up without issues.

## Release timing

After the `datarepo` umbrella Helm chart has been moved to `terra-helmfile` and Sherlock has been pointed at it there instead of `datarepo-helm`. Doing so makes the `helmtagbump.yaml` workflow useless.


.github/workflows/dev-image-update.yaml
```diff
@@ -98,3 +94,29 @@ jobs:
...
+  report-to-sherlock:
+    name: Report App Version to DevOps
+    uses: broadinstitute/sherlock/.github/workflows/client-report-app-version.yaml@main
+    needs: [update_image, cherry_pick_image_to_production_gcr]
+    with:
+      new-version: ${{ needs.update_image.outputs.ui_image_tag }}
+      chart-name: datarepo-ui
+    permissions:
+      contents: read
+      id-token: write
```
snf2ye: It makes sense to me to add the report to sherlock, but I'm surprised we're ready to pull the helm_tag_bump action already. Would it help to report to both places for an interim time period?

em-may: For context, this PR (which is still WIP) will only be merged once the `datarepo` umbrella Helm chart is moved to `terra-helmfile` and Sherlock is tracking version updates from there, so `datarepo-helm` getting updates wouldn't matter at that point.

rushtong (approved):
Looks reasonable to me 👍🏽

.github/workflows/build.yaml
```diff
@@ -0,0 +1,144 @@
...
+    branches:
+      - develop
+    paths:
+      - '!*'
+      - 'src/**'
+      - 'package.json'
+      - 'cypress.json'
+      - 'cypress/**'
+      - 'Dockerfile'
+      - '.github/workflows/dev-image-update.yaml'
```
pshapiro4broad: This file no longer exists. It seems like it should refer to itself again, as it did before. Or maybe it should include all github workflows?
```suggestion
      - '.github/workflows/build.yaml'
```

.github/workflows/build.yaml
```diff
@@ -0,0 +1,144 @@
...
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
```
pshapiro4broad: What is the purpose of filtering based on type here? `lint.yml`, `test-e2e.yml`, `test-unit.yml` filter based on PR merge target branch
```
  pull_request:
    branches:
      - develop
```

`trivy.yml` didn't filter at all
```
on: [pull_request]
```

rushtong (approved):
👍🏽

pshapiro4broad (approved):
Looks OK, just a couple minor things

.github/workflows/build.yaml
em-may: Thanks, fixed.

.github/workflows/build.yaml
em-may: Fixed per our previous discussion in https://github.com/DataBiosphere/jade-data-repo/pull/1912

.github/workflows/trivy.yml
snf2ye, line 1: Why is trivy getting removed?

em-may: This workflow file is getting removed, but the check is still present and has been moved to `build.yaml`, so the pre-merge check will have to be temporarily removed to get the PR merged (since it was a job before and will be a step now, so to GH it's technically a different check) and then readded with the new name. There's no technical reason that it has to be done this way, but I'm used to having the check be integrated into the build job, and that way it's one less thing that runs in the event of a failed build, as having it separately will always run it.

PR #1761, opened 2025-02-26
[DDO-4001] Fix UI cherry-pick to public GCR

em-may:
Fixes the UI cherry-pick to public GCR step. The problem is that the dev SA doesn't have perms on the public GCR project, so the docker push fails.


fboulnois (approved):
👍

snf2ye (approved):

PR #1768, opened 2025-03-12
MUI Style the table directory

s-rubenstein:
Dataset Snapshots (Old)
Screenshot 2025-03-12 at 3 21 56 PM

Dataset Snapshots (New)
Screenshot 2025-03-12 at 3 44 12 PM

Dataset Table (Old)
Screenshot 2025-03-12 at 3 44 55 PM

Dataset Table (New)
Screenshot 2025-03-12 at 3 45 01 PM


s-rubenstein:
Job Table (Old):
Screenshot 2025-03-12 at 3 47 51 PM

Job Table (New):
Screenshot 2025-03-12 at 3 48 50 PM

(It changes from an X to a ! for failed - this is because of the move from FA to Material Icons)

s-rubenstein:
Journal Entry Table (old):
Screenshot 2025-03-12 at 3 53 30 PM

Journal Entry Table (new):
Screenshot 2025-03-12 at 3 53 50 PM

s-rubenstein:
Snapshot Access Request (old):
Screenshot 2025-03-12 at 3 56 46 PM

Snapshot Access Request (new):
Screenshot 2025-03-12 at 3 56 56 PM

s-rubenstein:
Snapshot Table (old):
Screenshot 2025-03-12 at 3 59 10 PM

Snapshot Table (new):
Screenshot 2025-03-12 at 3 59 21 PM

pshapiro4broad:
Some minor discrepancies I can see looking at before/after images:

- the new images are showing sort arrows on many column headers. Is that correct; did you click on those headers to cause the sorting to occur?
- in the new job table, the job ID is now bold and the line height is greater (possibly due to the icon change?)
- the last four images look OK but I couldn't open them up in the browser, I got an error saying `This private-user-images.githubusercontent.com page can’t be found`

src/components/table/DatasetSnapshotsTable.jsx
```diff
@@ -1,24 +1,20 @@
...
 import { connect } from 'react-redux';
 import LightTable from './LightTable';
 import { getDatasetSnapshots } from '../../actions';
 
-const styles = (theme) => ({
-  jadeLink: {
-    ...theme.mixins.jadeLink,
-  },
-});
+const JadeLink = styled('span')(({ theme }) => ({
```
pshapiro4broad: This is also defined in `AuthDomain.tsx`, and as`StyledLink` in `TextContent.tsx`. It's a one-line definition, but maybe it should be defined in one place?

src/components/table/JobTable.tsx
```diff
@@ -1,173 +1,177 @@
...
+      render: (row: any) => <span title={row?.submitted}>{moment(row?.submitted).fromNow()}</span>,
+      width: '10%',
+    },
+    {
+      label: 'Status',
+      name: 'job_status',
+      allowSort: false,
+      width: '10%',
+      render: (row: any) => {
+        const job = row as JobModel;
```
pshapiro4broad: Could this be changed to use `(row: JobModel)`? I see `render` defined in `DatasetTable` and `SnapshotTable` without using `any`.

src/components/table/JobTable.tsx
```diff
@@ -1,173 +1,177 @@
...
+        return (
+          <Box
+            sx={{ width: '100%', display: 'flex', alignItems: 'center' }}
+            title={statusMap[job.job_status]?.label}
+          >
+            {statusMap[job.job_status]?.icon(
```
pshapiro4broad: Instead of requiring `?` could `statusMap` provide a default icon and label?

src/components/table/StatusMapTypes.ts
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { CustomTheme } from '@mui/material/styles';
+
+interface StatusIconProps {
+  sx: React.CSSProperties;
+}
+
+export interface StatusMapItem {
```
pshapiro4broad: If this is only used in `JournalEntryTable`, does it need to be in its own file?

src/components/table/SearchTable.tsx
```diff
@@ -2,71 +2,67 @@ import React from 'react';
...
+      >
         <SearchSVG />
       </div>
       <InputBase
         placeholder="Search keyword or description"
-        classes={{
-          root: classes.inputRoot,
-          input: classes.searchInput,
+        sx={{
+          paddingLeft: '30px !important',
```
pshapiro4broad: The original used `${theme.spacing(3.7)}` instead of `30px`, did the original spacing no longer work here?

pshapiro4broad (approved):
Looks OK overall and I figure we can always fix things later too

src/components/table/StatusMapTypes.ts
s-rubenstein: Its used in `JobTable` and `JourneyEntryTable

s-rubenstein:
@pshapiro4broad good feedback! I am going to go back and see if I can figure some of these out. Please hold off on further reviews. (I have no idea what is going on with the images, though :O)

src/components/table/JobTable.tsx
s-rubenstein: So, when I do that, interestingly enough, it errors out on my IDE. But the other tables don't. As far as I can tell, there is no difference in the typing.

src/components/table/JobTable.tsx
s-rubenstein: I could have it be `statusMap[job.job_status] ? statusMap[job.job_status].icon... : undefined`? But because we are keying into an object with arbitrary strings, there is no trivial way to provide a default. Because the key could be anything.

src/components/table/JobTable.tsx
s-rubenstein: It compiles okay though.

src/components/WelcomeView.jsx
```diff
@@ -110,7 +107,12 @@ function WelcomeView({ terraUrl }) {
               target="_blank"
               rel="noopener noreferrer"
             >
-              <JadeLink>
+              <JadeLink
+                sx={{
+                  paddingTop: theme.spacing(2),
+                  paddingBottom: theme.spacing(2),
+                }}
```
snf2ye:
```suggestion
              <JadeLink
                sx={{
                  paddingTop: 2,
                  paddingBottom: 2,
                }}
```
We can use the sx shorthand here  (and then we do remove the reliance on useTheme, but I don't think we should need that anyway)

src/components/WelcomeView.jsx
```diff
@@ -89,6 +84,8 @@ const StyledLogoGrey = styled(LogoGrey)({
 });
 
 function WelcomeView({ terraUrl }) {
+  const theme = useTheme();
+
```
snf2ye:
```suggestion

```
Why do we need to call useTheme()? The custom theme should be piped through from the index page.

src/components/WelcomeView.jsx
```diff
@@ -127,8 +129,8 @@ function WelcomeView({ terraUrl }) {
           </a>
           <hr />
           <div>
-            <Box sx={{ paddingBottom: (theme) => theme.spacing(2) }}>
-              <Typography sx={{ fontWeight: '900', paddingBottom: (theme) => theme.spacing(2) }}>
+            <Box sx={{ paddingBottom: theme.spacing(2) }}>
+              <Typography sx={{ fontWeight: '900', paddingBottom: theme.spacing(2) }}>
```
snf2ye:
```suggestion
            <Box sx={{ paddingBottom: 2) }}>
              <Typography sx={{ fontWeight: '900', paddingBottom: 2 }}>
```

src/components/common/JadeLink.tsx
```diff
@@ -0,0 +1,10 @@
+import { styled } from '@mui/system';
+import { Box } from '@mui/material';
+
+export const JadeLink = styled(Box)(({ theme }) => ({
+  ...theme.mixins.jadeLink,
+}));
```
snf2ye:
```suggestion
const StyledLink = styled(Box)(({ theme }: { theme: CustomTheme }) => ({
  ...theme.mixins.jadeLink,
}));
```
I think we need this odd format in order to get the right custom theme type piped through

src/components/common/JadeLink.tsx
```diff
@@ -0,0 +1,10 @@
...
+
+export const JadeLinkInline = styled('span')(({ theme }) => ({
+  ...theme.mixins.jadeLink,
+}));
```
snf2ye:
```suggestion
const StyledLink = styled('span')(({ theme }: { theme: CustomTheme }) => ({
  ...theme.mixins.jadeLink,
}));
```
Same here

src/components/table/DatasetSnapshotsTable.jsx
```diff
@@ -37,15 +29,15 @@ class DatasetSnapshotsTable extends React.PureComponent {
...
+    const { loading, snapshotCount, snapshots } = this.props;
     const columns = [
       {
         label: 'Snapshot Name',
         name: 'name',
         allowSort: true,
         render: (row) => (
           <Link to={`/snapshots/${row.id}`}>
-            <span className={classes.jadeLink}>{row.name}</span>
+            <JadeLink>{row.name}</JadeLink>
```
snf2ye, line 40: Nice 👍  Glad to share this component

src/components/table/DatasetTable.tsx
```diff
@@ -1,122 +1,114 @@
...
-  datasetRoleMaps: { [key: string]: Array<string> };
-  datasetsCount: number;
-  filteredDatasetsCount: number;
-  handleFilterDatasets?: (
+interface IProps {
+  readonly datasets: Array<DatasetSummaryModel>;
+  readonly datasetRoleMaps: { [key: string]: Array<string> };
+  readonly datasetsCount: number;
+  readonly filteredDatasetsCount: number;
+  readonly handleFilterDatasets?: (
```
snf2ye, line 23: Apologies if I've asked this before, but why do we need readonly tag on the types?

src/components/table/JobTable.tsx
```diff
@@ -1,173 +1,177 @@
...
-        },
+  const columns: Array<TableColumnType> = [
+    {
+      label: 'Job ID',
+      name: 'id',
+      allowSort: false,
+      width: '20%',
+      render: (row: JobModel) => (
+        <div>
+          <SeeMoreLink onClick={() => handleSeeMoreOpen(row.id)} theme={theme}>
```
snf2ye: This seems like an anti-pattern to me: I don't think we should be passing around the theme.

src/components/table/JobTable.tsx
```diff
@@ -1,173 +1,177 @@
...
+      allowSort: false,
+      width: '10%',
+      render: (row: JobModel) => (
+        <Box
+          sx={{ width: '100%', display: 'flex', alignItems: 'center' }}
+          title={statusMap[row.job_status]?.label}
+        >
+          {statusMap[row.job_status]?.icon(
+            { sx: { fontSize: '1.2rem', marginRight: '10px' } },
+            theme,
```
snf2ye: Same here, why do we ned to pass through theme?

src/components/table/JournalEntryTable.tsx
```diff
@@ -1,150 +1,158 @@
...
+interface StatusMap {
+  CREATE: StatusMapItem;
+  UPDATE: StatusMapItem;
+  DELETE: StatusMapItem;
+}
+
+const statusMap: StatusMap = {
+  CREATE: {
+    icon: (props, theme) => (
+      <CheckCircle sx={{ ...props.sx, color: theme.palette.success.light }} />
```
snf2ye:
```suggestion
    icon: (props) => (
      <CheckCircle sx={{ ...props.sx, color:  success.light }} />
```
We can use the color shorthand here in the sx prop

src/components/table/JournalEntryTable.tsx
```diff
@@ -1,150 +1,158 @@
...
+    ),
+    label: 'Created',
   },
-  statusIconFailed: {
-    color: theme.palette.error.light,
+  UPDATE: {
+    icon: (props, theme) => (
+      <CheckCircle sx={{ ...props.sx, color: theme.palette.success.light }} />
```
snf2ye:
```suggestion
    icon: (props) => (
      <CheckCircle sx={{ ...props.sx, color: success.light }} />
```

src/components/table/JournalEntryTable.tsx
```diff
@@ -1,150 +1,158 @@
...
+    ),
+    label: 'Updated',
   },
-  detail: {
-    marginTop: 15,
+  DELETE: {
+    icon: (_props, _theme) => <Box>Unsupported</Box>,
```
snf2ye:
```suggestion
    icon: (_props) => <Box>Unsupported</Box>,
```

src/components/table/LightTable.tsx
```diff
@@ -309,52 +300,51 @@ function LightTable({
...
                         key={`${index}-${row[rowKey || 'id']}}`}
-                        className={clsx({
-                          [classes.row]: true,
-                          [classes.darkRow]: darkRow,
-                          [classes.lightRow]: !darkRow,
-                        })}
+                        sx={{
+                          backgroundColor: darkRow
+                            ? theme.palette.lightTable.cellBackgroundDark
+                            : theme.palette.lightTable.callBackgroundLight,
```
snf2ye:
```suggestion
                            ? lightTable.cellBackgroundDark
                            : lightTable.callBackgroundLight,
```

snf2ye (commented):
I haven't quite made it through everything, but I'd be curious to talk more about theming. It's been a pain with every MUI PR that has come through and I'm worried that there will be a bunch of different patterns present in the code base.

src/components/table/LightTable.tsx
s-rubenstein: This wouldn't work - it is accessing the property and needs the full path. Unless I could just use the string `lightTable.cellBackgroundDark`?

src/components/common/Ellipsis.tsx
```diff
@@ -0,0 +1,11 @@
+import { styled } from '@mui/system';
+import { Link } from 'react-router-dom';
+import { CustomTheme } from '@mui/material/styles';
+
+export const EllipsisLink = styled(Link)(({ theme }) => ({
+  ...(theme as CustomTheme).mixins.ellipsis,
+}));
```
snf2ye, line 8: Oh nice! This allows the format to match between sx props and styled components.

snf2ye (approved):
Given the linter is a little happier, LGTM! Thanks for talking through the theme stuff with me, I think you got to a much better solution that will be easy to reproduce in future PRs! 🎉

PR #1769, opened 2025-03-13
DT-1186: Fix server error display

pshapiro4broad:
Fix an issue where the TDR server error table wouldn't render properly. This table is shown when the TDR index page is initially loaded if the server `/status` response indicates an error. It's possible this has been broken for a while.

Before

before

After

after1

Note that this error page is only shown if a critical subsystem is down. If a non-critical subsystem is down (such as DUOS), no error is shown.

If the TDR host itself isn't responding, a message is shown instead

all down

### DIFF
--- a/src/components/ServerErrorView.jsx
+++ b/src/components/ServerErrorView.jsx
@@ -30,25 +30,23 @@ const MainContent = styled(Box)(({ theme }) => ({
 }));
 
 function ServerErrorView({ status }) {
-  let systemsCount = 0;
   let systemRows = [];
   if (status.apiIsUp) {
     const { systems } = status.serverStatus;
-    systemsCount = Object.keys(systems).length;
-    systemRows = Object.keys(status.serverStatus.systems).map((member) => ({
+    systemRows = Object.keys(systems).map((member) => ({
       system: member,
-      system_is_up: status.serverStatus.systems[member].ok,
+      system_is_up: systems[member].ok,
     }));
   }
 
   const columns = [
     {
       label: 'System',
-      property: 'system',
+      name: 'system',
     },
     {
       label: 'Status',
-      property: 'system_status',
+      name: 'system_status',
       render: (row) => {
         if (row.system_is_up) {
           return '✅';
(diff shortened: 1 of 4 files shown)


src/index.jsx
```diff
@@ -109,7 +109,7 @@ function bootstrap() {
 function render(Component) {
   const root = document.getElementById('react');
   const { configuration } = store.getState();
-  const isGoogleAuthority = configuration.configObject.authorityEndpoint.startsWith(
+  const isGoogleAuthority = configuration.configObject.authorityEndpoint?.startsWith(
```
pshapiro4broad, line 112: Without this change, the main page would fail to render if the server returned an error state. The field `authorityEndpoint` is optional.

src/components/table/LightTable.tsx
```diff
@@ -289,7 +291,7 @@ function LightTable({
     0,
   );
   const effectiveTableWidth = _.isNaN(tableWidth) || !supportsResize ? '100%' : tableWidth;
-  const showPagination = (rows && rows.length > 0) || infinitePaging;
+  const showPagination = (rows?.length > 0 || infinitePaging) && !hidePagination;
```
pshapiro4broad, line 294: For the error table, pagination isn't needed so I added a setting to force the pagination controls to be hidden.

src/components/ServerErrorView.jsx
```diff
@@ -30,25 +30,23 @@ const MainContent = styled(Box)(({ theme }) => ({
...
-      system_is_up: status.serverStatus.systems[member].ok,
+      system_is_up: systems[member].ok,
     }));
   }
 
   const columns = [
     {
       label: 'System',
-      property: 'system',
+      name: 'system',
```
pshapiro4broad, line 45: This is the required fix. I did look at converting this file to typescript (which would've caught this error) but it wasn't obvious to me how to do it.

fboulnois (approved):
👍

rjohanek (approved):
this looks good! are there any tests for the server error view component? could you add some if not?

pshapiro4broad:
> @rjohanek are there any tests for the server error view component? could you add some if not?

 No tests, I'll look at adding a few.

PR #1772, opened 2025-03-17
[DT-400][DT-1401] Upgrade the lockfile to v3

fboulnois:
### Addresses

- https://broadworkbench.atlassian.net/browse/DT-400
- https://broadworkbench.atlassian.net/browse/DT-1401

### Summary

We are using `node` >= v20, so `npm` will be >= v9. This only updates the lockfile version, no other changes.


rjohanek (approved):
seems reasonable to me!

rushtong (approved):
v3 seems like much more efficient at a net `-10,190` lines 👍🏽

PR #1773, opened 2025-03-20
[DT-1348] Fix line height and fix the height of the modal

s-rubenstein:
https://github.com/user-attachments/assets/6c66a2dc-0240-4f2c-96ff-c41fa1a59a65

### DIFF
--- a/src/components/common/FullViewSnapshotModal.tsx
+++ b/src/components/common/FullViewSnapshotModal.tsx
@@ -1,4 +1,4 @@
-import { Box, Button, Dialog, DialogActions, DialogTitle } from '@mui/material';
+import { Box, Button, Dialog, DialogActions, DialogTitle, Grid } from '@mui/material';
 import { entries, isEmpty, now, uniq, without } from 'lodash';
 import React, { Dispatch, useEffect } from 'react';
 import { changePolicyUsersToSnapshotRequest, createSnapshot, snapshotCreateDetails } from 'actions';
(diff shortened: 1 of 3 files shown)


src/components/common/FullViewSnapshotModal.tsx
```diff
@@ -148,13 +148,15 @@ function FullViewSnapshotModal(props: FullViewSnapshotModalProps) {
   };
 
   return (
-    <Dialog fullWidth maxWidth="sm" onClose={onDismiss} open={modalOpen}>
+    <Dialog fullWidth maxWidth="sm" onClose={onDismiss} open={modalOpen} sx={{ height: '700px' }}>
```
fboulnois, line 151: Is there any way to make the scrollbar more obvious?

fboulnois (approved):
One question, otherwise lgtm:

src/components/snapshot/ManagedSnapshotAccess.tsx
```diff
@@ -8,9 +8,17 @@ import { PolicyModel, SnapshotRequestModelPolicies } from 'generated/tdr';
 import { connect } from 'react-redux';
 import { TdrState } from 'reducers';
 import { SnapshotRoles } from 'constants';
+import { styled } from '@mui/material/styles';
 
 export const transformRoleToCreatePolicy = (role: string): string => `${_.camelCase(role)}s`;
 
+const ManagedSnapshotAccessGrid = styled(Grid)(() => ({
+  my: 1,
```
rjohanek, line 16: what does this value mean?

rjohanek (approved):
looks good!

src/components/snapshot/ManagedSnapshotAccess.tsx
s-rubenstein: It's shorthand for `marginY` which is shorthand for `marginTop` and `marginBottom`. I personally don't like the shortening, but it was already used by this code.

src/components/common/FullViewSnapshotModal.tsx
s-rubenstein: Not really - scrollbars are dictated by the browser. We could make it so the scrollbar always shows by making it `overflowY: 'scroll'` rather than `overflowY: 'auto'`, but then that would mean we have a disabled scrollbar even if there is no scrolling on the content.

PR #1775, opened 2025-03-24
[DT-1205]DC inherit: On datasets, if inheritSteward is true, add messaging to data custodian role area

rjohanek:
Summary:
Increase visibility for new inherit steward feature by adding messaging on the roles and memberships tab for a dataset.

Changes:
In UserList: add optional message. 
In DatasetAccess: pass in message to UserList if inherit steward is enabled.

Testing:
Add DatasetAccess.test.tsx to check basic rendering and conditional rendering of the new message.

Visual Aids:
When enabled: 
Screenshot 2025-03-24 at 1 15 04 PM

When disabled: 
Screenshot 2025-03-24 at 1 17 54 PM

### DIFF
--- a/src/components/UserList.tsx
+++ b/src/components/UserList.tsx
@@ -5,12 +5,12 @@ import Typography from '@mui/material/Typography';
 import ManageUsersView from './ManageUsersView';
 
 interface UserListProps {
-  message?: string;
-  canManageUsers: boolean;
-  defaultOpen?: boolean;
-  removeUser?: (removableEmail: string) => void;
-  typeOfUsers: string;
-  users: Array<string>;
+  readonly message?: string;
+  readonly canManageUsers: boolean;
+  readonly defaultOpen?: boolean;
+  readonly removeUser?: (removableEmail: string) => void;
+  readonly typeOfUsers: string;
+  readonly users: Array<string>;
 }
 
 function UserList({
--- a/src/components/UserList.tsx
+++ b/src/components/UserList.tsx
@@ -1,10 +1,11 @@
 import React from 'react';
-import { Accordion, AccordionDetails, AccordionSummary, Box } from '@mui/material';
+import { Accordion, AccordionDetails, AccordionSummary } from '@mui/material';
 import { ExpandMore } from '@mui/icons-material';
 import Typography from '@mui/material/Typography';
 import ManageUsersView from './ManageUsersView';
 
 interface UserListProps {
+  message?: string;
   canManageUsers: boolean;
   defaultOpen?: boolean;
   removeUser?: (removableEmail: string) => void;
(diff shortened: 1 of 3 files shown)


src/components/UserList.tsx
```diff
@@ -28,6 +36,9 @@ function UserList({ canManageUsers, defaultOpen, removeUser, typeOfUsers, users
         {typeOfUsers}
       </AccordionSummary>
       <AccordionDetails data-cy="user-email" sx={canManageUsers ? { paddingTop: 0 } : undefined}>
+        {message && (
+          <Typography sx={{ marginTop: '-16px', paddingBottom: '12px' }}>{message}</Typography>
```
rjohanek, line 40: using a negative margin here because I don't want to edit the Accordion Summary component. I thought about having the Accordion Summary component have a margin that is conditional on if there is a message, but it would also have to be conditional on if it is expanded. And it seemed much simpler to just do it this way.

fboulnois (approved):
looks good 👍

src/components/dataset/DatasetAccess.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+    setUp(mockDataset);
+    cy.contains('Custodians').click();
+    cy.contains(
+      'All dataset custodians are stewards on all snapshots created from this dataset.',
+    ).should('exist');
+    cy.contains('custodian1@example.com').should('exist');
+    cy.contains('custodian2@example.com').should('exist');
+  });
+
+  describe('renders Dataset Custodian information correctly when inherit steward is false or undefined', () => {
```
s-rubenstein: You shouldn't need a separate `describe` here.

src/components/dataset/DatasetAccess.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+    [false, undefined].forEach((inheritSteward) => {
+      it('custodians should appear without message about inheriting stewardship', () => {
```
s-rubenstein: Could you make it so that the `it` described what value it was using?

src/components/dataset/DatasetAccess.tsx
```diff
@@ -30,14 +30,17 @@ function DatasetAccess(props: DatasetAccessProps) {
...
-  const { policies, userRoles } = props;
+  const { dataset, policies, userRoles } = props;
   const stewards = getRoleMembersFromPolicies(policies, DatasetRoles.STEWARD);
   const custodians = getRoleMembersFromPolicies(policies, DatasetRoles.CUSTODIAN);
   const snapshotCreators = getRoleMembersFromPolicies(policies, DatasetRoles.SNAPSHOT_CREATOR);
 
   const canManageStewards = userRoles.includes(DatasetRoles.STEWARD);
   const canManageUsers =
     userRoles.includes(DatasetRoles.STEWARD) || userRoles.includes(DatasetRoles.CUSTODIAN);
+  const message = dataset.inheritSteward
```
s-rubenstein, line 41: Because the else is `undefined`, I think this can be `dataset.inheritSteward && ...`, because otherwise it will be falsy

rjohanek: I tried that but then I would need to make message (in UserList) of type string or false which seemed weirder than just an optional string.

s-rubenstein (approved):

rushtong (approved):

PR #1776, opened 2025-03-24
Add dataset inheritSteward to summary

s-rubenstein:
Screenshot 2025-03-24 at 2 02 04 PM

### DIFF
--- a/src/components/dataset/overview/DatasetOverviewPanel.jsx
+++ b/src/components/dataset/overview/DatasetOverviewPanel.jsx
@@ -244,6 +244,13 @@ function DatasetOverviewPanel(props) {
               </Accordion>
             </Grid>
           )}
+          <Grid item xs={4}>
+            {renderTextFieldValue(
+              'Custodians Inherit Steward roles on Snapshots',
+              dataset.inheritSteward,
+              'Custodians added to this dataset will be Stewards on all snapshots created from this dataset',
+            )}
+          </Grid>
         </Grid>
       </TabPanel>
       <TabPanel value={value} index={1}>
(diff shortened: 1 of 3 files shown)


fboulnois (approved):
👍

src/components/dataset/overview/DatasetOverviewPanel.jsx
```diff
@@ -244,6 +244,13 @@ function DatasetOverviewPanel(props) {
               </Accordion>
             </Grid>
           )}
+          <Grid item xs={4}>
+            {renderTextFieldValue(
+              'Custodians Inherit Steward roles on Snapshots',
+              dataset.inheritSteward,
```
pshapiro4broad: I don't think this field should ever be rendered as `(empty)` as it is in the screenshot. It's a boolean field that will always have a value as it's non-null in the service layer.

It looks like this may be a side effect of using how `renderTextFieldValue()` uses `TextContent` on a boolean value; the code in `TextContext` evaluates `text` in a way that conflates boolean with undefined.

What happens if you try to convert it to a string

```suggestion
              dataset.inheritSteward?.toString(),
```

Or, maybe handle this in `TextContent`

The fact that it's optional in the generated code is possibly a bug in the model, which could be fixed by making this property required in the openapi yaml.

pshapiro4broad (approved):
Looks OK but we should make sure that `false` doesn't render as `(empty)` if that's what happening.

src/components/dataset/overview/DatasetOverviewPanel.jsx
rjohanek: if we make it required do we need to back populate ?

rjohanek (commented):
now does it render on the page as 'yes' or 'no'? additionally, maybe we could add a unit test?

rjohanek (commented):
I think it should say yes or no instead of true or false to be consistent with the other values on the page

rjohanek (approved):

PR #1778, opened 2025-03-27
DT-1232: Show inherited stewards on snapshot roles tab

pshapiro4broad:
## Summary

https://broadworkbench.atlassian.net/browse/DT-1232

For snapshots whose source dataset have `inherit steward` enabled, add the dataset custodian users to the snapshot's list of stewards. These inherited stewards are read-only in the snapshot view. If inherit steward isn't enabled, or if there are no custodians, the steward list is unchanged.

Not yet done/bugs

- tests

## Screenshots

User has permission to view the source dataset; inherited stewards are visible. The inherited stewards can't be removed and have a tooltip.

inherited steward

User doesn't have permission to view the source dataset. An error is shown in the Stewards list.

image

### DIFF
--- a/src/components/ManageUsersView.test.tsx
+++ b/src/components/ManageUsersView.test.tsx
@@ -8,6 +8,11 @@ import history from '../modules/hist';
 import globalTheme from '../modules/theme';
 import ManageUsersView from './ManageUsersView';
 
+const user1 = 'user1';
+const user2 = 'user2';
+const user3 = 'user3';
+const readOnlyUser1 = 'readOnlyUser1';
+const readOnlyUser2 = 'readOnlyUser2';
 const mountComponent = (canManageUsers) => {
   const mockStore = createMockStore([]);
   const store = mockStore({});
@@ -18,8 +23,8 @@ const mountComponent = (canManageUsers) => {
           <ManageUsersView
             classes={{}}
             removeUser={canManageUsers ? () => <div /> : undefined}
-            users={['authdomain1', 'authdomain2', 'authdomain3']}
-            readOnlyUsers={['user1', 'user2']}
+            users={[user1, user2, user3]}
+            readOnlyUsers={[readOnlyUser1, readOnlyUser2]}
             readOnlyUserTooltip="read only user tooltip"
           />
         </ThemeProvider>
@@ -33,15 +38,20 @@ describe('ManageUsersView', () => {
     it('Renders user list independent of whether you can manage users', () => {
       mountComponent(canManageUsers);
       cy.get('[data-cy=chip-container]').within(() => {
-        cy.contains('authdomain1').should('exist');
-        cy.contains('authdomain2').should('exist');
-        cy.contains('authdomain3').should('exist');
-        cy.contains('user1').should('exist');
-        cy.contains('user2').should('exist');
+        cy.contains(user1).should('exist');
+        cy.contains(user2).should('exist');
+        cy.contains(user3).should('exist');
+        cy.contains(readOnlyUser1).should('exist');
+        cy.contains(readOnlyUser2).should('exist');
       });
-      cy.contains('user1').trigger('mouseover'); // to show the tooltip for read-only users
-      cy.contains('read only user tooltip').should('be.visible');
     });
+  });
+  it('Does not render remove user button for readOnly users even if canManageUsers is true', () => {
+    mountComponent(true);
+    cy.get(`[data-cy="chip-${user1}"]`).find('[data-testid="CancelIcon"]').should('exist');
+    cy.get(`[data-cy="chip-${readOnlyUser1}"]`)
+      .contains('[data-testid="CancelIcon"]')
+      .should('not.exist');
   });
   it('No container when there are no users', () => {
     const mockStore = createMockStore([]);
--- a/src/components/ManageUsersView.test.tsx
+++ b/src/components/ManageUsersView.test.tsx
@@ -21,6 +21,7 @@ describe('ManageUsersView', () => {
                 classes={{}}
                 removeUser={canManageUsers ? () => <div /> : undefined}
                 users={['authdomain1', 'authdomain2', 'authdomain3']}
+                readOnlyUsers={['user1', 'user2']}
               />
             </ThemeProvider>
           </Provider>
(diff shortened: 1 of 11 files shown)


src/sagas/repository.ts
```diff
@@ -551,15 +551,22 @@ export function* getDatasetById({ payload }: any): any {
...
     yield put({
       type: ActionTypes.GET_DATASET_POLICY_SUCCESS,
       policy: response,
     });
   } catch (err) {
-    showNotification(err);
+    if (options?.suppressErrorNotification) {
+      yield put({
+        type: ActionTypes.GET_DATASET_POLICY_SUCCESS,
+        policy: { data: { policies: [{ name: 'ERROR', members: [] }] } },
```
pshapiro4broad, line 565: This puts an error flag in the response and uses the success state to return the flag to the caller. I'm not sure if this is "good enough" or if I should add another return field for this one case. I didn't see another field for returning out of band errors like this.

src/components/ManageUsersView.tsx
```diff
@@ -3,39 +3,61 @@ import { Box } from '@mui/material';
...
+        onDelete={removeUser ? () => removeUser?.(user) : undefined}
+        variant="outlined"
+        data-cy={`chip-${user}`}
+      />
+    </Box>
+  ));
+  const readOnlyUserChips =
+    readOnlyUsers?.map((user) => (
       <Box data-cy="chip-item" key={user}>
         <Chip
```
pshapiro4broad: I can refactor the common code here

src/components/snapshot/ManagedSnapshotAccess.tsx
```diff
@@ -1,14 +1,16 @@
...
 import _ from 'lodash';
-import { PolicyModel, SnapshotRequestModelPolicies } from 'generated/tdr';
+import { PolicyModel, SnapshotModel, SnapshotRequestModelPolicies } from 'generated/tdr';
 import { connect } from 'react-redux';
 import { TdrState } from 'reducers';
-import { SnapshotRoles } from 'constants';
 import { styled } from '@mui/material/styles';
+import { getDatasetPolicy } from 'actions';
+import { Action } from 'redux';
+import { DatasetRoles, SnapshotRoles } from 'src/constants';
```
pshapiro4broad, line 13: This import change was suggested by sonar and IntelliJ thinks it's better too. The code seems to run OK but maybe there's a reason why the code was importing from `'constants'` before?

src/components/snapshot/ManagedSnapshotAccess.tsx
```diff
@@ -40,6 +62,13 @@ function ManagedSnapshotAccess(props: ManagedSnapshotAccessProps) {
   const readers = getUsers(SnapshotRoles.READER);
   const discoverers = getUsers(SnapshotRoles.DISCOVERER);
   const aggregateDataReaders = getUsers(SnapshotRoles.AGGREGATE_DATA_READER);
+  const inheritedStewards = snapshot.source?.[0].dataset.inheritSteward
+    ? datasetPolicies.find((policy) => policy.name === DatasetRoles.CUSTODIAN)?.members
+    : [];
+  const datasetPolicyError = datasetPolicies.find((policy) => policy.name === 'ERROR');
+  const datasetPolicyErrorMessage =
+    datasetPolicyError &&
+    'Beacause its source dataset has Inherit Steward enabled, this snapshot may have additional stewards that are not listed here.';
```
pshapiro4broad: We didn't originally plan for this case so this message wasn't in the ticket, let me know if there's a better way to word this.

src/components/common/FullViewSnapshotModal.test.tsx
```diff
@@ -34,9 +36,9 @@ const mountFullViewSnapshotModal = (
   });
 
   // Intercept the getBillingProfiles API call onMount
-  cy.intercept('GET', '/api/resources/v1/profiles?offset=0&limit=1000').as('getBillingProfiles');
+  cy.intercept('GET', '/api/resources/v1/profiles?offset=0&limit=1000');
```
pshapiro4broad, line 39: This and the change below were issues flagged by sonar.

snf2ye (approved):

src/components/ManageUsersView.test.tsx
```diff
@@ -8,29 +8,39 @@ import history from '../modules/hist';
...
+  const store = mockStore({});
+  mount(
+    <Router history={history}>
+      <Provider store={store}>
+        <ThemeProvider theme={globalTheme}>
+          <ManageUsersView
+            classes={{}}
+            removeUser={canManageUsers ? () => <div /> : undefined}
+            users={['authdomain1', 'authdomain2', 'authdomain3']}
+            readOnlyUsers={['user1', 'user2']}
```
s-rubenstein: For readability, how would you feel about making the strings for `user1` and `user2` variables and including the 'readonly' as part of the name? And then we can use the variables in the asserts?

src/components/ManageUsersView.tsx
```diff
@@ -3,39 +3,67 @@ import { Box } from '@mui/material';
...
-        />
-      </Box>
-    ));
+function ManageUsersView({
+  removeUser,
+  users,
+  readOnlyUsers,
+  readOnlyUserTooltip,
+}: ManageUsersProps) {
+  const createUserChip = (
```
s-rubenstein, line 18: Do you think it would be worth pulling `UserChip` into a component? Rather than defining it within this component?

src/components/snapshot/ManagedSnapshotAccess.test.tsx
```diff
@@ -35,39 +47,46 @@ const initialState = {
     ],
     userRoles: ['steward', 'reader', 'discoverer', 'aggregate_data_reader'],
   },
+  datasets: datasetState,
+};
+
+const setup = (state) => {
```
s-rubenstein, line 53: Optional: If you wanted, you could keep this to `stateOverrides` and then the mockstore could be: `{ ...initialState, ...stateOverrides }`?

s-rubenstein: Discussed offline, because of nested properties this isn't that simple.

s-rubenstein (approved):

PR #1784, opened 2025-04-10
[DT-1509] Fix billing profile id propagation

s-rubenstein:
One line change to fix the billing profile not getting propagated through createSnapshot from the UI. Can't be easily tested because we don't have tests for our sagas - at some point we should fix that, I will file a ticket.


fboulnois (approved):
👍

rushtong (approved):
👍🏽

PR #1793, opened 2025-06-04
DT-1794: return [] instead of [null] for auth domains when creating a snapshot

snf2ye:
We don't allow creating a snapshot with an auth domain if the source dataset has inherit steward enabled. 

In the UI, we were passing through a list when the value "null" as an auth domain. This was not passing the check in the backend for "no auth domains". ([jade-data-repo check](https://github.com/DataBiosphere/jade-data-repo/blob/9e556f3f52529b50e522564bb878e8d191d54e99/src/main/java/bio/terra/service/snapshot/SnapshotService.java#L218))

Screenshot 2025-06-04 at 12 17 26 PM

Solution: Instead provide an empty array if the user does not select any auth domains in the UI on snapshot create.

### DIFF
--- a/src/reducers/snapshot.ts
+++ b/src/reducers/snapshot.ts
@@ -317,7 +317,7 @@ export default {
           name,
           description,
           mode,
-          dataAccessControlGroups: [authDomain],
+          dataAccessControlGroups: authDomain == null ? [] : [authDomain],
           billingProfileId,
         };
         if (mode === SnapshotRequestContentsModelModeEnum.ByQuery) {


fboulnois (approved):
Nice catch! Should we have a test for this?

rushtong (approved):
👍🏽

snf2ye:
> Should we have a test for this?

@fboulnois 
Good point - I thought about this, but I'm not sure that there is a good way to test this. 
If we want to reproduce the error, we could add an integration test setup with a dataset that has inherit steward enabled and then confirm that we can still create a snapshot. 
Otherwise, I think we would need a unit test to confirm that the payload looked as expected and we don't do any component testing on the javascript files right now.

snf2ye:
I'm going to follow up with a component/unit test tomorrow

PR #1794, opened 2025-06-05
DT-1794 - Add test

snf2ye:
Follow on to https://github.com/DataBiosphere/jade-data-repo-ui/pull/1793

### DIFF
--- a/src/reducers/reducer-utils.ts
+++ b/src/reducers/reducer-utils.ts
@@ -1,0 +1,2 @@
+export const getAuthDomain = (authDomain?: string | null): string[] =>
+  authDomain == null ? [] : [authDomain];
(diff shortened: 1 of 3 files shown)


rushtong (approved):
👍🏽

src/reducers/reducer-utils.ts
```diff
@@ -0,0 +1,2 @@
+export const getAuthDomain = (authDomain?: string | null): string[] =>
+  authDomain == null ? [] : [authDomain];
```
fboulnois:
```suggestion
export const getAuthDomain = (authDomain?: string): string[] =>
  authDomain === undefined ? [] : [authDomain];
```

Since `authdomain` is optional, can we switch the signature to this?

fboulnois (approved):
Looks good, one suggestion:

PR #1798, opened 2025-06-17
[DT-400] Remove Jest types

fboulnois:
Jest isn't being used in the project as far as I can tell.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -10,8 +10,7 @@
   "parser": "@babel/eslint-parser",
   "env": {
     "browser": true,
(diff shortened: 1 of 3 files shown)


MatthewBemis (approved):
nice cleanup!

looks like the jest types dependency was originally introduced in https://github.com/DataBiosphere/jade-data-repo-ui/pull/1072, and wasn't used at that time either.

rjohanek (approved):

rushtong (approved):
👍🏽

PR #1815, opened 2025-08-22
[DT-2094] Upgrade Node to 20.19+ for @vitejs/plugin-react v5.0.1

kevinmarete:
## Addresses 
https://broadworkbench.atlassian.net/browse/DT-2094

## Summary
Bumps Node in the Dockerfile to 20.19.4 so we can update @vitejs/plugin-react to v5.0.1.

- All tests pass
- Verified UI by clicking through a few links

Update


package.json
```diff
@@ -5,7 +5,7 @@
   "version": "0.435.0",
   "author": "Jade Team <jadeteam@broadinstitute.org>",
   "engines": {
-    "node": ">=20.11"
+    "node": ">=20.x.x"
```
fboulnois: We usually pin the engine to a specific version range with a minimum version, see here: https://github.com/DataBiosphere/duos-ui/blob/develop/package.json#L84

fboulnois (commented):
See below:

package.json
```diff
@@ -5,7 +5,7 @@
   "version": "0.435.0",
   "author": "Jade Team <jadeteam@broadinstitute.org>",
   "engines": {
-    "node": ">=20.11"
+    "node": ">=20.19.4 <23"
```
fboulnois:
```suggestion
    "node": ">=20.19.4 <22"
```

fboulnois: I don't know if we've tried `jade-data-repo-ui` with Node LTS 22

fboulnois (approved):
👍

package.json
rushtong: Agreed, and 23 is not LTS.

rushtong (approved):
👍🏽  Pending feedback

PR #1816, opened 2025-08-22
[DT-2094] Upgrade Node 20.11  to 20.19.4

kevinmarete:
## Addresses 
https://broadworkbench.atlassian.net/browse/DT-2094

## Summary
Bumps All Node references to 20.19.4


rushtong (approved):
This looks reasonable to me. I do see that there is no similar buster image in docker hub for 20.19.4. 👍🏽

fboulnois (approved):
👍

PR #1822, opened 2025-08-28
[DT-2149] Updated @cypress/vite-dev-server to 6.0.3

kevinmarete:
## Addresses
https://broadworkbench.atlassian.net/browse/DT-2149

## Summary

To upgrade to `@cypress/15.0.0` and `@cypress/vite-dev-server/7.0.0`, we will first need to migrate the project to at least React 18. The main challenge in moving from `React 17` to `React 18` is that Material-UI’s legacy styling solution (`@mui/styles`) is deprecated and no longer compatible. As a result, components that rely on `makeStyles` or `withStyles` must be refactored to use MUI’s modern styling approaches—either the `styled()` API or the `sx` prop.

| Package | Current | Limit | Latest
| --- | --- | --- | --- |
| [@cypress](https://www.npmjs.com/package/cypress) | `13.17.0` | `14.0.0` | `15.0.0`
| [@cypress/vite-dev-server](https://www.npmjs.com/package/@cypress/vite-dev-server) | `6.0.3` | `7.0.0` | `7.0.0`

### Recommended Next Steps

- Find and replace any use of `@mui/styles` (`makeStyles`, `withStyles`).
- Refactor those components to use `styled()` or the `sx` prop.
- Upgrade to `React 18` once styling is updated.
- Update `@cypress` to `15.0.0` and `@cypress/vite-dev-server` to `7.0.0`.


.github/workflows/test-unit.yml
```diff
@@ -25,16 +25,16 @@ jobs:
...
         run: npm run codegen
       - name: Cypress Run Component Tests
         uses: cypress-io/github-action@v5
         env:
           DISABLE_ESLINT_PLUGIN: true
         with:
-          command: |
-            npx cypress run-ct
+          command: npx cypress run-ct
+          install: false
```
fboulnois, line 37: Can you describe why this is necessary?

.github/workflows/test-unit.yml
```diff
@@ -25,16 +25,16 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - name: npm ci
-        run: npm ci
+        run: npm ci --legacy-peer-deps
```
fboulnois, line 28: Ditto here?

fboulnois (approved):
Looks good, a few minor questions, thank you!

.github/workflows/test-unit.yml
kevinmarete: When  using  `cypress-io/github-action@v5` and run the command `npx cypress run-ct` it reinstalls all dependencies again so `install: false` just tells it not to reinstall.

.github/workflows/test-unit.yml
kevinmarete: The parameter `--legacy-peer-deps` ensures that if there are any conflicts in packages, they are resolved instead of failing.

eweitz (approved):

PR #1839, opened 2025-10-23
[DC-1801] repository.ts calls a DUOS autocomplete endpoint that doesn't exist

rjohanek:
### Why
TDR UI was calling an endpoint in DUOS that had been removed. The impact of this was that no DUOS datasets were viewable in the TDR UI and it always gave the message that users did not have access to any DUOS datasets. This also meant that the DUOS datasets could not be linked to a TDR snapshot on TDR UI either.

### Summary of Changes
Call the a different endpoint the search/index endpoint and filter for datasets.
Update the DUOS Dataset Model in TDR UI to match the response

### Testing
Adds unit tests for the SnapshotOverviewPanel

### DIFF
--- a/src/sagas/repository.ts
+++ b/src/sagas/repository.ts
@@ -949,19 +949,18 @@ export function* getDuosDatasets(): any {
   try {
     const duosUrl = yield select(getDuosUrl);
     const url = `${duosUrl}/api/dataset/search/index`;
-    const params = {
-      query: {
-        bool: {
-          must: [
-            {
-              match: {
-                _type: 'dataset',
-              },
-            },
-          ],
-        },
-      },
+    const params = 
+    {
+      "query":
+      {"bool":
+        {"must":
+          [{"match":{"_type":"dataset"}},
+           {"term":{"study.publicVisibility": true}},
+           {"term":{"dataLocation.keyword": "TDR Location"}}]
+        }
+      }
     };
+
     const response = yield call(authPost, url, params);
     yield put({
       type: ActionTypes.GET_DUOS_DATASETS_SUCCESS,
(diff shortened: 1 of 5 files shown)


calypsomatic (approved):

src/sagas/repository.ts
```diff
@@ -948,8 +948,21 @@ export function* changePage(page: number): any {
...
-    const url = `${duosUrl}/api/dataset/autocomplete`;
-    const response = yield call(authGet, url);
+    const url = `${duosUrl}/api/dataset/search/index`;
+    const params = {
+      query: {
+        bool: {
+          must: [
+            {
+              match: {
+                _type: 'dataset',
```
rushtong: This will find any dataset, even those that might not be approved by a Data Access Committee, Open Access, External Access, or publicly visible to users for adding to DARs. It would be nice to add some filtering to ensure users are not presented with datasets that might not be appropriate for linking.

rushtong: There is also a field for location that might be helpful in filtering:
```
   ...
   "dataLocation": "TDR Location",
   ...
```

rjohanek: I'd like to better understand the filtering requirements here. Yes, definitely to the TDR location filter. In terms of approved by a Data Access Committee, Open Access, External Access, or publicly visible, I have some questions. Can an open access dataset be in TDR?
1. approved by a DAC - if we are including open access data then we don't want this to be a requirement, right?
2. external access - this will be filtered out by the TDR data location, right?
3. how is publicly visible relevant here? I assumed that if the dataset is not publicly visible it will not be returned. Is that not the case? Do I need to filter out private datasets?

rjohanek: So maybe something like: 
If controlled access, must be dac approved
Always must be tdr data location 
Always must be publicly visible
Or would we want a user to see non DAC approved datasets? if for example they are data custodians of the dataset setting up their dataset in TDR and DUOS. They would be linking before the DAC approves it and it is available to researchers.

fboulnois (approved):
looks great, thank you! 👍

